PR #6, opened 2023-03-08
Add tests

PhilippeMoussalli:
This PR adds basic pytest for the helper functionalities provided in Express:

- Logger
- Storage
- kfp Utils
- io
- manifest

Next steps is to include tests for the Express pandas components + run tests with github actions


RobbeSneyders:
Thanks Philippe. Can you rebase on main? Now it shows some changes already in main in the diff which is confusing.

PhilippeMoussalli:
> Thanks Philippe. Can you rebase on main? Now it shows some changes already in main in the diff which is confusing.

done

requirements.txt
```diff
@@ -0,0 +1,5 @@
+bandit[toml]==1.7.4
```
NielsRogge, line 1: This file shouldn't be added, I assume

pyproject.toml
```diff
@@ -48,6 +48,7 @@ bandit = { version = "^1.7.4", extras = ["toml"] }
 liccheck = "^0.7.3"
 pylint = "2.16.4"
 pre-commit = "^3.1.1"
+pytest= "^7.2.2"
```
NielsRogge, line 51: Can't comment on the file above, but the PR adds an empty setup.cfg file in a mlpipelines/common folder, should not be the case

express/manifest_example.json
```diff
@@ -1,25 +1,25 @@
 {
```
RobbeSneyders, line 1: What is the purpose of this file? The `express` directory seems like a strange place for it.

tests/__init__.py
```diff
@@ -0,0 +1,3 @@
+"""Tests for pandas component"""
+
+from express_components import pandas_components
```
RobbeSneyders, line 3: What is the purpose of this import?

RobbeSneyders (commented):
You are importing from `express_components` in the tests. Is this based on the old repo structure?

Could you also add
- Pylint for the tests in pre-commit
- Running the tests in github actions

tests/__init__.py
philippe-ml6: added by accident

express/manifest_example.json
philippe-ml6: It's to illustrate the layout of the manifest. I'll include it somewhere in the docs

PhilippeMoussalli:
> You are importing from `express_components` in the tests. Is this based on the old repo structure?
> 
> Could you also add
> 
> * Pylint for the tests in pre-commit
> * Running the tests in github actions

@RobbeSneyders 
I was just checking the pre-commit setup. It differs a bit from the bitbucket setup in a few ways: 

- It seems like we do not have a `.pylintrc' configured. Should we add one modify `.pre-commit-config.yaml` to use it? 
- In BB we use `pylint-modules` along with the `pylint.py` script and that is integrated with bitbucket pipelines but here we're only using the `pylint-modules` for the license checker. How to define which modules to lint? how to run it locally?

RobbeSneyders:
> It seems like we do not have a .pylintrc configured. Should we add one modify .pre-commit-config.yaml to use it?

No `.pylintrc` indeed. I fixed all the issues so it works with the default settings. I say we try to keep it this way :)

> In BB we use pylint-modules along with the pylint.py script and that is integrated with bitbucket pipelines but here we're only using the pylint-modules for the license checker. How to define which modules to lint? how to run it locally?

We should actually remove the `pylint-modules` and `run_license_check.sh` since we no longer use it after https://github.com/ml6team/express/pull/7. 
We can run pylint on multiple directories by extendingo the `files` argument of the pre-commit hook, or adding a separate hook. Eg. (untested):

```
    hooks:
      - id: pylint
        ...
        files: |
            (?x)^(
                express/|
                tests/
            )$
        ...
```

PhilippeMoussalli:
> We can run pylint on multiple directories by extendingo the files argument of the pre-commit hook

Thanks for clarifying. It should be good now

.pre-commit-config.yaml
```diff
@@ -9,7 +9,8 @@ repos:
         entry: pylint
         language: system
         types: [python]
-        files: "^express/"
+        files: "^express/\n
+                |^tests/"
```
RobbeSneyders, line 13: Any reason to go with this instead of what I proposed?
My proposal came from the [pre-commit docs](https://pre-commit.com/#regular-expressions) and I think it's more readable and easier to extend.

PhilippeMoussalli: I switched it over to the regex approach, the regex statement did not work initially so that's why I opted for the string approach. But then I modified it a bit and now it works.

express/io.py
```diff
@@ -8,15 +8,15 @@
 logger = logging.getLogger(__name__)
 
 
-def get_file_extension(file_url: str) -> str:
+def get_file_extension(file_name: str) -> str:
     """
     Function that returns a file extension from a file url
     Args:
-        file_url (str): the file url to return the extension from
+        file_name (str): the file url to return the extension from
```
NielsRogge, line 15: The docstring still says URL

RobbeSneyders (approved):
Thanks Philippe!
One minor comment from Niels and we can merge.

PR #8, opened 2023-03-08
Add HF Dataset components

NielsRogge:
This PR adds boilerplate `HFDatasetLoaderComponent` and `HFDatasetTransformComponent` classes.

Questions:
- in the Datasets library, there's a distinction between a `Dataset` and a `DatasetDict`. The latter is just a dictionary of several `Dataset` instances, with keys like "train", "validation", "test". I assume the index could be stored as a `Dataset`, whereas data sources can be either a `Dataset` or a `DatasetDict`.
- if you use [`load_dataset`](https://huggingface.co/docs/datasets/v2.10.0/en/package_reference/loading_methods#datasets.load_dataset) without providing a split (like "train", "validation"), this method returns a `DatasetDict` instead of a `Dataset`. So was wondering how we could handle these splits a user might have.

To do:

- [ ] add an example

### DIFF
--- a/express/components/hf_datasets_components.py
+++ b/express/components/hf_datasets_components.py
@@ -52,7 +52,7 @@ class HFDatasetsDataset(ExpressDataset[List[str], Union[datasets.Dataset, datase
                 tmp_dir
             )
 
-            data_source_hf_datasets = load_dataset("parquet", data_dir=local_parquet_path)
+            data_source_hf_datasets = load_dataset("parquet", data_dir=local_parquet_path, split="train")
 
             if index_filter:
                 return data_source_hf_datasets.filter([index_filter])
@@ -70,7 +70,6 @@ class HFDatasetsDatasetHandler(ExpressDatasetHandler[List[str], datasets.Dataset
         with tempfile.TemporaryDirectory() as temp_folder:
             # TODO: uploading without writing to temp file
             # TODO: sharded parquet? not sure if we should shard the index or only the data sources
-            # HF shards by default
             dataset_path = f"{temp_folder}/{name}.parquet"
 
             data.to_parquet(path=dataset_path)
--- a/express/components/hf_datasets_components.py
+++ b/express/components/hf_datasets_components.py
@@ -32,13 +32,14 @@ class HFDatasetsDataset(ExpressDataset[List[str], Union[datasets.Dataset, datase
         with tempfile.TemporaryDirectory() as tmp_dir:
             local_parquet_path = STORAGE_HANDLER.copy_file(self.manifest.index.location, tmp_dir)
 
-            dataset = load_dataset("parquet", data_files=local_parquet_path)
+            # we specify "train" here to get a `Dataset` instead of a `DatasetDict`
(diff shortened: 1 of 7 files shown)


express/components/hf_datasets_components.py
```diff
@@ -0,0 +1,131 @@
...
+# pylint: disable=unsubscriptable-object
+HFDatasetsDatasetDraft = ExpressDatasetDraft[List[str], Union[pd.DataFrame, pd.Series]]
+
+# pylint: disable=no-member
+STORAGE_MODULE_PATH = StorageHandlerModule().to_dict()[os.environ.get('CLOUD_ENV', 'GCP')]
+STORAGE_HANDLER = importlib.import_module(STORAGE_MODULE_PATH).StorageHandler()
+
+
+# pylint: disable=too-few-public-methods
+class HFDatasetsDataset(ExpressDataset[List[str], Union[datasets.Dataset, datasets.DatasetDict]]):
```
NielsRogge, line 27: Also was wondering whether we could have a better name (DatasetsDataset isn't the best name)

express/components/hf_datasets_components.py
```diff
@@ -0,0 +1,131 @@
...
+        with tempfile.TemporaryDirectory() as tmp_dir:
+
+            data_source_location = data_source.location
+
+            local_parquet_path = STORAGE_HANDLER.copy_parquet(
+                data_source_location,
+                tmp_dir
+            )
+
+            data_source_hf_datasets = load_dataset("parquet", data_dir=local_parquet_path, split="train")
```
NielsRogge, line 55: load_dataset also supports things like multiprocessing (if there are multiple CPU cores) and streaming, not sure if we want to use that too

NielsRogge:
@RobbeSneyders so the pipeline fails because `datasets` is not in the dependencies of pyproject.toml. However I don't think it makes sense to make this a hard dependency, which is why I was arguing for soft dependency checks depending on which components a user might want to use. Thoughts?

RobbeSneyders:
I think handling soft dependencies would be good indeed. I would add it as a dependency in this PR, and then open a new PR where we implement the soft dependency checks for both `datasets` and `pandas`.

RobbeSneyders:
Seems like I was a bit too strict with the allowed licenses. We can allow weak-copyleft licenses as well.

You can add the following ones and it will fix the pipeline:
- GNU Library or Lesser General Public License (LGPL)
- GNU Lesser General Public License v2 (LGPLv2)
- GNU Lesser General Public License v2 or later (LGPLv2+)
- GNU Lesser General Public License v3 (LGPLv3)
- GNU Lesser General Public License v3 or later (LGPLv3+)
- Mozilla Public License 1.0 (MPL)
- Mozilla Public License 1.1 (MPL 1.1)
- Mozilla Public License 2.0 (MPL 2.0)
- The Unlicense (Unlicense)

RobbeSneyders (approved):
Thanks! LGTM.

Can you just remove the `pylint: disable`s and resolve the conflict?

PR #10, opened 2023-03-08
Add black formatting

RobbeSneyders:
From the black website:

> Black is the uncompromising Python code formatter. By using it, you agree to cede control over minutiae of hand-formatting. In return, Black gives you speed, determinism, and freedom from pycodestyle nagging about formatting. You will save time and mental energy for more important matters.
> 
> Blackened code looks the same regardless of the project you're reading. Formatting becomes transparent after a while and you can focus on the content instead.
> 
> Black makes code review faster by producing the smallest diffs possible.

We've been using it on Connexion for about a year and it's a pleasure to work with once its usage is automated by pre-commit.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,6 +20,7 @@ repos:
             "-sn", # Don't display the score
             "--disable=fixme"
           ]
+
   - repo: https://github.com/PyCQA/bandit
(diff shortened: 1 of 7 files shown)


NielsRogge (approved):
Thanks for adding! We're using it for HF Transformers as well

PR #11, opened 2023-03-08
Add isort to pre-commit

RobbeSneyders:
> isort your imports, so you don't have to.
> 
> isort is a Python utility / library to sort imports alphabetically, and automatically separated into sections and by type.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,3 +16,11 @@ repos:
             "-sn", # Don't display the score
             "--disable=fixme"
           ]
+  - repo: https://github.com/PyCQA/bandit
+    rev: 1.7.4
+    hooks:
+      - id: bandit
+        name: bandit
+        types: [python]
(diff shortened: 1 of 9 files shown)


NielsRogge:
Was wondering if we could use Ruff for isort and flake8: https://github.com/charliermarsh/ruff. It's written in Rust and blazingly fast. It's also used by Pandas and I recently made HF make the switch to it :D

RobbeSneyders:
Never heard of it, but looks interesting

> 10-100x faster than existing linters

Pylint is already quite slow in our pre-commit

> pyproject.toml support

Nice

> Autofix support, for automatic error correction (e.g., automatically remove unused imports)

Perfect for pre-commit

>  Monorepo-friendly, with [hierarchical and cascading configuration]

Could be great for ML6 to replace those `run_pylint` scripts

I'll try it out and replace both pylint and isort if I don't run into any issues.

PR #12, opened 2023-03-09
Switch pylint for Ruff

RobbeSneyders:
Based on [Niels' suggestion](https://github.com/ml6team/express/pull/11#issuecomment-1461521485).

Ruff is a lot faster than Pylint and an also prevents the need for isort. While it [doesn't completely cover all pylint checks](https://github.com/charliermarsh/ruff/issues/970), a lot of the non-covered issues will be covered by bandit or mypy (still to add) instead.


NielsRogge (approved):
Thanks for adding!

PR #15, opened 2023-03-14
Add kfp utils

PhilippeMoussalli:
PR to refactor pipelines directory:

- Pulling out common configs to root directory
- moving kfp compiler to common express helpers
- resolving some pip issues with kfp and adding kfp as a dependency
- cleaning up some files
- Bugfix kfp upload


express/kfp_utils.py
```diff
@@ -1,6 +1,16 @@
 """General kfp helpers"""
+import os
 import ast
 import logging
+import yaml
+from typing import Callable
+
+import kfp
```
RobbeSneyders, line 8: This adds kfp as a dependency to express, which is a heavy dependency that's not needed for components. I would like to challenge if we really need it. And if we do, we should make it a soft dependency (can be done in a separate PR as this needs to be done for other dependencies as well).

express/kfp_utils.py
```diff
@@ -18,3 +28,56 @@ def parse_kfp_list(kfp_parsed_string: str) -> list:
         list: the list representation of the json string
     """
     return ast.literal_eval("".join(kfp_parsed_string))
+
+
+def configure_pipeline(file_path: str) -> None:
```
RobbeSneyders, line 33: Two comments on this function:

- It's not clear what it does based on the name. If it only sets the image pull policy, let's name the function accordingly.
- Do we actually need to do this? This is a bit magical, which may lead to unexpected behavior, and I don't see why we cannot just add this to all our `component.yaml` files explicitly?

RobbeSneyders (commented):
Thanks @PhilippeMoussalli 

Could you remove the change to the location of the `components.config` file? The change seems unrelated to the rest of the PR but adds a lot of files to the diff of this PR.

I'm also not convinced of the change, because I think we should get rid of the `components.config` file by lifting the `build_image.sh` script and adding the docker registry to it directly.

I would like to move towards the following repo structure:
```
- express
- components  # These are the components we want to offer as images
  |- component-X
  |- component-Y
  |- build_image.sh  # Contains the docker registry for Express components
- examples
  |- example-pipeline
      |- components  # These are examples on how to define custom components
      |  |- example-component
      |- build_image.sh  # Contains an example docker registry
```

I left some more comments on the actual focus of this PR below :)

express/kfp_utils.py
RobbeSneyders: @NielsRogge would like your view on this as well since we've been discussing this before.

examples/pipelines/example_pipeline/components/example_component/README.md
```diff
@@ -0,0 +1,39 @@
+# example_component
+
+### Description
```
NielsRogge, line 3: Not sure if we need a new example pipeline with a single component if there's already an example pipeline added in #16 ?

express/kfp_utils.py
NielsRogge: We can create a `is_kfp_available` function to check whether kfp is installed, and throw an error if it it's not. This way, we can handle soft dependencies

examples/pipelines/example_pipeline/components/example_component/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM python:3.7-slim
```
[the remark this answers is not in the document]
NielsRogge: I'm using Python 3.8 in #16

NielsRogge (commented):
Ok for me to add the kfp_utils module in express with a soft dependency on kfp, but not sure whether a new example pipeline needs to be added here.

examples/pipelines/example_pipeline/components/example_component/README.md
PhilippeMoussalli: It guess we won't need it once we have the example HF pipeline, i'll remove it

pyproject.toml
```diff
@@ -40,9 +40,12 @@ classifiers = [
 
 [tool.poetry.dependencies]
 python = "^3.8"
+datasets = "2.10.1"
```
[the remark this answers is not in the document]
PhilippeMoussalli: good catch, i'll update it

express/kfp_utils.py
PhilippeMoussalli: I'll remove it, this was originally copied over from some the KFP boilerplate but I guess we won't need it in our case

examples/pipelines/example_pipeline/components/example_component/Dockerfile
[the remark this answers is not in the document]
PhilippeMoussalli: I'm removing the example component from this PR

examples/pipelines/config.py
```diff
@@ -30,4 +30,4 @@ class KubeflowConfig(GeneralConfig):
     ARTIFACT_BUCKET = f"{GeneralConfig.GCP_PROJECT_ID}-kfp-output"
     CLUSTER_NAME = "kfp-express"
     CLUSTER_ZONE = "europe-west4-a"
-    HOST = "https://472c61c751ab9be9-dot-europe-west1.pipelines.googleusercontent.com"
+    HOST = "https://472c61c751ab9be9-dot-europe-west1.pipelines.googleusercontent.com"
```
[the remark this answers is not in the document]
PhilippeMoussalli: not sure why this wasn't picked up by the linters

examples/pipelines/example_pipeline/components/example_component/build_image.sh
```diff
@@ -0,0 +1,28 @@
...
+# Set some variables
+ARTIFACT_PATH=${PROJECT_ARTIFACT_PATH}
+IMAGE_NAME="components/${PWD##*/}"
+IMAGE_TAG="latest"
+
+# Create full name of the image
+FULL_IMAGE_NAME=${ARTIFACT_PATH}/${IMAGE_NAME}:${IMAGE_TAG}
+echo $FULL_IMAGE_NAME
+
+gcloud builds submit --machine-type n1-highcpu-32 . -t "$FULL_IMAGE_NAME"
```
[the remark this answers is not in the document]
PhilippeMoussalli: Is it still possible to have caching enabling when switching to `docker` with kaniko? 
https://cloud.google.com/build/docs/optimize-builds/kaniko-cache

express/kfp_utils.py
PhilippeMoussalli: I'll implement this in a separate PR. Any pointers, examples on how this is handled in other repos?

examples/pipelines/config.py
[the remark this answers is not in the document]
RobbeSneyders: This directory is not linted yet

express/kfp_utils.py
```diff
@@ -18,3 +28,56 @@ def parse_kfp_list(kfp_parsed_string: str) -> list:
...
+        if "container" in template:
+            # Set imagePullPolicy
+            template["container"]["imagePullPolicy"] = "Always"
+
+    # Store the pipeline yaml file
+    with open(file_path, "w") as f:
+        yaml.dump(pipeline, f)
+
+
+def compile_and_upload_pipeline(
```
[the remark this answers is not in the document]
PhilippeMoussalli: Should it be added at the pipeline level? In that case, it should be at the root level of the pipelines but referenced from the parent directory in every pipeline.py for the different pipelines

- examples
  |- example-pipelines-a
      |- components  # These are examples on how to define custom components
      |  |- example-component
      |- build_image.sh  # Contains an example docker registry
      |- pipeline.py  # Calls the compile_and_upload_pipeline
  |- example-pipelines-b
      |- 
      |-
  |- kfp-utils.py # contains the upload func

PhilippeMoussalli:
> Thanks @PhilippeMoussalli
> 
> Could you remove the change to the location of the `components.config` file? The change seems unrelated to the rest of the PR but adds a lot of files to the diff of this PR.
> 
> I'm also not convinced of the change, because I think we should get rid of the `components.config` file by lifting the `build_image.sh` script and adding the docker registry to it directly.
> 
> I would like to move towards the following repo structure:
> 
> ```
> - express
> - components  # These are the components we want to offer as images
>   |- component-X
>   |- component-Y
>   |- build_image.sh  # Contains the docker registry for Express components
> - examples
>   |- example-pipeline
>       |- components  # These are examples on how to define custom components
>       |  |- example-component
>       |- build_image.sh  # Contains an example docker registry
> ```
> 
> I left some more comments on the actual focus of this PR below :)

Left some comments and updated the PR. I agree with the current suggestion, I'll handle this in a separate PR, I would modify the `build_image.sh` to build all components in the current directory with the possibility to only build some if a flag is passed (avoid building everytime during development_

RobbeSneyders:
> Left some comments and updated the PR. I agree with the current suggestion, I'll handle this in a separate PR, I would modify the `build_image.sh` to build all components in the current directory with the possibility to only build some if a flag is passed (avoid building everytime during development_

My idea was to still have it build only one image at a time, but to pass in the path to the directory to build.

examples/pipelines/example_pipeline/components/example_component/build_image.sh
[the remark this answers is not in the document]
RobbeSneyders: Not sure, but we will move away from google cloud registry for these images in the future anyway.

express/kfp_utils.py
RobbeSneyders: There's different ways to tackle this.
How we do it at Connexion: https://github.com/spec-first/connexion/pull/1668
HuggingFace does it differently, you can have a look there as well.

PhilippeMoussalli: Thanks, i'll check them out

PhilippeMoussalli:
> My idea was to still have it build only one image at a time, but to pass in the path to the directory to build.

Alright i'll handle this in a separate PR. Anything else to update in this one?

examples/pipelines/finetune_stable_diffusion/components/image_caption_component/README.MD
```diff
@@ -13,7 +13,7 @@ See [`component.yaml`](component.yaml) for a more detailed description on all th
 * In general, it is [common practice](https://huggingface.co/datasets/lambdalabs/pokemon-blip-captions) 
 to generate relatively short descriptions when finetuning stable diffusion.  
 * You can test the captioning of the BLIP model interactively in this
-[notebook](../../../notebooks/sd-training/dataset_creation.ipynb).
+[notebook](../../notebooks/sd-training/dataset_creation.ipynb).
```
RobbeSneyders, line 16: This notebook is no longer part of the repo, should it?

RobbeSneyders (approved):
Thanks @PhilippeMoussalli. 
One small final comment.

PR #16, opened 2023-03-14
Add HF dataset example pipeline

NielsRogge:
This PR adds a pipeline example that consists of a single component.

The component loads a dataset from the 🤗 hub and uploads it to GCS, leveraging Express. As can be seen in the requirements.txt file of the component, it has a dependency on Express.


examples/pipelines/hf_dataset_pipeline/README.md
```diff
@@ -0,0 +1,63 @@
+In this guide, we explain how to implement a KubeFlow pipeline and run it on KubeFlow on GKE (Google Kubernetes Engine).
+
+# 1. Implement the components
+
+1. Create a "components" folder where all KubeFlow components will live.
```
PhilippeMoussalli, line 5: I think a lot of this documentation is repeated from something i'v already included before and seems to be more general to kubeflow rather than this specific pipeline.

Could we:

* Define this somewhere else? Maybe on the root level of the pipelines.
* Combine it with the current existing documentation?https://github.com/ml6team/express/blob/main/docs/kubeflow_101.md

examples/pipelines/hf_dataset_pipeline/README.md
```diff
@@ -0,0 +1,63 @@
+In this guide, we explain how to implement a KubeFlow pipeline and run it on KubeFlow on GKE (Google Kubernetes Engine).
```
PhilippeMoussalli, line 1: The name 'hf_dataset_pipeline` is a bit abstract and can indicate many things. Can we change it to Stable diffusion data collection or finetuning?

examples/pipelines/hf_dataset_pipeline/components/load_from_hub/build_image.sh
```diff
@@ -0,0 +1,28 @@
...
+
+# Set some variables
+ARTIFACT_PATH=${PROJECT_ARTIFACT_PATH}
+IMAGE_NAME="components/${PWD##*/}"
+IMAGE_TAG="latest"
+
+# Create full name of the image
+FULL_IMAGE_NAME=${ARTIFACT_PATH}/${IMAGE_NAME}:${IMAGE_TAG}
+echo $FULL_IMAGE_NAME
+
```
PhilippeMoussalli, line 27: I think you can include here `gcloud config set builds/use_kaniko True` to enable it automatically and avoid having the user look through the docs for figuring out this step.
 Can you please include it in all the build_image.sh that exist? you can also update the docs then to just mention that cashing is used

examples/pipelines/hf_dataset_pipeline/components/load_from_hub/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+ENV PATH $PATH:/usr/local/gcloud/google-cloud-sdk/bin
+
+# install requirements
+COPY requirements.txt /
+RUN pip3 install --no-cache-dir -r requirements.txt
+
+# Copy over src-files of the component
+COPY src /src
+
+# TODO: ADD EXTRA FILES HERE IF NEEDED!
```
PhilippeMoussalli, line 26: This line can be removed

examples/pipelines/hf_dataset_pipeline/components/load_from_hub/component.yaml
```diff
@@ -0,0 +1,24 @@
...
+outputs:
+    - name: output_manifest
+      description: Path to the output manifest
+
+implementation:
+    container:
+        image: europe-west4-docker.pkg.dev/soy-audio-379412/soy-audio-379412-default-repository/kubeflow-components/components/load_from_hub:latest
+        command: [
+            python3, main.py,
+            --extra-args,       {inputValue: extra_args},
```
PhilippeMoussalli, line 21: Nitpick alert: could you align the values starting with `{`

examples/pipelines/hf_dataset_pipeline/components/load_from_hub/requirements.txt
```diff
@@ -0,0 +1,2 @@
+# Add component requirements here
```
PhilippeMoussalli, line 1: The line can be removed

examples/pipelines/hf_dataset_pipeline/components/load_from_hub/requirements.txt
```diff
@@ -0,0 +1,2 @@
...
+git+https://github.com/ml6team/express.git@main
```
PhilippeMoussalli, line 2: We might have to think about a more efficient way of using this (e.g. if you make local changes in Express and you want to test them). Could you add a TODO or a ticket for it?

examples/pipelines/hf_dataset_pipeline/helpers/lightweight.py
```diff
@@ -0,0 +1,11 @@
+"""Collection of lightweight components"""
```
PhilippeMoussalli, line 1: This file is not needed I guess

examples/pipelines/hf_dataset_pipeline/components/load_from_hub/src/main.py
```diff
@@ -0,0 +1,49 @@
...
+
+        # 2) Create an example index
+        index_list = [f"image_{idx}" for idx in range(len(caption_dataset))]
+        index_df = pd.DataFrame(index_list, columns=['index'])
+        index = Dataset.from_pandas(index_df)
+
+        caption_dataset = concatenate_datasets([index, caption_dataset])
+        
+        # 2.2) Create data_source dictionary
+        data_sources = {"captions": caption_dataset}
```
PhilippeMoussalli, line 40: ok for the initial loading but we might want to think of a more efficient way of doing this. Right now we're materializing the dataset in both the `data_sources` dict and also the `caption_dataset` which can lead to OoO issues with large datasets

examples/pipelines/hf_dataset_pipeline/hf_dataset_pipeline.py
```diff
@@ -0,0 +1,44 @@
...
+from kfp import dsl
+
+from config.general_config import GeneralConfig, KubeflowConfig
+from helpers.upload import compile_and_upload_pipeline
+
+# Load Component
+load_from_hub_op = comp.load_component(
+    '/Users/nielsrogge/Documents/python_projects/express/examples/pipelines/hf_dataset_pipeline/components/load_from_hub/component.yaml')
+
+run_id = '{{pod.name}}'
```
PhilippeMoussalli, line 17: This won't work since it's not materialized, Could you add a TODO or a ticket for this and replace with a default value?

PhilippeMoussalli (changes requested):
Reviewed! Overall looks good but left a few comments for a few things to adjust

examples/pipelines/hf_dataset_pipeline/README.md
RobbeSneyders: Depends on what we want to showcase. If this example is meant to showcase how to work with a HuggingFace dataset, I think the current name is appropriate.

examples/pipelines/hf_dataset_pipeline/components/load_from_hub/README.md
```diff
@@ -0,0 +1,14 @@
+This component can be run as follows:
+
+```
+python3 main.py --extra-args '{"dataset_name": "lambdalabs/pokemon-blip-captions"}' \
+ --output-manifest <local_path> \
+  --metadata-args '{"run_id":"test","component_name":"test_component", \
+  "artifact_bucket":"soy-audio-379412_kfp-artifacts"}'
```
RobbeSneyders, line 7: Let's replace the artifact bucket with a placeholder as well.

examples/pipelines/hf_dataset_pipeline/components/load_from_hub/README.md
```diff
@@ -0,0 +1,14 @@
+This component can be run as follows:
```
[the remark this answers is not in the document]
RobbeSneyders: It documents the arguments specific to this component, so for me this is fine to have here.

examples/pipelines/hf_dataset_pipeline/components/load_from_hub/build_image.sh
RobbeSneyders: Would propose not to do this in this PR and move to `docker` instead of `gcloud` anyway.

examples/pipelines/hf_dataset_pipeline/components/load_from_hub/src/main.py
RobbeSneyders: I don't think this is the case since both are referring to the same object.

But the current setup is indeed limited by the local memory, so we should find a way to stream this data instead.

NielsRogge: Does Python duplicate the memory when using a variable in a dict?

examples/pipelines/hf_dataset_pipeline/helpers/upload.py
```diff
@@ -0,0 +1,83 @@
+"""
```
RobbeSneyders, line 1: See https://github.com/ml6team/express/pull/15

examples/pipelines/hf_dataset_pipeline/hf_dataset_pipeline.py
RobbeSneyders: What do you mean with this? How should this value be substituted?

examples/pipelines/hf_dataset_pipeline/hf_dataset_pipeline.py
```diff
@@ -0,0 +1,44 @@
...
+import json
+
+from kfp import components as comp
+from kfp import dsl
+
+from config.general_config import GeneralConfig, KubeflowConfig
+from helpers.upload import compile_and_upload_pipeline
+
+# Load Component
+load_from_hub_op = comp.load_component(
```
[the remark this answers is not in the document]
NielsRogge: Ok, I copied this from the example component

examples/pipelines/hf_dataset_pipeline/hf_dataset_pipeline.py
PhilippeMoussalli: It should be defined inside the pipeline function for it to take place otherwise you get an error. Haven't dug deep into it yet. Can also use `kfp.dsl.RUN_ID_PLACEHOLDER` somewhere inside Express 

Screenshot from 2023-03-15 16-59-47

examples/pipelines/hf_dataset_pipeline/components/load_from_hub/src/main.py
PhilippeMoussalli: @RobbeSneyders you're right I think it just references it. I would still opt to use streaming instead 
https://huggingface.co/docs/datasets/stream

examples/pipelines/hf_dataset_pipeline/components/load_from_hub/README.md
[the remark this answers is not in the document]
PhilippeMoussalli: I think it's more general to data loading in general which will be similar for all dataset frameworks. IMO it should be only mentioned once in a reference doc. I think it would be nice to have a description of the component as well as a schema description for the args in `--extra-args`

examples/pipelines/hf_dataset_pipeline/README.md
NielsRogge: Going to keep the name for now

examples/pipelines/hf_dataset_pipeline/components/load_from_hub/build_image.sh
NielsRogge: Will keep this for now, and in a follow-up PR we can update all components to leverage `docker` (in a single script that builds the images for all components)

examples/pipelines/hf_dataset_pipeline/helpers/lightweight.py
NielsRogge: Will remove

examples/pipelines/hf_dataset_pipeline/hf_dataset_pipeline.py
NielsRogge: Ok great!

examples/pipelines/hf_dataset_pipeline/components/load_from_hub/requirements.txt
NielsRogge: Yes so @PhilippeMoussalli the examples should always rely on the main branch (not on a certain commit.

I guess we can update the requirements.txt to a certain commit only when creating a new example (for debug purposes), but once it's ready we should update to just the main branch

examples/pipelines/hf_dataset_pipeline/components/load_from_hub/src/main.py
NielsRogge: Streaming is mostly useful when you'd like to transform the dataset, not when you want to upload it to the cloud like here, so will keep it like it is. However for our transform components we can leverage streaming

examples/pipelines/hf_dataset_pipeline/helpers/upload.py
NielsRogge: Thanks, updated

NielsRogge:
@PhilippeMoussalli @RobbeSneyders PR is ready for a final review. Addressed all comments + added a transform component to the pipeline which leverages `HFDatasetsTransformComponent`.

examples/pipelines/hf_dataset_pipeline/components/add_captions/component.yaml
```diff
@@ -0,0 +1,29 @@
...
+outputs:
+    - name: output_manifest
+      description: Path to the output manifest
+
+implementation:
+    container:
+        image: europe-west4-docker.pkg.dev/soy-audio-379412/soy-audio-379412-default-repository/kubeflow-components/components/add_captions:latest
+        command: [
+            python3, main.py,
+            --input-manifest,       {inputPath: input_manifest},
```
PhilippeMoussalli, line 25: nitpick: can you indent the value and path arguments so that they allign?

examples/pipelines/hf_dataset_pipeline/components/load_from_hub/src/main.py
```diff
@@ -0,0 +1,47 @@
...
+        Dict[str, Union[str, int, float, bool]]] = None) -> HFDatasetsDatasetDraft:
+        """
+        An example function showcasing the data loader component using Express functionalities
+        Args:
+            extra_args (Optional[Dict[str, Union[str, int, float, bool]]): optional args to pass to
+             the function (e.g. seed data source)
+        Returns:
+            HFDatasetsDatasetDraft: a dataset draft that creates a plan for an output datasets/manifest
+        """
+        configure_logging()
```
PhilippeMoussalli, line 27: Include some logging statements in this example as well

PhilippeMoussalli (approved):
LGTM! Left a few more small comments and then it's ready to ship

NielsRogge:
@RobbeSneyders feel free to approve

PR #17, opened 2023-03-16
Handle soft dependencies

PhilippeMoussalli:
PR that adds checks for soft dependencies that might me missing. Includes both the addition of the dependencies in the `pyproject.toml` as well as a function to handle missing dependencies and instruct the user to install express with extra flags.


RobbeSneyders:
Thanks @PhilippeMoussalli 

This should be an internal function though, so it should not be necessary to use it in the examples. The reason you need this, is because you are still raising an `ImportError` at import time, just with a clearer message. We should instead delay this error until the dependency is actually used.

See the way it is done in [Connexion](https://github.com/spec-first/connexion/blob/8a85a4fe01d6e3adfae3d886d77832be8726b3c5/connexion/__init__.py#L21) or [HuggingFace](https://discuss.huggingface.co/t/adding-a-new-model-to-transformers-with-additional-dependencies/916).

express/utils.py
```diff
@@ -0,0 +1,30 @@
...
+"""
+import logging
+import importlib.util
+import importlib.metadata
+from typing import Sequence, Union, List
+
+logger = logging.getLogger(__name__)
+
+
+def is_module_available(module_name: Union[str, Sequence[str]]) -> List[str]:
```
NielsRogge, line 12: Maybe we can have a single utils folder in express that includes:
- kfp_utils.py
- import_utils.py which includes the soft dependency checkers. This can be done similar to how it's done here: https://github.com/huggingface/transformers/blob/main/src/transformers/utils/import_utils.py

PhilippeMoussalli: > Thanks @PhilippeMoussalli
> 
> This should be an internal function though, so it should not be necessary to use it in the examples. The reason you need this, is because you are still raising an `ImportError` at import time, just with a clearer message. We should instead delay this error until the dependency is actually used.
> 
> See the way it is done in [Connexion](https://github.com/spec-first/connexion/blob/8a85a4fe01d6e3adfae3d886d77832be8726b3c5/connexion/__init__.py#L21) or [HuggingFace](https://discuss.huggingface.co/t/adding-a-new-model-to-transformers-with-additional-dependencies/916).

Thanks for reviewing. Indeed it makes more sense to implement it in the express package. Discussed with @NielsRogge  and decided to go for the current approach: 

- Soft dependencies include HF dataset and pandas. If they are not installed, the use will be asked to install them with pip. Reason we didn't include them in `pyproject.toml` is that it's not worth defining an extra for one packages and both pandas and HF dataset should support backwards compatibility. 
- The other KFP dependencies which are just necessary for compiling and uploading pipelines to KFP are now in the  `express/pipelines` repo under a `requirements.txt`.

Also made some small changes to the functions that check for package availability to make them similar to HF

tests/test_kfp_utils.py
```diff
@@ -1,20 +0,0 @@
-"""
```
NielsRogge, line 1: Any reason this test file gets deleted?

PhilippeMoussalli: re-added a new one, delted this one since it's no longer existant in kfp_utils

examples/pipelines/pipeline_utils.py
```diff
@@ -1,37 +1,15 @@
-"""General kfp helpers"""
+"""General pipeline utils"""
```
NielsRogge, line 1: Any reason the pipeline utils moved out of Express to here?

PhilippeMoussalli: the reason is that we wanted to move the `kfp` dependencies out of express and have it only in the `requirements.txt` the `example/pipelines` level

RobbeSneyders:
Thanks @PhilippeMoussalli, looks good to me.

Final step is to update the `requirements.txt` in the examples I think?

PhilippeMoussalli:
> Thanks @PhilippeMoussalli, looks good to me.
> 
> Final step is to update the `requirements.txt` in the examples I think?

I removed the `requirements.txt` from there, currently if you want to compile any of the pipelines in the examples and if you don't have the dependencies installed it will throw an error to install them with the `pipelines` extra (similar to how hf_dataset and pandas). This happens mainly when trying to import the `pipeline_utils` from `express`.

Any reason to include them there as well?

RobbeSneyders:
I think it would be good if the examples are complete and runnable. Showing how to define the necessary dependencies seems useful to me.

RobbeSneyders:
Oops, I merged #18 without noticing it targeted this branch instead of main. Let me work some git magic to fix this.

PhilippeMoussalli:
> I think it would be good if the examples are complete and runnable. Showing how to define the necessary dependencies seems useful to me.

I see, I added them again with `pandas` and `datasets` since they're also included in `examples`

examples/requirements.txt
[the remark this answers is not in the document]
PhilippeMoussalli: done :)

examples/components/single_node/README.md
```diff
@@ -3,26 +3,7 @@
...
-Run the following command to load a Hugging Face Dataset from the [hub](https://huggingface.co/) and upload it to a Google Cloud Storage bucket:
-
-```
-python3 data_loading_hf_datasets.py --extra-args '{"project_id": "soy-audio-379412"}' \
- --output-manifest <local_path> \
-  --metadata-args '{"run_id":"test","component_name":"test_component", \
-  "artifact_bucket":"soy-audio-379412_kfp-artifacts"}'
-```
\ No newline at end of file
+* Huffingface datasets
```
NielsRogge, line 8:
```suggestion
* Huggingface datasets
```
;)

examples/components/single_node/README.md
```diff
@@ -3,26 +3,7 @@
 This folder contains an example script showcasing data loading with Express on a single node. Make sure to install express
 with the `poetry install` command from the root of the repository.
 
-# Usage
+# Current supported framework
```
NielsRogge, line 6:
```suggestion
# Current supported frameworks
```

tests/test_kfp_utils.py
NielsRogge: Hmm I don't see the new one

PR #18, opened 2023-03-17
Simplify component build

PhilippeMoussalli:
- PR to add a single build image script that builds all the images for a certain pipeline (or a single image if a flag is passed)

- change registry to a public artifact registry on docker hub (cloud agnostic image build) + add missing metadata to manifest and
- Add missing metadata to docker variables and fetch them when creating the manifest 

Still TODO: Decide on a public docker repo to push the images (right now it's on my personal one)

Ensure that all future components dockerfile have the following commands to register the metadata: 

```
ARG GIT_BRANCH
ARG COMMIT_SHA
ARG BUILD_TIMESTAMP
ENV GIT_BRANCH $GIT_BRANCH
ENV COMMIT_SHA $COMMIT_SHA
ENV BUILD_TIMESTAMP $BUILD_TIMESTAMP
```


RobbeSneyders:
Thanks @PhilippeMoussalli.

I would propose to use Github for the docker images, what do you think?

NielsRogge (approved):
Great, thanks for adding this. Will then also leverage this for #16

PhilippeMoussalli:
> Thanks @PhilippeMoussalli.
> 
> I would propose to use Github for the docker images, what do you think?

You mean publishing to the container registry (under packages)  of the `ml6-team` org account? That might make it easier indeed to integrate later on with Github workflows. 

https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry

PhilippeMoussalli:
> Thanks @PhilippeMoussalli.
> 
> I would propose to use Github for the docker images, what do you think?

@RobbeSneyders I did the switch over 
Put some notes in a doc on the setup instructions 
https://docs.google.com/document/d/1ZaTEui_n3TBb4Ex8eueENK0sk9hq0Q7va2WXNvgun88/edit

examples/pipelines/finetune_stable_diffusion/build_images.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
+source ../docker_build.config
```
RobbeSneyders, line 3: The IMAGE_TAG should probably be a parameter of the script, right?

I'm more in favor of adding these config variables as parameters and just setting a default. Eg:
```
while [[ "$#" -gt 0 ]]; do case $1 in
  -n|--namespace) namespace="$2"; shift;;
  -r|--repo) repo="$2"; shift;;
  -t|--tag) tag="$2"; shift;;
  *) echo "Unknown parameter passed: $1"; exit 1;;
esac; shift; done

[ -n "${namespace-}" ] || namespace="ml6team"
[ -n "${repo-}" ] || repo="express"
[ -n "${tag-}" ] || tag="latest"
```

PhilippeMoussalli: Seems like a much neater solution than having a separate file for the default configs. Thanks for the suggestion

RobbeSneyders (approved):
Thanks @PhilippeMoussalli!

NielsRogge:
Thanks @PhilippeMoussalli ! As a follow-up, we can update the HF dataset pipeline as well to leverage a single build_image.sh script: https://github.com/ml6team/express/tree/main/examples/pipelines/hf_dataset_pipeline

PR #19, opened 2023-03-20
Add function to log cuda devices

PhilippeMoussalli:
PR that adds a lightweight function to check for CUDA device availability using a lightweight library https://github.com/gpuopenanalytics/pynvml. The advantage is avoiding to install heavy libraries in express such as torch or TF

### DIFF
--- a/examples/pipelines/finetune_stable_diffusion/components/image_filter_component/greeting.sh
+++ b/examples/pipelines/finetune_stable_diffusion/components/image_filter_component/greeting.sh
@@ -1,3 +1,0 @@
-#!/bin/sh
-
-echo Hello $name
--- a/examples/pipelines/finetune_stable_diffusion/components/image_filter_component/greeting.sh
+++ b/examples/pipelines/finetune_stable_diffusion/components/image_filter_component/greeting.sh
@@ -1,0 +1,3 @@
+#!/bin/sh
+
+echo Hello $name
--- a/express/utils.py
+++ b/express/utils.py
@@ -6,6 +6,8 @@ import importlib.util
 import importlib.metadata
 from typing import Sequence, Union, List
 
+import pynvml
+
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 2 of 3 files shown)


RobbeSneyders:
Thanks @PhilippeMoussalli 

Why do we need this function? It only logs information, so I'm not sure if it's worth adding a dependency.

PhilippeMoussalli:
> Thanks @PhilippeMoussalli
> 
> Why do we need this function? It only logs information, so I'm not sure if it's worth adding a dependency.

This will be used at the beginning of every component as a sanity check if there is a GPU device running (check if GPU drivers are installed properly on the underlying nodes running the job). It will be probably be repeated across different component. 

Alternatively we can create a soft dependency check for it (similar to pandas, HF datasets) and then include in the `requirements.txt` of the components that will use it. Wdyt?

RobbeSneyders:
But it only logs information, right? Shouldn't it either return a value or raise an error in some cases?

If we just want to check if an Nvidia GPU is available, we can do that without a dependency as well:

```
import subprocess

try:
    subprocess.check_output('nvidia-smi')
    print('Nvidia GPU detected!')
except Exception: # this command not being found can raise quite a few different errors depending on the configuration
    print('No Nvidia GPU in system!')
```

PhilippeMoussalli:
> But it only logs information, right? Shouldn't it either return a value or raise an error in some cases?
> 
> If we just want to check if an Nvidia GPU is available, we can do that without a dependency as well:
> 
> ```
> import subprocess
> 
> try:
>     subprocess.check_output('nvidia-smi')
>     print('Nvidia GPU detected!')
> except Exception: # this command not being found can raise quite a few different errors depending on the configuration
>     print('No Nvidia GPU in system!')
> ```

makes sense! Thank you for the suggestion. Implemented :)

examples/pipelines/finetune_stable_diffusion/components/image_filter_component/greeting.sh
RobbeSneyders, line 1: I don't think we need this file?

express/utils.py
```diff
@@ -28,3 +30,13 @@ def is_module_available(module_name: Union[str, Sequence[str]]) -> List[str]:
...
     return missing_modules
+
+
+def get_cuda_availability():
+    """Function that checks for cuda device availability"""
+    try:
+        output = subprocess.check_output("nvidia-smi")  # nosec
+        logger.info("Nvidia GPU detected!")
+        logger.info(output.decode("utf-8"))
+    except FileNotFoundError:
```
RobbeSneyders, line 41: Is this the only error that can be raised?

RobbeSneyders (commented):
Great that we could remove the dependency @PhilippeMoussalli, but my original question still stands. Why do we need this function if it does not do or return anything except for logging some info about the GPU.

examples/pipelines/finetune_stable_diffusion/components/image_filter_component/greeting.sh
PhilippeMoussalli: no indeed i'll remove it

express/utils.py
PhilippeMoussalli: The two cases that I tested for was if we have the 

- nvidia drivers installed + no GPU
- nvidia drivers not installed + no GPU

and they both give the same error. I think those might be the only cases we have.

RobbeSneyders:
> Mainly since it's going to be repeated across components that have GPU. A bit similar to the functions that we have that check for soft dependencies. Alternatively we can just add it at the try except at beginning of each component if you see it more fit.

My main issue is that it only logs things. It does not return a value or raise an error, so I don't see how this is useful.

PhilippeMoussalli:
> > Mainly since it's going to be repeated across components that have GPU. A bit similar to the functions that we have that check for soft dependencies. Alternatively we can just add it at the try except at beginning of each component if you see it more fit.
> 
> My main issue is that it only logs things. It does not return a value or raise an error, so I don't see how this is useful.

I see your point, I think then we'll just include them in the components themselves. I'll close the PR.

PR #22, opened 2023-03-30
Add sanity check

NielsRogge:
This PR adds logic to verify that, when creating an output manifest, all data sources have an index and share data for (at least) the same index. Each framework (like HuggingFace datassets, Pandas, etc) needs to implement this.

### DIFF
--- a/express/components/hf_datasets_components.py
+++ b/express/components/hf_datasets_components.py
@@ -21,9 +21,33 @@ if is_datasets_available():
     import datasets
     from datasets import load_dataset
 
+
 # Define interface of pandas draft
 # pylint: disable=unsubscriptable-object
-HFDatasetsDatasetDraft = ExpressDatasetDraft[List[str], datasets.Dataset]
+class HFDatasetsDatasetDraft(ExpressDatasetDraft[List[str], datasets.Dataset]):
+    "HuggingFace Datasets dataset draft (plan for an output manifest)"
+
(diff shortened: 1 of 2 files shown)


express/components/hf_datasets_components.py
```diff
@@ -21,9 +21,35 @@
...
 # Define interface of pandas draft
 # pylint: disable=unsubscriptable-object
-HFDatasetsDatasetDraft = ExpressDatasetDraft[List[str], datasets.Dataset]
+class HFDatasetsDatasetDraft(ExpressDatasetDraft[List[str], datasets.Dataset]):
+    "HuggingFace Datasets dataset draft (plan for an output manifest)"
+
+    def _is_sublist(self, list1, list2):
+        return set(list1) <= set(list2)
+
+    def sanity_check(
```
PhilippeMoussalli, line 33: Could you add a more descriptive name for the function?

PhilippeMoussalli (commented):
Thanks for adding this! 

Left a few comments.
Do you think it would be possible to add a testing functionality for this?

RobbeSneyders:
I would like to decide on the user interface first. If we go for a single dataframe as proposed, I don't think we need this functionality, or it will at least work differently.

NielsRogge:
Closing this PR for now as it will depend on the final API of the Fondant framework.

PR #23, opened 2023-03-31
Feature/add pandas example pipeline 

PhilippeMoussalli:
PR that adds a simple pipeline showcasing data loading and filtering with pandas + includes some fixes to Express from Niels from this [PR](https://github.com/ml6team/express/pull/20)

### DIFF
--- a/examples/pipelines/pandas_pipeline/components/filter_images/Dockerfile
+++ b/examples/pipelines/pandas_pipeline/components/filter_images/Dockerfile
@@ -1,5 +1,12 @@
 FROM python:3.8-slim
 
+ARG GIT_BRANCH
+ARG COMMIT_SHA
+ARG BUILD_TIMESTAMP
+ENV GIT_BRANCH $GIT_BRANCH
+ENV COMMIT_SHA $COMMIT_SHA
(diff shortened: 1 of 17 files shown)


examples/pipelines/pandas_pipeline/components/filter_images/Dockerfile
```diff
@@ -0,0 +1,36 @@
+FROM python:3.8-slim
```
NielsRogge, line 1: Should we add `--platform=linux/arm64` here? To make sure those images can also be built from a mac

examples/pipelines/pandas_pipeline/components/filter_images/component.yaml
```diff
@@ -0,0 +1,29 @@
+name: filter images
```
NielsRogge, line 1:
```suggestion
name: filter_images
```

examples/pipelines/pandas_pipeline/components/filter_images/Dockerfile
PhilippeMoussalli: can we add add this in the bash script of `build_images.sh` as a flag during build? I saw you had this in one your PRs. Then we don't have to remember about including it in every base image.

PR #24, opened 2023-04-03
Feature/simplify metadata argument passing

PhilippeMoussalli:
PR that simplifies the way the way that components arguments are built and passed to the components. 

The previous implementation relied on the user providing both the component arguments and the metadata arguments. The latter is something that the end-user does not need to take care off. It also makes the arguments passed to each component more straightforward (args, manifest input, mainfest output).

**Note:** the PR branches off this [PR](https://github.com/ml6team/express/pull/21)

### DIFF
--- a/examples/components/single_node/hf_datasets/data_loading_hf_datasets.py
+++ b/examples/components/single_node/hf_datasets/data_loading_hf_datasets.py
@@ -1,9 +1,10 @@
-from typing import Optional, Dict, Union
+from typing import Dict, Union
 
 import pandas as pd
 from datasets import Dataset, load_dataset, concatenate_datasets
 
-from express.components.hf_datasets_components import HFDatasetsLoaderComponent, HFDatasetsDatasetDraft
+from express.components.hf_datasets_components import HFDatasetsLoaderComponent, \
+    HFDatasetsDatasetDraft
 from express.logger import configure_logging
 
 
@@ -12,18 +13,19 @@ class SeedDatasetLoader(HFDatasetsLoaderComponent):
     """Class that inherits from Hugging Face data loading """
 
     @classmethod
-    def load(cls, extra_args: Optional[
-        Dict[str, Union[str, int, float, bool]]] = None) -> HFDatasetsDatasetDraft:
+    def load(cls,
+             args: Dict[str, Union[str, int, float, bool]] = None) -> HFDatasetsDatasetDraft:
         """
         An example function showcasing the data loader component using Express functionalities
         Args:
-            extra_args (Optional[Dict[str, Union[str, int, float, bool]]): optional args to pass to
-             the function (e.g. seed data source)
+            args ([Dict[str, Union[str, int, float, bool]]): Component arguments passed as a
+             json dict string
         Returns:
-            HFDatasetsDatasetDraft: a dataset draft that creates a plan for an output datasets/manifest
+            HFDatasetsDatasetDraft: a dataset draft that creates a plan for an output
+             datasets/manifest
         """
         configure_logging()
-        
+
         # 1) Create example data source(s)
         caption_dataset = load_dataset("lambdalabs/pokemon-blip-captions", split="train")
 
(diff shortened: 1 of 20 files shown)


examples/components/single_node/hf_datasets/data_loading_hf_datasets.py
```diff
@@ -12,18 +13,19 @@ class SeedDatasetLoader(HFDatasetsLoaderComponent):
...
-        Dict[str, Union[str, int, float, bool]]] = None) -> HFDatasetsDatasetDraft:
+    def load(cls,
+             args: Dict[str, Union[str, int, float, bool]] = None) -> HFDatasetsDatasetDraft:
         """
         An example function showcasing the data loader component using Express functionalities
         Args:
-            extra_args (Optional[Dict[str, Union[str, int, float, bool]]): optional args to pass to
-             the function (e.g. seed data source)
+            args ([Dict[str, Union[str, int, float, bool]]): Component arguments passed as a
+             json dict string
```
NielsRogge, line 22: I guess we can't get around the "json dict string", which is a bit weird

express/components/common.py
```diff
@@ -328,21 +317,22 @@ class ExpressTransformComponent(ExpressDatasetHandler, Generic[IndexT, DataT]):
...
             DataManifest: the output manifest
         """
-        args = cls._parse_args()
-        input_dataset = cls._load_dataset(
-            input_manifest=DataManifest.from_path(args.input_manifest)
-        )
-        output_dataset_draft = cls.transform(
-            data=input_dataset, extra_args=json.loads(args.extra_args)
+        parsed_args = cls._parse_args()
+        args_dict = json.loads(parsed_args.args)
```
NielsRogge, line 324: Maybe the json.loads can be directly in the `_parse_args` method?

express/components/common.py
```diff
@@ -418,17 +402,17 @@ class ExpressLoaderComponent(ExpressDatasetHandler, Generic[IndexT, DataT]):
...
         Parses input data, executes the data loader, and creates output artifacts.
-
         Returns:
             DataManifest: the output manifest
         """
-        args = cls._parse_args()
-        output_dataset_draft = cls.load(extra_args=json.loads(args.extra_args))
-        metadata = cls._create_metadata(metadata_args=json.loads(args.metadata_args))
+        parsed_args = cls._parse_args()
+        args_dict = json.loads(parsed_args.args)
```
NielsRogge, line 409: Same comment here

examples/components/single_node/pandas/README.MD
```diff
@@ -5,8 +5,7 @@
 Run the following command to upload a Pandas dataframe to a Google Cloud Storage bucket:
 
 ```
-python3 data_loading_pandas.py --extra-args '{"project_id": "storied-landing-366912"}' \
- --output-manifest <local_path> \
-  --metadata-args '{"run_id":"test","component_name":"test_component", \
-  "artifact_bucket":"storied-landing-366912-kfp-output"}'
+--args '{"project_id": "soy-audio-379412", "run_id":"test","component_name":"test_component", \
```
NielsRogge, line 8:
```suggestion
--args '{"project_id": "soy-audio-379412", "run_id":"test", "component_name":"test_component", \
```

NielsRogge: Isn't it a bit weird that we need to pass the name of the component to the component itself?

NielsRogge: Run_id and project_id makes sense as metadata as those are different each time you run a component, but the name of the component seems inherent to the component itself

NielsRogge (approved):
Thanks for improving!

NielsRogge (changes requested):
On second thought, it might be better to expose the individual arguments that a component expects in its specification. By just accepting `args` it may be unclear that a component for instance requires `model_id` and `batch_size`. 

It might be better that the user specifies in the function (as arguments) or in the class (as attributes) which arguments are required, which are then automatically turned into CLI arguments (a bit like how Google's [Fire](https://github.com/google/python-fire) works).

examples/pipelines/hf_dataset_pipeline/components/add_captions/component.yaml
```diff
@@ -1,14 +1,10 @@
-name: load_from_hub
+name: add_captions
 description: A basic component that takes a dataset name from the 🤗 hub as input and uploads it to a GCS bucket.
 inputs: 
-    - name: extra_args
-      description: Additional arguments passed to the component, as a json dict string
+    - name: args
```
RobbeSneyders, line 4: Why do we pass these arguments as json instead of as separate arguments?

NielsRogge:
PR #27 updates the way metadata is updated, so it depends on the review of that PR regarding how we can proceed

PR #25, opened 2023-04-04
[WIP] Add general taxonomy

NielsRogge:
From the dataset doc:

> Fondant should define a taxonomy of general data types that components can use. We should try to cover as many general cases as we can without making it too complex.

This is a first draft of how this could look like. Basically, each component would need to define expected input and output data sources (they are still called data sources in the code base). Fondant will then check whether those are met based on the input and output manifests.

### DIFF
--- a/examples/pipelines/finetune_stable_diffusion/components/embedding_component/src/main.py
+++ b/examples/pipelines/finetune_stable_diffusion/components/embedding_component/src/main.py
@@ -4,7 +4,7 @@ This component adds a data source to the manifest by embedding the images.
 import logging
 from typing import Optional, Union, Dict
 
-
+import express
 from express.components.hf_datasets_components import (
     HFDatasetsTransformComponent,
     HFDatasetsDataset,
@@ -51,6 +51,8 @@ class EmbeddingComponent(HFDatasetsTransformComponent):
     """
     Component that embeds the images using a CLIP model from Hugging Face.
     """
+    data_sources_in = [("images", express.Image)]
+    data_sources_out = [("embeddings", express.Text)]
 
     @classmethod
     def transform(
--- a/examples/pipelines/finetune_stable_diffusion/components/embedding_component/src/main.py
+++ b/examples/pipelines/finetune_stable_diffusion/components/embedding_component/src/main.py
@@ -52,7 +52,7 @@ class EmbeddingComponent(HFDatasetsTransformComponent):
     Component that embeds the images using a CLIP model from Hugging Face.
     """
     data_sources_in = [("images", express.Image)]
-    data_sources_out = [("embeddings", express.Text)]
+    data_sources_out = [("embeddings", express.Vector)]
 
     @classmethod
     def transform(
--- a/examples/pipelines/finetune_stable_diffusion/dataset_creation_pipeline.py
+++ b/examples/pipelines/finetune_stable_diffusion/dataset_creation_pipeline.py
@@ -12,7 +12,6 @@ from config.components_config import (
     LoadFromHubConfig,
     ImageFilterConfig,
     EmbeddingConfig,
-    ClipRetrievalConfig,
 )
 from express.pipeline_utils import compile_and_upload_pipeline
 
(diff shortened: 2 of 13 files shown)


express/components/common.py
```diff
@@ -321,6 +331,34 @@ def _create_output_dataset(
...
+        for expected_data_source in cls.data_sources_in:
+            # get the data source from the manifest
+            if expected_data_source[0] not in data.manifest.data_sources:
+                raise ValueError(
+                    f"Input dataset is missing data source {expected_data_source}"
+                )
+            data.manifest.data_sources[expected_data_source[0]]
+            # TODO verify that the required fields are present
+            # according to the taxonomy expected_data_source[1]
+            continue
```
NielsRogge, line 344: This is the main part to be filled in, based on the taxonomy we define for each modality

examples/pipelines/finetune_stable_diffusion/components/embedding_component/src/main.py
```diff
@@ -51,6 +51,8 @@ class EmbeddingComponent(HFDatasetsTransformComponent):
     """
     Component that embeds the images using a CLIP model from Hugging Face.
     """
+    data_sources_in = [("images", express.Image)]
```
RobbeSneyders, line 54: I think we should define this as config somewhere, so this is easily accessible for other classes and external tools.

examples/pipelines/finetune_stable_diffusion/dataset_creation_pipeline.py
```diff
@@ -68,25 +67,6 @@
 embedding_metadata_args = json.dumps(embedding_metadata_args)
 
 
-# Component 4
```
RobbeSneyders, line 71: These are removed since they are not ported yet I assume?

express/components/common.py
```diff
@@ -192,6 +192,16 @@ class ExpressDatasetHandler(ABC, Generic[IndexT, DataT]):
      Pandas DataFrame or a Spark RDD.
     """
 
+    @property
+    @abstractmethod
+    def data_sources_in(self):
```
RobbeSneyders, line 197: I would expect these attributes on a class which represents the component specification.

express/components/common.py
```diff
@@ -341,10 +391,17 @@ def run(cls) -> DataManifest:
         input_dataset = cls._load_dataset(
             input_manifest=DataManifest.from_path(args.input_manifest)
         )
+        # verify incoming data sources
+        cls._verify_data_sources_in(input_dataset)
```
RobbeSneyders, line 395: I would like to see functionality to load a dataset based on a manifest and component specification instead. Where the component specification is used to decide which modalities and columns need to be loaded from the manifest, and the manifest is used to get the path to the actual data.

express/components/common.py
```diff
@@ -429,9 +486,15 @@ def run(cls) -> DataManifest:
             DataManifest: the output manifest
         """
         args = cls._parse_args()
+        if cls.data_sources_in is not None:
+            raise ValueError("Loader components should not have input data sources.")
         output_dataset_draft = cls.load(extra_args=json.loads(args.extra_args))
+        print("Output draft:", output_dataset_draft)
+        # verify outgoing data sources
+        cls._verify_data_sources_out(draft=output_dataset_draft)
```
RobbeSneyders, line 494: I think we still need a better view on how we want to convert the user dataframe to a new manifest, which is still an open topic in the dataset doc.

express/taxonomy.py
```diff
@@ -0,0 +1,40 @@
...
+class Image:
+    """
+    A single image
+    """
+
+    def __init__(self):
+        pass
+
+    def required_columns(self):
+        return ["width", "height"]
```
RobbeSneyders, line 13: We need info on the data types here as well.

RobbeSneyders (commented):
Thanks @NielsRogge 

I left some comments, but I think we need to get some more clarity on the design first before continuing with the implementation.

RobbeSneyders:
Closing this since we'll start without a taxonomy.

PR #26, opened 2023-04-06
Mount remote storage with FUSE

PhilippeMoussalli:
This PR replaces the current way of interacting with cloud storage. It uses FUSE to mount the remote cloud storage as a local file system. This enables us to read/write files directly from files stored remotely instead of downloading them to storage. 

Check out this [document](https://docs.google.com/document/d/1Yn8ayLBovv2PHlLQhsxp81Wzp8I23pOjW4xlMkqwmjs/edit ) for more description on FUSE and considered alternatives 

ToDos: 

- [X ] Add a Dockerfile for installation of fuse 
- [ X] Read/write files directly from the mount direction (right now they are being written to a local directory and copied over to the mount location) 
- [ X] Add necessary changes to the HF dataset component 
- [ ] Migrate changes to existing pipeline 

Streaming functionality will be implemented in a separate PR

### DIFF
--- a/docker/gcsfuse/Dockerfile
+++ b/docker/gcsfuse/Dockerfile
@@ -1,23 +1,40 @@
-FROM python:3.8-slim
+# Start with a base Ubuntu image
+FROM ubuntu:latest
 
-ENV DEBIAN_FRONTEND=noninteractive
+# Install necessary packages
+RUN apt-get update && apt-get install -y \
+    apt-transport-https \
+    ca-certificates \
+    curl \
+    gnupg \
+    fuse \
+    git \
+    libfuse-dev \
+    libssl-dev \
+    pkg-config \
+    wget
 
-## System dependencies
-# Install system dependencies
-RUN set -e; \
-    apt-get update -y && apt-get install curl gnupg2 -y \
-    lsb-release; \
-    gcsFuseRepo=gcsfuse-`lsb_release -c -s`; \
-    echo "deb http://packages.cloud.google.com/apt $gcsFuseRepo main" | \
-    tee /etc/apt/sources.list.d/gcsfuse.list; \
-    curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | \
-    apt-key add -; \
-    apt-get update; \
-    apt-get install -y gcsfuse \
-    && apt-get clean
+# Install GCSFuse
+RUN echo "deb http://packages.cloud.google.com/apt gcsfuse-bionic main" > /etc/apt/sources.list.d/gcsfuse.list && \
+    curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key add - && \
+    apt-get update && \
+    apt-get install -y gcsfuse
 
-ENV APP_HOME /app
-WORKDIR $APP_HOME
+## Install S3FS
+RUN echo "deb http://deb.debian.org/debian buster-backports main" | tee /etc/apt/sources.list.d/backports.list && \
+    apt-get update && \
+    apt-get -t buster-backports install -y s3fs
 
-CMD ["/bin/bash"]
+## Install BlobFuse
+RUN wget https://packages.microsoft.com/config/ubuntu/20.04/packages-microsoft-prod.deb && \
+    dpkg -i packages-microsoft-prod.deb && \
+    apt-get update && \
+    apt-get install -y blobfuse
 
+## Clean up
+RUN apt-get autoremove -y && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
+
+# Set the entrypoint
+ENTRYPOINT ["/bin/bash"]


docker/gcsfuse/Dockerfile
RobbeSneyders, line 1: How should this Dockerfile be used? As a base image for components?

express/storage_mount.py
```diff
@@ -0,0 +1,102 @@
...
+    mount_args: Optional[List[str]] = None,
+    mount_kwargs: Optional[Dict[str, any]] = None,
+):
+    """
+    Function that mounts a remote bucket to a local directory using FUSE
+    Args:
+        mount_buckets (str): the name of the bucket(s) to be mount
+        mount_dir (str): the local directory where to mount the buckets to
+        mount_command (str): the command used to mount the storage system
+        mount_args (Optional[List[str]]): Additional arguments to pass to the FUSE command.
```
RobbeSneyders, line 56: Which args and kwargs do we expect the user might want to pass here?

express/io.py
```diff
@@ -31,3 +49,38 @@ def get_file_name(file_uri: str, return_extension=False):
     if return_extension:
         file_name = f"{file_name}{extension}"
     return file_name
+
+
+def create_subprocess_arguments(
+    args: Optional[List[str]] = None, kwargs: Optional[Dict[str, Any]] = None
+) -> List[str]:
```
RobbeSneyders, line 56: I feel like we're taking quite some assumptions here on how arguments will be parsed, which can often differ between programs. If we want the user to be able to pass in arguments, I would let them pass them in already formatted.

express/components/common.py
```diff
@@ -428,15 +485,43 @@ def run(cls) -> DataManifest:
...
-        output_dataset_draft = cls.load(extra_args=json.loads(args.extra_args))
-        metadata = cls._create_metadata(metadata_args=json.loads(args.metadata_args))
-        # Create metadata
+        storage_args = json.loads(args.storage_args)
+        extra_args = json.loads(args.extra_args)
+        metadata_args = json.loads(args.metadata_args)
+
+        # Get specific cloud storage configs
+        cloud_storage = CloudProvider[storage_args["cloud_env"]].value
+        storage_mount_cmd, storage_prefix = (
```
RobbeSneyders, line 496: Same here.

RobbeSneyders (commented):
Thanks @PhilippeMoussalli 

High level, I think this is a good approach. I left some comments in line.

Some general comments:
- Let's leverage `pathlib` everywhere instead of `os.path` and accept `Path` type where relevant.
- Let's leverage keyword only arguments whenever useful: https://blog.ml6.eu/designing-better-python-interfaces-with-keyword-only-arguments-61888040d971
- Remove typing from the docstrings to have a single source of truth

We can tackle these in a separate PR across the whole codebase as well.

express/storage_mount.py
PhilippeMoussalli: There are quite a few extra arguments that can be passed to the fuse commands (e.g. passing in keys, enable additional logging, enable debugging). And I think they are all different depending on which cloud you use for fuse. We can maybe restrict the options once we figure out how we want to handle the authentication and enable other ones with extra argument flags that we pass in the function

RobbeSneyders: Question is which ones we want to expose to the user and which one we want to configure.

docker/gcsfuse/Dockerfile
PhilippeMoussalli: I think we should have a base image component for Express that contains common system dependencies. In this case, it should be all the Fuse installations for the cloud providers. All the other images will build on top of this image.

Do you agree with this approach?

express/components/common.py
```diff
@@ -72,24 +74,30 @@ def load(
...
+            **kwargs
         )
 
     @staticmethod
     @abstractmethod
     def _load_data_source(
-        data_source: DataSource,
-        index_filter: Optional[IndexT],
-        **kwargs,
+            *,
```
NielsRogge, line 86: Same comment here

docker/gcsfuse/Dockerfile
NielsRogge: I think that's a good idea

express/components/common.py
```diff
@@ -43,23 +43,25 @@ def extend(self) -> "ExpressDatasetDraft[IndexT, DataT]":
...
+    def load_index(self, mount_dir: str) -> IndexT:
         """
         Loads the index data.
+        Args:
+            mount_dir: the local directory mounted with fuse
         """
 
     def load(
-        self, data_source: str, for_index: Optional[IndexT] = None, **kwargs
+            self, data_source: str, *, for_index: Optional[IndexT] = None, **kwargs
```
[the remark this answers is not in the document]
RobbeSneyders: Why does this make it harder to debug? Keyword-only arguments add more clarity to the code base. I wrote a blog post on this which I linked above:
https://blog.ml6.eu/designing-better-python-interfaces-with-keyword-only-arguments-61888040d971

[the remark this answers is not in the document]
RobbeSneyders: Yes indeed.

NielsRogge (approved):
PR looks good to me, but maybe it might be easier to integrate after #27 gets merged

PR #28, opened 2023-04-11
Update manifest class to match discussed specification

RobbeSneyders:
This PR updates the manifest class to match the discussion on the dataset doc. Let's focus all further discussion on the manifest in this PR.

- I included a schema defining the manifest
- Data sources has been renamed to subsets
- The manifest now defines which fields / columns are available in a subset and their types
- The location for each subset is resolved with the base path defined in the metadata
- I'm no longer leveraging `dataclasses-json` as I don't just want to build a 1 to 1 mapping of the json in Python
- All unused keys have been removed from the manifest

The `Manifest` class is currently mostly focused on making an existing json manifest easily accessible, which is useful for input manifests. We might want to add functionality to create a new manifest programmatically as well, which is useful for output manifests. It might be best that these are two separate classes though, an immutable one for the input manifests and a mutable one for the output manifests.

I'd like to see a similar PR for the component specification as well.

### DIFF
--- a/express/manifest.py
+++ b/express/manifest.py
@@ -1,112 +1,113 @@
-"""
-Script for generating the dataset manifest that will be passed and updated through different
-components of the pipeline
-"""
+"""This module defines classes to represent an Express manifest."""
+
 import json
-from dataclasses import dataclass, field
-from enum import Enum
-from typing import List, Dict
+import pkgutil
+import typing as t
+from dataclasses import dataclass
 
-from dataclasses_json import dataclass_json
+import jsonschema.exceptions
+from jsonschema import Draft4Validator
+
+from express.exceptions import InvalidManifest
 
 
-class DataType(str, Enum):
+@dataclass
+class Field:
+    """Class representing a single field or column in an Express subset."""
+
+    name: str
+    type: str
+
+
+class Subset:
     """
-    Supported types for stored data.
+    Class representing an Express subset.
+
+    Args:
+        specification: The part of the manifest json representing the subset
+        base_path: The base path which the subset location is defined relative to
     """
 
-    PARQUET = "parquet"
-    BLOB = "blob"
+    def __init__(self, specification: dict, *, base_path: str) -> None:
+        self._specification = specification
+        self._base_path = base_path
+
+    @property
+    def location(self) -> str:
+        """The resolved location of the subset"""
+        return self._base_path.rstrip("/") + self._specification["location"]
+
+    @property
+    def fields(self) -> t.Dict[str, Field]:
+        return {
+            name: Field(name=name, type=field["type"])
+            for name, field in self._specification["fields"].items()
+        }
+
+
+class Index(Subset):
+    """Special case of a subset for the index, which has fixed fields"""
+
+    @property
+    def fields(self) -> t.Dict[str, Field]:
+        return {
+            "id": Field(name="id", type="str"),
+            "source": Field(name="source", type="str"),
+        }
+
+
+class Manifest:
+    """
+    Class representing an Express manifest
+
+    Args:
+        specification: The manifest specification as a Python dict
+    """
+
+    def __init__(self, specification: dict) -> None:
+        self._validate_spec(specification)
+        self._specification = specification
 
     @classmethod
-    def is_valid(cls, data_type: str) -> bool:
-        """Check if data type is valid"""
-        return data_type in cls.__members__.values()
+    def _validate_spec(cls, spec: dict) -> None:
+        """Validate a manifest specification against the manifest schema
 
-
-@dataclass
-class DataSource:
-    """
-    Information about the location and contents of a single data source.
-    Args:
-        location (str): the URI of the data root
-        type (Union[DataType, str]): the type of data, which determines how it's interpreted
-        extensions (List[str]): under what file extensions the data is stored.
-        n_files (int): how many files make up the data
-        n_items (int): how many items (e.g. distinct images, captions, etc.) are covered by the
-         data.
-    """
-
-    location: str
-    type: DataType
-    extensions: List[str]
-    n_files: int = field(default_factory=int)
-    n_items: int = field(default_factory=int)
-
-
-@dataclass_json
-@dataclass
-class Metadata:
-    """
-    The metadata associated with the manifest
-    Args:
-        artifact_bucket (str): remote location of all newly created job artifacts.
-        run_id (str): the kfp run id associated with the manifest (kfp.dsl.EXECUTION_ID_PLACEHOLDER)
-        component_id (str): if this metadata is passed as an input to a component, this is the id of
-         the current component. If this metadata is stored as part of a manifest, this is the id of
-        the component that generated the manifest.
-        component_name (str): name of the current or originating component (see component_id for
-         more details).
-        branch (str): the git branch associated with that manifest
-        commit_hash (str): the commit hash associated with that manifest
-        creation_date (str): the creation date of the manifest
-        num_items (int): total number of rows in the index
-    """
-
-    # TODO: get rid of defaults
-    artifact_bucket: str = field(default_factory=str)
-    run_id: str = field(default_factory=str)
-    component_id: str = field(default_factory=str)
-    component_name: str = field(default_factory=str)
-    branch: str = field(default_factory=str)
-    commit_hash: str = field(default_factory=str)
-    creation_date: str = field(default_factory=str)
-    num_items: int = field(default_factory=int)
-
-
-@dataclass_json
-@dataclass
-class DataManifest:
-    """
-    The dataset Manifest
-    Args:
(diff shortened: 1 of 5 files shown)


express/schemas/manifest.json
```diff
@@ -0,0 +1,75 @@
...
+    "index",
+    "subsets"
+  ],
+  "definitions": {
+    "field": {
+      "type": "object",
+      "properties": {
+        "type": {
+          "type": "string",
+          "enum": ["str", "int", "float", "bool", "bytes"]
```
RobbeSneyders, line 43: Currently used the most relevant Python types here. Do we need additional types such as `vector`?  Any existing type systems we can leverage?

express/manifest.py
```diff
@@ -1,112 +1,113 @@
...
+        """Dump the manifest to the file specified by the provided path"""
+        with open(path, "w", encoding="utf-8") as file_:
+            json.dump(self._specification, file_)
+
+    @property
+    def metadata(self) -> dict:
+        return self._specification.get("metadata")
+
+    @property
+    def base_path(self) -> str:
```
RobbeSneyders, line 101: Rewrote this from scratch and only then saw that we were leveraging the bucket before instead of a general base path. If we use the bucket, we probably don't need to resolve the locations of the subsets.

@PhilippeMoussalli I think you have the best view on what approach would be most useful here.

express/schemas/manifest.json
NielsRogge: Vectors are typically stored as `float`, so I guess we can use that one

express/manifest.py
```diff
@@ -1,112 +1,113 @@
...
-components of the pipeline
-"""
+"""This module defines classes to represent an Express manifest."""
+
 import json
-from dataclasses import dataclass, field
-from enum import Enum
-from typing import List, Dict
+import pkgutil
+import typing as t
```
NielsRogge, line 5: Can we avoid one-letter imports in favor of `from typing import Dict, Union` etc

express/manifest.py
```diff
@@ -1,112 +1,113 @@
...
+        base_path: The base path which the subset location is defined relative to
     """
 
-    location: str
-    type: DataType
-    extensions: List[str]
-    n_files: int = field(default_factory=int)
-    n_items: int = field(default_factory=int)
+    def __init__(self, specification: dict, *, base_path: str) -> None:
+        self._specification = specification
```
NielsRogge, line 32:
```suggestion
        self.specification = specification
```
(just a question) can we just use regular variable names without _? I know they mean private but they hurt readability imo, just easier to have regular variable names. Also done like that in HF libraries

express/manifest.py
```diff
@@ -1,112 +1,113 @@
...
+    """Special case of a subset for the index, which has fixed fields"""
 
-@dataclass_json
-@dataclass
-class DataManifest:
+    @property
+    def fields(self) -> t.Dict[str, Field]:
+        return {
+            "id": Field(name="id", type="str"),
+            "source": Field(name="source", type="str"),
```
NielsRogge, line 55: If I understand correctly, "source" is the same thing as "namespace"?

express/schemas/manifest.json
RobbeSneyders: The vector would be an array of `float`s though, so the type of the column is not `float`.

NielsRogge (approved):
I like the overall design and use of `jsonschema`!

express/manifest.py
RobbeSneyders: Yes, both terms have been used interchangeably.

express/manifest.py
RobbeSneyders: Private variables and methods convey meaning on what should and should not be used by external classes. It also lets you limit the public interface of your class which you need to keep backward compatible.

I think the benefit of private variables is more than worth the underscore :slightly_smiling_face:

express/manifest.py
RobbeSneyders: I like `import typing as t` because:
- Importing every type directly clutters the namespace and obfuscate where an import is coming from (eg. `Iterable` can be imported from `typing` or from `collections`)
- Importing typing as a module and repeating `typing.` becomes long very quickly.

I'm not married to this approach though, and the important thing is probably to standardize on a way to do this across the repo.

My second choice would still be `import typing` instead of `from typing import ...`

express/schemas/manifest.json
NielsRogge: Ok yes `list` should be supported, as well as `tuple` and `dict`

express/manifest.py
PhilippeMoussalli: Before the path included both the protocol(e.g. gs://)+bucket+location (a.k.a blob). 

I would prefer to move away towards something where maybe we have a dataclass that separates all 3 of those? 

- Protocol: can be used to reconstruct the full path and identify the remote location of an object
- Bucket:  fuse takes a bucket and a mount path for mounting where the full mount path will be `mount_path= Path(mount_path/bucket)
- Location: used to track down the subset paths for reading/writing `Path(mount_path/location)` 

You could also just have the full path (e.g. gs://bucket/blob) and parse it to extract what you need from it. Not sure which approach is more suitable.

express/schemas/manifest.json
RobbeSneyders: Data in `parquet` is not stored as a `list`, `tuple`, or `dict`.

We can probably leverage the parquet types:
- https://parquet.apache.org/docs/file-format/types/
- https://github.com/apache/parquet-format/blob/master/LogicalTypes.md
- https://learn.microsoft.com/en-us/common-data-model/sdk/parquet-to-cdm-datatype

PhilippeMoussalli: Indeed the types are probably not one-to-one mapping to the python data types but have their own notations (probably byte array). 

https://arrow.apache.org/docs/python/generated/pyarrow.ListArray.html

PhilippeMoussalli (approved):
Thanks Robbe! LGTM. I'll use the same approach for the component spec

express/manifest.py
```diff
@@ -1,112 +1,113 @@
...
-        for name, dataset in self.data_sources.items():
-            if not DataType.is_valid(dataset.type):
-                raise TypeError(
-                    f"Data type '{dataset.type}' for data source '{name}' is not valid."
-                )
+    def __init__(self, specification: dict) -> None:
+        self._validate_spec(specification)
+        self._specification = specification
+
+    @classmethod
```
PhilippeMoussalli, line 71: I don't think this is a class method

express/schemas/manifest.json
GeorgesLorre: For component validating:

do we validate the parquet schema or load it into something python and validate?

express/manifest.py
PhilippeMoussalli: I tried out this feature but it crashed my OS when running in background mode ([see doc](https://docs.google.com/document/d/1Yn8ayLBovv2PHlLQhsxp81Wzp8I23pOjW4xlMkqwmjs/edit#)). Also tried running it in the foreground with a thread and ran into similar issues. It's documented as an experimental feature but feel free to have a try it out. If you can make it work it will make things easier for us

express/manifest.py
RobbeSneyders: I changed this indeed.

RobbeSneyders:
Added functionality to create a manifest using the class as well.

express/schemas/manifest.json
RobbeSneyders: What I currently had in my mind is:
- We validate the manifests and components against each others. The used type system doesn't really matter here as long as it's consistent across all manifests and components.
- We instruct the dataframe library to load the data from parquet. I think we can provide a schema if we want the library to validate the parquet schema, otherwise it will just infer it.

Not sure at which level we need to define the types. Referring to the [parquet docs](https://parquet.apache.org/docs/file-format/types/), I think we need to define it on the logical level, since that will be important for the components, not the underlying storage type.
- PyArrow specifies logical types ([docs](https://arrow.apache.org/docs/python/api/datatypes.html#api-types)).
- So does eg. Polars ([docs](https://pola-rs.github.io/polars/py-polars/html/reference/datatypes.html))

They mostly match and I based the current list on these.

Following a dataframe library would be easiest, but probably also limits us to that framework which we might not want to do on the long term. `PyArrow` is used by most of these frameworks, so that seems like a good basis for multi-library compatibility.

GeorgesLorre (approved):
Nice work

express/manifest.py
RobbeSneyders: Does it provide a lot of benefit to split the protocol and the bucket? This is quite easy to do, right? Just thinking what would be easiest for the users.

express/manifest.py
RobbeSneyders: As discussed offline, let's go with `import typing as t`

PR #29, opened 2023-04-12
Host documentation

NielsRogge:
This PR adds a draft for documentation.

It uses [MkDocs](https://github.com/mkdocs/mkdocs), and hosting can happen on readthedocs/Github pages. The reason I went for MkDocs is because it uses Markdown, which is the most handy format to write documentation in my opinion. Markdown is a lot easier to work with compared to restructed text (rst) if you ask me (HF also updated from rst to Markdown and hot damn that was a relief).

You can easily view the docs locally by doing (assuming you have done `pip install mkdocs`):
```
cd docs
mkdocs serve
```
This will locally serve the docs.

Good discussion on MkDocs vs Sphinx here: https://github.com/encode/httpx/discussions/1220


RobbeSneyders:
Thanks @NielsRogge!

The discussion you link mostly seems to favor Sphinx though, which can also work with markdown. So preference for Markdown might not have to mean preference for MkDocs?

What do we need to do to serve this with either ReadTheDocs or Github pages?

NielsRogge:
> The discussion you link mostly seems to favor Sphinx though, which can also work with markdown. So preference for Markdown might not have to mean preference for MkDocs?

Yes I saw that too. But another feature I like about MkDocs is the fact that you can easily view your documentation as you're updating it. The local server updates automatically when you change a markdown file. With Sphinx on the other hand, you need to run make html, wait for everything to be built, then manually click on the html page to display it in your browser. 

Lots of libraries like FastAPI and Starlette leverage mkdocs-material: https://github.com/squidfunk/mkdocs-material. We might want to look into that as well, it's built on top of Mkdocs

NielsRogge:
Also @RobbeSneyders fun fact, the discussion above ended up with Mkdocs-material being chosen :D

RobbeSneyders:
Ok, for me this is fine. Let's use mkdocs-material as well?

- Can we use the readme as the index?
- Can you add the mkdocs dependency to the `pyproject.toml` under a `docs` dependency group?

GeorgesLorre:
> Ok, for me this is fine. Let's use mkdocs-material as well?
> 
> * Can we use the readme as the index?
> * Can you add the mkdocs dependency to the `pyproject.toml` under a `docs` dependency group?

@NielsRogge can we fix this ? Then we can merge this

RobbeSneyders:
> Let's use mkdocs-material as well?

We're still using the ReadTheDocs theme, right? I like the material one better.

> * Can we use the readme as the index?

I meant using `README.md` directly, not duplicating the content. Is this possible?

Does the `mkdocs.yml` need to be at the root level? For me it's fine to keep it in the `docs` folder, that way we don't clutter the root directory.

NielsRogge:
> I meant using README.md directly, not duplicating the content. Is this possible?

I'm afraid we need to duplicate it. All markdown files need to be in the docs folder. The mkdocs.yml file needs to be at the root level of the project.

RobbeSneyders:
Seems like it is possible to reference markdown files outside of the docs folder using the snippets plugin: https://github.com/mkdocs/mkdocs/issues/2831. I would prefer this over copying the contents.

.github/workflows/pipeline.yaml
```diff
@@ -37,4 +37,7 @@ jobs:
           pylint fondant --exit-zero
       - name: Test with pytest
         run: |
-          poetry run pytest tests
\ No newline at end of file
+          poetry run pytest tests
+      - name: Deploy documentation
```
RobbeSneyders, line 41: This will now run on every PR and overwrite the same version since there is only one.

I would propose to use ReadTheDocs instead, since it can build PRs and host multiple versions at the time.

RobbeSneyders:
Thanks for the changes!

I think we need to have a file named `index.md` in the `docs` folder though.

I tried to configure ReadTheDocs, but [the build is failing](https://readthedocs.org/projects/fondant/builds/20317978/). I think we need to add a `readthedocs.yaml` to configure the installation. See [Connexion example](https://github.com/spec-first/connexion/blob/main/.readthedocs.yaml).

RobbeSneyders (approved):
Docs for this PR are online here: https://fondant--29.org.readthedocs.build/en/29/ and you can also see the build status in the Github checks now.

Can you quickly review my changes @NielsRogge?

NielsRogge (commented):
Thanks for fixing!

PR #30, opened 2023-04-12
Create ComponentSpec class to match discussed specification

PhilippeMoussalli:
This PR introduces a component spec class to match the discussion on the dataset doc. 

- Add a component spec with specific schema that defines input and output subsets and their fields.
- Follows the same approach for defining the manifest schema as implemented in [PR 28](https://github.com/ml6team/express/pull/28)

[Edit]
PR updated to integrate kubeflow component spec within Express component. The Express component spec consists of kubeflow spec + input/output subset definition. We now define a minimal yaml file used to create the components and that is used for creating the kfp yaml files required to build pipelines


RobbeSneyders (commented):
Thanks @PhilippeMoussalli!

Left some comments below.

There will be some conflicts with the manifest PR, so we might want to merge that one already and address the open comments in follow-up PRs.

express/component_spec.py
```diff
@@ -0,0 +1,116 @@
+"""This module defines classes to represent an Express component specifications."""
```
NielsRogge, line 1:
```suggestion
"""This module defines classes to represent an Express component specification."""
```

express/component_spec.py
```diff
@@ -0,0 +1,116 @@
...
+import json
+import pkgutil
+import yaml
+import typing as t
+from typing import Dict
```
NielsRogge, line 6: Choose one or the other please

express/component_spec.py
```diff
@@ -0,0 +1,116 @@
...
+            name: ComponentField(name=name, type=field["type"])
+            for name, field in self._specification["fields"].items()
+        }
+
+
+class ExpressComponent:
+    """
+    Class representing an Express component
+    Args:
+        specification: The component specification as a Python dict
```
NielsRogge, line 61: The yaml_spec_path can be added to the docstring

express/component_spec.py
```diff
@@ -0,0 +1,116 @@
...
+    """
+
+    def __init__(self, specification: dict, yaml_spec_path: str) -> None:
+        self._validate_spec(specification)
+        self._specification = specification
+        self._metadata = load_yaml(yaml_spec_path)
+
+    @staticmethod
+    def _validate_spec(spec: dict) -> None:
```
NielsRogge, line 70: If this functionality is shared with @RobbeSneyders's manifest spec, does it make sense to have a superclass that defines this functionality?

NielsRogge, line 70: Can we have this in a superclass if this is also used for the manifest spec?

cc @RobbeSneyders

express/component_spec.py
```diff
@@ -0,0 +1,116 @@
...
+    """Class representing a single field or column in an Express component subset."""
+
+    name: str
+    type: str
+
+
+@dataclass
+class ComponentSubsets:
+    """
+    Class representing an Express Component subset.
```
NielsRogge, line 35:
```suggestion
    Class representing an Express Component's subsets.
```

tests/test_components_specs.py
```diff
@@ -0,0 +1,93 @@
...
+        "subset2": {
+            "fields": {
+                "score": {"type": "int"},
+                "grade": {"type": "str"}
+            }
+        }
+    }
+}
+
+print(ExpressComponent(VALID_COMPONENT_SPEC, yaml_path()))
```
NielsRogge, line 73: Can be deleted

express/component_spec.py
RobbeSneyders: It loads a different schema and throws a different error, so I don't think it's worth creating a base class. Class hierarchy has downsides as well, so if it doesn't create enough benefits, I would keep them separate.

express/component_spec.py
```diff
@@ -0,0 +1,116 @@
...
+    except Exception as e:
+        raise ValueError(f"Failed to load YAML file: {e}")
+
+
+@dataclass
+class ComponentField:
+    """Class representing a single field or column in an Express component subset."""
+
+    name: str
+    type: str
```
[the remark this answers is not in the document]
NielsRogge: Agree, we can just have a `Field` class that is shared between the Manifest and Component specs

express/component_spec.py
NielsRogge: Yes but you could have the schema as an additional argument and use it when calling.

But fine for me if you want to keep them separate.

express/component_spec.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+@dataclass
+class ComponentSubsets:
```
NielsRogge, line 33:
```suggestion
class ComponentSubset:
```
Looks like this class represents a single subset

express/component_spec.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+class ExpressComponent:
+    """
+    Class representing an Express component
+    Args:
+        specification: The component specification as a Python dict
+    """
+
+    def __init__(self, specification: dict, yaml_spec_path: str) -> None:
```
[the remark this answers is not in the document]
NielsRogge: Agree, if it's not possible including the Kubeflow spec in the Fondant spec seems like an elegant solution.

express/component_spec.py
GeorgesLorre: I'm also a fan of somehow merging the 2 specs. I would say using the Fondant spec to generate the kubeflow spec (while compiling the pipeline). The current kubeflow spec seems very minimal.

express/component_spec.py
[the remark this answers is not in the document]
PhilippeMoussalli: There is not much information be added from the kubeflow **component** spec (`component.yaml`) besides the name and description. However, when compiling the pipeline can get a kubeflow **pipeline** spec which you can use to get information about your DAG (e.g. how components are sequenced).

Eventually, we can create an `ExpressPipeline`  class that takes into account both the kubeflow pipeline spec and the ExpressComponent class to check if all the subsets and columns that are needed are created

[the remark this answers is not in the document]
RobbeSneyders: The idea would be to have a single specification like this:
```yaml
name: Example component
description: This is an example component
image: example_component:latest

accepts:
  images:
    data: binary
 
produces:
  captions:
    data: utf8

args:
  # Additional arguments accepted by the component
```
Which we then translate internally to the KfP component spec:
```yaml
name: Example component
description: This is an example component

inputs:
  - name: input_manifest
    description: Path to the input manifest
    type: String

  # args from above

outputs:
  - name: output_manifest
    description: Path to the output manifest

implementation:
  container:
    image: example_component:latest
    command: [
      python3, main.py,
      --input-manifest,       { inputPath: input_manifest },
      --output-manifest,      { outputPath: output_manifest },
      # args from above
    ]
```

Something like this. I just quickly typed this out to sketch the idea.

express/component_spec.py
PhilippeMoussalli: I think this would be a good idea eventually to have a single reference point for defining the components (right now you have to define both a component spec class + component yaml file). We can expand the ComponentSpec class to generate the yaml file but best to tackle this in a separate PR. I'll create a ticket for it.

tests/test_components_specs.py
```diff
@@ -0,0 +1,40 @@
...
+
+valid_path = os.path.join("component_example", "valid_component")
+invalid_path = os.path.join("component_example", "invalid_component")
+
+
+def valid_express_schema() -> str:
+    return os.path.join(valid_path, "express_component.yaml")
+
+
+def valid_kubeflow_schema() -> str:
```
GeorgesLorre, line 15: this is not used ? I guess you want to add a test where you compare the generated one with this one ?

express/component_spec.py
```diff
@@ -0,0 +1,335 @@
...
+        self._set_component_run_cmd()
+
+    @staticmethod
+    def _get_default_inputs() -> t.List[Input]:
+        """
+        Set default inputs to components
+        """
+        inputs = [
+            Input(
+                name="input_manifest",
```
NielsRogge, line 91:
```suggestion
                name="input_manifest_path",
```
Maybe it'd be clearer to define that we're expecting a path

express/component_spec.py
```diff
@@ -0,0 +1,335 @@
...
+                description="Path to the the input manifest",
+                type="String",
+            ),
+            Input(
+                name="extra_args",
```
NielsRogge, line 96:
```suggestion
                name="args",
```
Let's just call it args. Does this include the metadata like run_id and component_id?

express/common.py
```diff
@@ -0,0 +1,56 @@
+"""This module defines common schemas and datatypes used to define Express manifests, components
+and pipelines."""
+
+import enum
+import typing as t
+
+
+class Type(enum.Enum):
```
NielsRogge, line 8: This is already defined in express/manifest.py so we can reuse it

express/component_spec.py
```diff
@@ -0,0 +1,335 @@
...
+        return inputs
+
+    @staticmethod
+    def _get_default_outputs() -> t.List[Output]:
+        """
+        Set default output to components
+        """
+        outputs = [
+            Output(
+                name="output_manifest", description="The path to the output manifest"
```
NielsRogge, line 110:
```suggestion
                name="output_manifest_path", description="The path to the output manifest"
```
Same here

express/component_spec.py
```diff
@@ -0,0 +1,335 @@
...
+
+
+@dataclass
+class KFPComponent:
+    """
+    A class representing a Kubeflow Pipelines component.
+    Attributes:
+        name: The name of the component.
+        description: The description of the component.
+        image: The Docker image for the component.
```
NielsRogge, line 57: The name of the Docker image you mean? Or the URL?

tests/test_components_specs.py
PhilippeMoussalli: indeed, was missing that one, just implemented one and will update

express/component_spec.py
```diff
@@ -0,0 +1,328 @@
...
+import jsonschema.exceptions
+from jsonschema import Draft4Validator
+
+from express.exceptions import InvalidComponentSpec
+from express.io_utils import load_yaml
+from express.common import Field
+from express.pipeline_utils import get_kubeflow_type
+
+
+@dataclass
```
NielsRogge, line 19: Do we benefit from this being a dataclass?

express/component_spec.py
```diff
@@ -0,0 +1,328 @@
...
+class Input:
```
NielsRogge, line 20:
```suggestion
class InputArgument:
```
Maybe let's call it like that to not confuse it with input subsets

express/component_spec.py
```diff
@@ -0,0 +1,328 @@
...
+    """
+    Component input arguments
```
NielsRogge, line 22:
```suggestion
    Component input argument
```

express/component_spec.py
```diff
@@ -0,0 +1,328 @@
...
+        type: the python argument type (str, int, ...)
+    """
+
+    name: str
+    description: str
+    type: str
+
+
+@dataclass
+class Output:
```
NielsRogge, line 35:
```suggestion
class OutputArgument:
```

express/component_spec.py
```diff
@@ -0,0 +1,328 @@
...
+    Attributes:
+        name: name of the argument
+        description: argument description
+    """
+
+    name: str
+    description: str
+
+
+T = t.TypeVar("T", Input, Output)
```
NielsRogge, line 47: this looks very cryptic to me, makes the code pretty hard to read

I'd just use the longer names to make it clearer

express/component_spec.py
```diff
@@ -0,0 +1,328 @@
...
+    name: str
+    description: str
+    image: str
+    args: t.Dict[str, t.Dict[str, str]]
+    inputs: t.Union[Input, t.List[Input]] = None
+    outputs: t.Union[Output, t.List[Output]] = None
+    command: t.List[t.Union[str, t.Dict[str, str]]] = None
+
+    def __post_init__(self):
+        self.inputs = self._add_defaults_values(
```
NielsRogge, line 73:
```suggestion
        self.input_arguments = self._add_defaults_values(
```

express/component_spec.py
```diff
@@ -0,0 +1,328 @@
...
+
+        self.outputs = self._add_defaults_values(
+            self.outputs, self._get_default_outputs(), Output
+        )
+
+        self._add_input_args()
+        self._set_component_run_cmd()
+
+    @staticmethod
+    def _get_default_inputs() -> t.List[Input]:
```
NielsRogge, line 85:
```suggestion
    def _get_default_input_arguments() -> t.List[Input]:
```

express/component_spec.py
```diff
@@ -0,0 +1,328 @@
...
+        for arg_name, arg_info in self.args.items():
+            self.inputs.append(
+                Input(
+                    name=arg_name.strip(),
+                    description=arg_info["description"].strip(),
+                    type=get_kubeflow_type(arg_info["type"].strip()),
+                )
+            )
+
+    def _set_component_run_cmd(self):
```
NielsRogge, line 144:
```suggestion
    def _set_component_run_command(self):
```
Let's use clear names everywhere

express/component_spec.py
```diff
@@ -0,0 +1,328 @@
...
+        """
+        Function that return the run command of the kubeflow component
```
NielsRogge, line 146:
```suggestion
        Function that returns the run command of the kubeflow component
```

express/component_spec.py
```diff
@@ -0,0 +1,328 @@
...
+                self.command.append(arg_name_cmd)
+                self.command.append({arg_kfp_value_type: arg_name})
+
+        self.command = ["python3", "main.py"]
+        _add_run_arguments(self.inputs)
+        _add_run_arguments(self.outputs)
+
+    def get_specification(self) -> dict:
+        """
+        Function that returns the specifications of the kubeflow component
```
NielsRogge, line 172:
```suggestion
        Function that returns the specification of the KubeFlow component
```

express/component_spec.py
NielsRogge, line 20:
```suggestion
class KubeflowInput:
```
Maybe let's call it KubeflowInput to not confuse it with the input subsets?

express/component_spec.py
NielsRogge, line 35:
```suggestion
class KubeflowOutput:
```

express/component_spec.py
```diff
@@ -0,0 +1,328 @@
...
+            "outputs": [asdict(output_obj) for output_obj in self.outputs],
+            "implementation": {
+                "container": {"image": self.image, "command": self.command}
+            },
+        }
+
+        return specification
+
+
+class ComponentSubset:
```
NielsRogge, line 201: Is there a need for a different class given that we already have a `Subset` class?

PhilippeMoussalli: the other class has a `location` field which is not needed when you define the component

express/component_spec.py
PhilippeMoussalli: fine by me

express/component_spec.py
```diff
@@ -0,0 +1,328 @@
...
+        )
+        validator = Draft4Validator(spec_schema)
+        try:
+            validator.validate(self.yaml_spec)
+        except jsonschema.exceptions.ValidationError as e:
+            raise InvalidComponentSpec.create_from(e)
+
+    def set_kubeflow_comp_specification(self) -> t.Dict[str, any]:
+        """
+        Function that returns the kubeflow specifications as a dictionary
```
NielsRogge, line 255:
```suggestion
        Function that returns the kubeflow specification as a dictionary
```

express/component_spec.py
```diff
@@ -0,0 +1,328 @@
...
+        """
+        express_component_spec = copy.deepcopy(self._kubeflow_comp_specs)
+        express_component_spec["input_subsets"] = self.yaml_spec["input_subsets"]
+        express_component_spec["output_subsets"] = self.yaml_spec["output_subsets"]
+
+        return express_component_spec
+
+    def write_component(self, path: str):
+        """
+        Function that write the component yaml file required to compile a kubeflow pipeline
```
NielsRogge, line 280:
```suggestion
        Function that writes the component yaml file required to compile a kubeflow pipeline
```

express/component_spec.py
NielsRogge: Ok but people will see a T and will wonder what it means

express/component_spec.py
NielsRogge: Ok, but the location is just defined as a `property`, right?

PhilippeMoussalli: also takes a `base_path` as a input for initialization so I think they're better separated

express/component_spec.py
PhilippeMoussalli: T" is a common convention used in Python (and other programming languages) to represent a generic type parameter. 
https://peps.python.org/pep-0585/

express/common.py
RobbeSneyders, line 1: I don't like the name `common.py` here, it doesn't really say anything about what is in the file. I would propose `schema.py`, but open to other suggestions.

express/component_spec.py
```diff
@@ -0,0 +1,346 @@
...
+from express.io_utils import load_yaml
+from express.common import Field
+from express.pipeline_utils import get_kubeflow_type
+
+
+@dataclass
+class KubeflowInput:
+    """
+    Kubeflow component input argument
+    Attributes:
```
RobbeSneyders, line 25: This should be `Args` instead of `Attributes`. We follow the google docstring format.

RobbeSneyders: Same for a lot of docstrings in this PR.

express/component_spec.py
```diff
@@ -0,0 +1,346 @@
...
+        base_uri = (Path(__file__).parent / "schemas").as_uri()
+        resolver = RefResolver(base_uri=f"{base_uri}/", referrer=spec_schema)
+        validator = Draft4Validator(spec_schema, resolver=resolver)
+
+        try:
+            validator.validate(self.yaml_spec)
+        except jsonschema.exceptions.ValidationError as e:
+            raise InvalidComponentSpec.create_from(e)
+
+    def set_kubeflow_component_specification(self) -> t.Dict[str, any]:
```
RobbeSneyders, line 263: Don't like the `set_` here, as the method is not setting anything. Same for the method below.

express/component_spec.py
```diff
@@ -0,0 +1,346 @@
...
+        Function that return the express component specification which contains both the Kubeflow
+        component specifications in addition to the subsets defining the input and output datasets
+        """
+        express_component_spec = copy.deepcopy(self._kubeflow_component_specification)
+        express_component_spec["input_subsets"] = self.yaml_spec["input_subsets"]
+        express_component_spec["output_subsets"] = self.yaml_spec["output_subsets"]
+
+        return express_component_spec
+
+    def write_component(self, path: str):
```
RobbeSneyders, line 288: Would clarify that this is the kubeflow `component.yaml` in the method name.

express/io_utils.py
```diff
@@ -0,0 +1,16 @@
...
+    """
+    Loads a YAML file and returns a dictionary.
+    Args:
+        yaml_path (str): the path to the yaml file
+    """
+    try:
+        with open(yaml_path) as f:
+            return yaml.safe_load(f)
+    except Exception as e:
+        raise ValueError(f"Failed to load YAML file: {e}")
```
RobbeSneyders, line 16: Why do we throw a `ValueError` here instead of just raising the original yaml error??

express/pipeline_utils.py
```diff
@@ -12,6 +12,31 @@
 logger = logging.getLogger(__name__)
 
 
+def get_kubeflow_type(python_type: str):
```
RobbeSneyders, line 15: This should probably not be in `pipeline_utils.py` if we're using it in the component.

tests/component_example/valid_component/express_component.yaml
```diff
@@ -0,0 +1,18 @@
+name: Example component
+description: This is an example component
+image: example_component:latest
+
+input_subsets:
+  images:
+    fields:
+      data: binary
```
RobbeSneyders, line 8: I wrapped this in an object in the manifest, so we can extend it in the future if needed. We might want to do the same for the component spec:

```yaml
input_subsets:
  images:
    fields:
      data:
        type: binary
```

tests/test_components_specs.py
```diff
@@ -0,0 +1,49 @@
...
+import pytest
+from express.exceptions import InvalidComponentSpec
+from express.component_spec import ExpressComponent
+from express.io_utils import load_yaml
+
+valid_path = os.path.join("component_example", "valid_component")
+invalid_path = os.path.join("component_example", "invalid_component")
+
+
+def valid_express_schema() -> str:
```
RobbeSneyders, line 12: You can turn these into fixtures.

express/component_spec.py
```diff
@@ -0,0 +1,346 @@
...
+        """
+        Set default inputs to components
+        """
+        inputs = [
+            KubeflowInput(
+                name="input_manifest_path",
+                description="Path to the the input manifest",
+                type="String",
+            ),
+            KubeflowInput(
```
RobbeSneyders, line 97: Why are these always added? I would expect all of them to be defined in the Fondant component spec.

RobbeSneyders (commented):
Thanks @PhilippeMoussalli, looks really good overall.

Left some more comments, but they are pretty small.

express/component_spec.py
PhilippeMoussalli: Do you mean in the yaml Component that the user has to define or directly set in the `FondantComponent` class. I have a preference for the latter since the user then does not have to define the manifest inputs and outputs (or need to know about those concepts)

tests/component_example/valid_component/express_component.yaml
PhilippeMoussalli: good point, i'v added it

express/io_utils.py
PhilippeMoussalli: My bad, should be a general YAML error indeed

RobbeSneyders: Just re-raising the error doesn't provide a lot of value, right? Then we can just remove the `try ... except`, in which case the whole function doesn't really provide a lot of value.

express/component_spec.py
PhilippeMoussalli: yes @NielsRogge pointed this out as well, it is now removed from the default inputs. 

This means that we're now parsing each argument individually as opposed to passing one json string with all the arguments. Kubeflow has a weird way of dealing with bools and iterables so we have to make sure we have a proper way of dealing with this in the `Datasets` class

RobbeSneyders (approved):
Thanks @PhilippeMoussalli, great work. Let's merge!

PR #32, opened 2023-04-19
Add input and output arguments as ComponentSpec properties 

PhilippeMoussalli:
PR that adds the input and output arguments of the component as a property to the `ExpressComponent` class for ease of retrieval


NielsRogge (approved):
Thanks for adding! Will merge as I'll use it

PR #33, opened 2023-04-20
Add Dataset and Component Class

NielsRogge:
This PR adds the `FondantDataset` wrapper class around the `Manifest`, which loads data as Dask dataframes and allows to upload Dask dataframes back to the cloud.

To test everything, the PR also includes a pipeline called "simple pipeline" that includes 3 components: loading from hub, image filtering and embedding. Each component needs to overwrite the `FondantComponent` class, which uses the `FondantDataset` class behind the scenes.

To be discussed:

- [ ] I've added `project_name` to the metadata of the manifest, in order to know the name of the cloud project. This is needed to load/upload data using `fsspec`.

To do:

- [ ] for now I'm manually adding the `"gcs://"` prefix and `".parquet"` suffix when loading data to and from the cloud, this needs to be addressed (we need a cleaner way that is not hardcoded)
- [ ] only first 2 components are implemented, embedding component is to do
- [ ] for the moment I'm still manually creating the KubeFlow component yaml file for each component. This should be updated to automatically create it based on the Fondant spec using the [write_kubeflow_specification](https://github.com/ml6team/express/blob/db5807ae868fe36091d8d7f0061450312ab7477b/express/component_spec.py#L207) method
- [ ] nicer way of creating and passing metadata. The only metadata that is different per component is the `component_id`. Ideally getting rid of `args.metadata`
- [ ] enforce usage of data types defined in output subsets when creating the dataset (currently only the column names are checked)


examples/pipelines/simple_pipeline/components/embedding/embedding.yaml
```diff
@@ -0,0 +1,23 @@
+name: Embedding
+description: Component that embeds images using CLIP
+image: embedding:latest
+
+input_subsets:
```
PhilippeMoussalli, line 5: why is the input subsets empty here?

examples/pipelines/simple_pipeline/components/embedding/embedding.yaml
```diff
@@ -0,0 +1,23 @@
+name: Embedding
+description: Component that embeds images using CLIP
+image: embedding:latest
```
PhilippeMoussalli, line 3: The image needs to be updated

examples/pipelines/simple_pipeline/components/image_filtering/requirements.txt
```diff
@@ -0,0 +1 @@
+git+https://github.com/ml6team/express.git@b1855308ca9251da5ddd8e6b88c34bc1c082a71b#egg=express[datasets]
```
PhilippeMoussalli, line 1: I guess this will change to express[Polars] at some point. Are we still leveraging HF datasets?

examples/pipelines/simple_pipeline/components/image_filtering/src/fondant_component.yaml
```diff
@@ -0,0 +1,27 @@
+name: Image filtering
```
PhilippeMoussalli, line 1: I think the image width and height can be int16 (max value of 32767). Saves us some memory for the parquet file (not relevant now but might be more relevant with large scale datasets)

examples/pipelines/simple_pipeline/components/embedding/embedding.yaml
```diff
@@ -0,0 +1,23 @@
...
+    fields:
+      data:
+        type: binary
+  captions:
+    fields:
+      data:
+        type: utf8
+
+args:
+  model_id:
```
PhilippeMoussalli, line 18: Why is the model id used here?

examples/pipelines/simple_pipeline/components/image_filtering/src/main.py
```diff
@@ -0,0 +1,39 @@
...
+class ImageFilterComponent(FondantComponent):
+    """
+    Component that filters images based on height and width.
+    """
+
+    @classmethod
+    def process(cls, dataset: Dataset, args: Dict) -> Dataset:
+        """
+        Args:
+            dataset
```
PhilippeMoussalli, line 25: Dataset description missing

examples/pipelines/simple_pipeline/components/image_filtering/src/main.py
```diff
@@ -0,0 +1,39 @@
...
+logger = logging.getLogger(__name__)
+
+
+class ImageFilterComponent(FondantComponent):
+    """
+    Component that filters images based on height and width.
+    """
+
+    @classmethod
+    def process(cls, dataset: Dataset, args: Dict) -> Dataset:
```
PhilippeMoussalli, line 22: Can we swtich over to using the t.Dict everywhere to be consistent. Also, if you're the `Dict` from the `typing` module, it's better to define the datatypes inside of it

examples/pipelines/simple_pipeline/components/image_filtering/Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM --platform=linux/amd64 python:3.8-slim
```
PhilippeMoussalli, line 1: Do we need still need to define the `platform` flag here if you're passing it in the `build_image.sh` script at build time

examples/pipelines/simple_pipeline/components/load_from_hub/src/main.py
```diff
@@ -0,0 +1,74 @@
...
+        index_list = [idx for idx in range(len(dask_df))]
+        source_list = ["hub" for _ in range(len(dask_df))]
+
+        dask_df["id"] = da.array(index_list)
+        dask_df["source"] = da.array(source_list)
+
+        # 3) Rename columns
+        dask_df = dask_df.rename(columns={"image": "images_data", "text": "captions_data"})  
+        
+        # 4) Add width and height columns
```
PhilippeMoussalli, line 66: Seems a bit inefficient to do two full passes over the same column, maybe you can try something like this: 

```
def get_dims(row):
    # perform some computation on the row
    return width, height

# apply the function using map and assign the results to two new columns
df = df['image_data'].map(get_dims, ...)
df = df.assign(image_width=df[0], image_height=df[1])
```
Not tested but something like this should do it

examples/pipelines/simple_pipeline/components/load_from_hub/src/main.py
```diff
@@ -0,0 +1,74 @@
...
+        """
+        Args:
+            args: additional arguments passed to the component
+        
+        Returns:
+            Dataset: HF dataset
+        """
+        # 1) Load data, read as Dask dataframe
+        logger.info("Loading dataset from the hub...")
+        dataset = load_dataset(args.dataset_name, split="train")
```
PhilippeMoussalli, line 51: hmm Not really a big fan or this approach (loading, writing and reading again with another framework) but I guess it's fine for now. Maybe create a ticket for it?

examples/pipelines/simple_pipeline/components/load_from_hub/src/main.py
```diff
@@ -0,0 +1,74 @@
...
+        dataset.to_parquet("data.parquet")
+        dask_df = dd.read_parquet("data.parquet")
+
+        # 2) Add index to the dataframe
+        logger.info("Creating index...")
+        index_list = [idx for idx in range(len(dask_df))]
```
PhilippeMoussalli, line 57: I think those are created in memory as a list and then passed to the "unmaterialized" dask dataframe, ideally they would be created on the fly similar to the map function. Similar comment to the `source_list`

examples/pipelines/simple_pipeline/components/load_from_hub/src/main.py
```diff
@@ -0,0 +1,74 @@
...
+from PIL import Image
+
+from express.dataset import FondantComponent
+from express.logger import configure_logging
+
+configure_logging()
+logger = logging.getLogger(__name__)
+
+
+def extract_width(image_bytes):
```
PhilippeMoussalli, line 22: Prefer docstrings than comments

examples/pipelines/simple_pipeline/components/load_from_hub/src/main.py
```diff
@@ -0,0 +1,74 @@
...
+
+        dask_df["id"] = da.array(index_list)
+        dask_df["source"] = da.array(source_list)
+
+        # 3) Rename columns
+        dask_df = dask_df.rename(columns={"image": "images_data", "text": "captions_data"})  
+        
+        # 4) Add width and height columns
+        dask_df['images_width'] = dask_df['images_data'].map(extract_width, meta=pd.Series([], dtype=np.float64))
+        dask_df['images_height'] = dask_df['images_data'].map(extract_height, meta=pd.Series([], dtype=np.float64))
```
PhilippeMoussalli, line 68: the `np.float64` does not really match the type specified in the subsets. We need a way to check for that or propagate them through, they seem a bit redundant for now

examples/pipelines/simple_pipeline/components/load_from_hub/src/main.py
NielsRogge: What's the alternative? Starting from a dataset in a bucket?

express/component_spec.py
```diff
@@ -16,29 +16,23 @@
...
+    "int": "Integer",
+    "float": "Float",
+    "bool": "Boolean",
+    "dict": "Map",
+    "list": "List",
+    "tuple": "List",
+    "set": "Set",
+}
+
+kubeflow2python_type = {
```
PhilippeMoussalli, line 30: Some datatypes are missing, Can't we just invert the one from above?

express/component_spec.py
```diff
@@ -329,10 +323,9 @@ def output_arguments(self):
         """The output arguments of the component as an immutable mapping"""
         return types.MappingProxyType(
             {
-                info["name"]: KubeflowInput(
+                info["name"]: KubeflowOutput(
```
PhilippeMoussalli, line 326: my bad

express/dataset.py
```diff
@@ -0,0 +1,212 @@
+from abc import abstractmethod
```
PhilippeMoussalli, line 1: Missing Docstring at the beginning of the module

pyproject.toml
```diff
@@ -42,15 +42,13 @@ classifiers = [
 python = "^3.8"
 dataclasses-json = "^0.5.7"
 jsonschema = "^4.17.3"
+dask = "^2023.4.0"
```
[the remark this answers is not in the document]
NielsRogge: For now, Dask is a hard requirement as it's the only framework we support

express/dataset.py
```diff
@@ -0,0 +1,212 @@
...
+        location = "gcs://" + subset.location + ".parquet"
+        fields = list(subset.fields.keys())
+
+        df = dd.read_parquet(
+            location, columns=fields, storage_options={"project": self.project_name}
+        )
+
+        return df
+
+    def load_data(self, component_spec: ExpressComponent) -> dd.DataFrame:
```
PhilippeMoussalli, line 39: Missing docstring

express/dataset.py
```diff
@@ -0,0 +1,212 @@
...
+    """Wrapper around the manifest to download and upload data into a specific framework.
+
+    Uses Dask Dataframes for the moment.
+    """
+
+    def __init__(self, manifest):
+        self.manifest = manifest
+
+    @property
+    def project_name(self):
```
PhilippeMoussalli, line 22: not all clouds have the concept of `project_name`. What's the reason for adding this to the manifest?

examples/pipelines/simple_pipeline/components/load_from_hub/src/main.py
NielsRogge: Dask has some weird functionality, like why do I need to pass this `meta` thing with an empty Series?

express/dataset.py
NielsRogge: fsspec requires the full path, like "gcs://soy-audio-379412_datasets/pokemon.parquet". This includes the project name, bucket name and blob path.

express/dataset.py
```diff
@@ -0,0 +1,212 @@
...
+    def _load_subset(self, name: str) -> dd.DataFrame:
+        # get subset from the manifest
+        subset = self.manifest.get_subset(name)
+        # get its location and fields
+        # TODO remove gcp prefix
+        location = "gcs://" + subset.location + ".parquet"
+        fields = list(subset.fields.keys())
+
+        df = dd.read_parquet(
+            location, columns=fields, storage_options={"project": self.project_name}
```
PhilippeMoussalli, line 34: oh I see you're using it here, I would just define custom `storage_options` that the user has to pass. They might differ from one provider to the other + I'm not sure if the it's required to pass the `project` if you have the right credentials

express/dataset.py
```diff
@@ -0,0 +1,212 @@
...
+        remote_path = "gcs://" + self.manifest.index.location + ".parquet"
+        # upload to the cloud
+        dd.to_parquet(
+            df,
+            remote_path,
+            storage_options={"project": self.gcs_project},
+            overwrite=True,
+        )
+
+    def add_index(self, output_dataset):
```
PhilippeMoussalli, line 63: docstrings and typehints

tests/test_manifest.py
```diff
@@ -94,7 +94,7 @@ def test_manifest_creation():
     component_id = "component_id"
 
     manifest = Manifest.create(base_path=base_path, run_id=run_id, component_id=component_id)
-    manifest.add_subset("images", [("width", Type.int32), ("height", Type.int32)])
+    manifest.add_subset("images", [("width", "int32"), ("height", "int32")])
```
PhilippeMoussalli, line 97: what' the reason for changing this? Shouldn't it be a datatype

express/manifest.py
```diff
@@ -163,16 +171,25 @@ def subsets(self) -> t.Mapping[str, Subset]:
...
+
     def add_subset(self, name: str, fields: t.List[t.Tuple[str, Type]]) -> None:
         if name in self._specification["subsets"]:
             raise ValueError("A subset with name {name} already exists")
 
         self._specification["subsets"][name] = {
-            "location": f"/{name}/{self.run_id}/{self.component_id}",
-            "fields": {name: {"type": type_.value} for name, type_ in fields},
+            "location": f"/custom_artifact/{self.run_id}/{self.component_id}/{name}",
+            "fields": {name: {"type": type_} for name, type_ in fields},
```
PhilippeMoussalli, line 186: why did we remove the `values` from fields

examples/pipelines/simple_pipeline/components/image_filtering/requirements.txt
PhilippeMoussalli: my bad I confused them 🙃 I meant Dask

examples/pipelines/simple_pipeline/components/image_filtering/src/fondant_component.yaml
NielsRogge: Thanks, have added a todo for that, currently only the appropriate column names are checked.

examples/pipelines/simple_pipeline/components/load_from_hub/src/main.py
PhilippeMoussalli: that or writing/git clone directly to the bucket and then reading the data from there. That way we avoid reading the data in memory. Although I'm not sure how to achieve this. Can be tackled later

examples/pipelines/simple_pipeline/components/load_from_hub/src/main.py
PhilippeMoussalli: I noticed that as well. I read a bit on it and I think when you apply a function to a Dask DataFrame using the map() method, Dask needs to know the data type of the output to determine the optimal execution plan and to allocate the necessary resources for the computation.

Will have to look more into it

express/dataset.py
```diff
@@ -0,0 +1,209 @@
...
+    @property
+    def project_name(self):
+        return self.manifest.project_name
+
+    def _load_subset(self, name: str) -> dd.DataFrame:
+        # get subset from the manifest
+        subset = self.manifest.get_subset(name)
+        # get its location and fields
+        # TODO remove gcp prefix
+        location = "gcs://" + subset.location + ".parquet"
```
PhilippeMoussalli, line 35: I don't think the location is always a parquet file. Sometimes you need to refer to a parquet folder that contains all the different paritions

express/dataset.py
NielsRogge: Where would you define these?

PhilippeMoussalli: I think we can have the user define them once somewhere in a central config and then pass them in internally in the `FondantPipeline` class. Similar to the `metadata args` which might have to passed just as an input

express/dataset.py
```diff
@@ -0,0 +1,209 @@
...
+        fields = list(subset.fields.keys())
+
+        df = dd.read_parquet(
+            location, columns=fields, storage_options={"project": self.project_name}
```
RobbeSneyders, line 40: The columns to read here should be based on the `ComponentSpec` instead of the `Manifest`, right?

express/manifest.py
```diff
@@ -163,16 +171,25 @@ def subsets(self) -> t.Mapping[str, Subset]:
             }
         )
 
+    def get_subset(self, name: str) -> Subset:
```
RobbeSneyders, line 174: Do we need this? `manifest.subsets[name]` should already provide the same behavior. I don't think the subsets list will become that long that the dict comprehension overhead matters.

I designed the manifest class following:
    - Fixed properties should be accessible as an attribute
    - Dynamic properties should be accessible by lookup

express/dataset.py
```diff
@@ -0,0 +1,209 @@
...
+    type: str = "transform"
+
+    @classmethod
+    def _load_spec(cls) -> ExpressComponent:
+        # note: Fondant spec always needs to be called like this
+        # and placed in the src directory
+        spec_path = "fondant_component.yaml"
+        return ExpressComponent(spec_path)
+
+    @classmethod
```
RobbeSneyders, line 125: Is there a reason to have all `classmethod`s instead of initializing the class?

express/dataset.py
```diff
@@ -0,0 +1,209 @@
...
+        Returns:
+            Manifest: the output manifest
+        """
+        # step 1: load component spec
+        spec = cls._load_spec()
+        # step 2: add and parse arguments
+        args = cls._add_and_parse_args(spec)
+        # step 3: create Fondant dataset based on input manifest
+        metadata = json.loads(args.metadata)
+        if cls.type == "load":
```
RobbeSneyders, line 139: I would prefer to split this into two subclasses.

express/dataset.py
```diff
@@ -0,0 +1,209 @@
...
+        output_manifest = dataset.upload(save_path=args.output_manifest_path)
+
+        return output_manifest
+
+    @classmethod
+    def _add_and_parse_args(cls, spec):
+        """
+        Add and parse component arguments based on the component specification.
+        """
+        parser = argparse.ArgumentParser()
```
RobbeSneyders, line 176: Do we need to define these arguments since we're building on top of KfP 1.X? Shouldn't we build on top of 2.X?

express/dataset.py
```diff
@@ -0,0 +1,209 @@
...
+
+    def _upload_subset(self, name, fields, df) -> Subset:
+        # add subset to the manifest
+        fields = [(field.name, field.type) for field in fields.values()]
+        self.manifest.add_subset(name, fields=fields)
+        # upload to the cloud
+        # TODO remove prefix and suffix?
+        remote_path = "gcs://" + self.manifest.get_subset(name).location + ".parquet"
+        dd.to_parquet(
+            df,
```
RobbeSneyders, line 79: Shouldn't we remove the column name prefixes?

RobbeSneyders (commented):
Thanks @NielsRogge! I did a quick review, but haven't been able to go through it completely yet.

express/dataset.py
NielsRogge: Yes that's right, so what happens is we read the `input_subsets` from the `ComponentSpec`, which list which subsets to load. Next we get the corresponding subset from the manifest to get its location.

express/dataset.py
NielsRogge: Does KubeFlow 2.X not require arguments?

express/dataset.py
NielsRogge: You mean the subset name? That is already happening

express/dataset.py
NielsRogge: Yes the prefix and suffix additions are added as a todo at the top of this PR

express/manifest.py
NielsRogge: Ok but that just returns a dictionary right, not a `Subset` instance? The latter allows to easily get properties like `location` and `fields`. Also you would get a `KeyError` in case the subset is not present in the manifest.

express/dataset.py
RobbeSneyders: Yes, but I'm talking about which columns of the subset to load

express/dataset.py
RobbeSneyders: Can you point me to it?

express/manifest.py
NielsRogge: Ok I've removed the `get_subset` method in favor of this

express/dataset.py
NielsRogge: Ok yes that's right, have addressed that!

express/dataset.py
NielsRogge: Can probably be addressed in a follow-up PR, would focus on getting the pipeline to run on GCP first

examples/pipelines/simple_pipeline/components/load_from_hub/src/main.py
```diff
@@ -0,0 +1,77 @@
...
+        dataset.to_parquet("data.parquet")
+        dask_df = dd.read_parquet("data.parquet")
+
+        # 2) Add index to the dataframe
+        logger.info("Creating index...")
+        index_list = [idx for idx in range(len(dask_df))]
+        source_list = ["hub" for _ in range(len(dask_df))]
+
+        dask_df["id"] = da.array(index_list)
+        dask_df["source"] = da.array(source_list)
```
PhilippeMoussalli, line 59: we probably want the id to be Tuple of (id, source) to avoid any index clashing with other data sources. @RobbeSneyders what's your take on this?

PhilippeMoussalli, line 59: we probably want the id to be Tuple of (id, source) to avoid any index clashing with other data sources. @RobbeSneyders what's your take on this?

express/dataset.py
RobbeSneyders: Let's upgrade in a separate PR.

express/dataset.py
```diff
@@ -0,0 +1,209 @@
...
+        # add output args
+        for arg in spec.output_arguments.values():
+            parser.add_argument(
+                f"--{arg.name}",
+                required=True,
+                type=str,
+                help=arg.description,
+            )
+        # add metadata
+        parser.add_argument(
```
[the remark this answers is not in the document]
RobbeSneyders: Probably irrelevant if we upgrade to KfP 2.X.

express/dataset.py
```diff
@@ -0,0 +1,200 @@
...
+            # create HF dataset, based on component spec
+            input_dataset = dataset.load_data(self.spec)
+            # provide this dataset to the user
+            df = self.transform(
+                dataset=input_dataset,
+                args=args,
+            )
+
+        # step 4: create output manifest
+        output_manifest = dataset.upload(save_path=args.output_manifest_path)
```
RobbeSneyders, line 157: This method doesn't return anything.

But in any case, I think we should evolve the manifest from the dataset separately. We should be able to deduct the output manifest from just the input manifest and the component spec, which we can then use for validation of the output dataset. This functionality will also be useful for static analysis in the future.

We can rework this in a separate PR.

RobbeSneyders (commented):
Let's try to get this PR merged and do follow-up work in separate PRs.

The work still to be done is:
- [ ] Upgrade to KfP 2.X
- [ ] Generate output manifest based on input manifest and component spec
- [ ] Split component into load and transform subclasses
- [ ] Optimize the `load_from_hub` component
- [ ] Merge loaded subsets into a single dataframe
- [ ] Remove hardcoded `"gcs://"` and `".parquet"`
- [ ] Automatically create the Kubeflow component yaml
- [ ] Validate data types of input and output data

I resolved all related comments. Let's address the open comments still in this PR as soon as possible.

express/dataset.py
RobbeSneyders: Let's tackle in a separate PR.

express/dataset.py
NielsRogge: Sorry this was indeed not included yet, this is done now.

express/dataset.py
NielsRogge: Ok!

express/manifest.py
```diff
@@ -95,21 +95,25 @@ def _validate_spec(self) -> None:
...
         """
         specification = {
             "metadata": {
+                "project_name": project_name,
                 "base_path": base_path,
                 "run_id": run_id,
                 "component_id": component_id,
             },
-            "index": {"location": f"/index/{run_id}/{component_id}"},
+            "index": {"location": f"/custom_artifact/{run_id}/{component_id}/index"},
```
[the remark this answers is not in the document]
NielsRogge: The reason I put the names at the end is to easily find them on GCP. If a file ends with something it might be easier to find compared to the beginning, but no strong opinion.

Also, the "custom_artificat" was included here as that is the name of the of the bucket on GCP. This is something that needs to be included in the `base_path`, I assume?

express/schema.py
```diff
@@ -4,6 +4,19 @@
 import enum
 import typing as t
 
+import pyarrow as pa
+
+
+type_to_pyarrow = {
```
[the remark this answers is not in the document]
NielsRogge: This was used to define a `schema` argument when uploading a Dask parquet file to the cloud.

Ideally `dd.to_parquet` works without `schema` argument, I'm trying right now to see if it works without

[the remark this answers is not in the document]
NielsRogge: FYI this is the error when writing Dask subset dataframes to the cloud using `dd.to_parquet`:

Screenshot 2023-04-21 at 15 41 08

express/dataset.py
NielsRogge: Let's add docstrings in a follow-up PR

express/dataset.py
NielsRogge: Let's add docstrings in a follow-up PR, have added typehints

tests/test_manifest.py
NielsRogge: The reason for this was that, when getting the fields of an output subset of the component spec, this just gives me a string:

```
>>> from express.component_spec import ComponentSpec
>>> spec = ComponentSpec("fondant_component.yaml")
>>> spec.output_subsets
mappingproxy({'images': ComponentSubset({'fields': {'data': {'type': 'binary'}, 'width': {'type': 'int16'}, 'height': {'type': 'int16'}}}, 'captions': ComponentSubset({'fields': {'data': {'type': 'utf8'}}}})
>>> spec.output_subsets["images"]
ComponentSubset({'fields': {'data': {'type': 'binary'}, 'width': {'type': 'int16'}, 'height': {'type': 'int16'}}}
>>> spec.output_subsets["images"].fields["data"]
Field(name='data', type='binary')
>>> spec.output_subsets["images"].fields["data"].type
'binary'
```
It doesn't give me an object of type `Type`

express/manifest.py
NielsRogge: Have addressed this by rolling back

examples/pipelines/simple_pipeline/components/load_from_hub/src/main.py
PhilippeMoussalli: Yes this could work, the ids might not be unique in this case so we need to make sure to use both columns for joining/filtering.

express/component_spec.py
PhilippeMoussalli: I think it's fine for now since we're not using any other data types or the current testing pipeline. This is currently used to convert the data type in the argument parser but won't be needed when upgrading to kfpv2. 

Will add a TODO.

GeorgesLorre:
> Let's try to get this PR merged and do follow-up work in separate PRs.
> 
> The work still to be done is:
> 
> * [ ]  Upgrade to KfP 2.X
> * [ ]  Generate output manifest based on input manifest and component spec
> * [ ]  Split component into load and transform subclasses
> * [ ]  Optimize the `load_from_hub` component
> * [ ]  Merge loaded subsets into a single dataframe
> * [ ]  Remove hardcoded `"gcs://"`, `".parquet"`, and project name
> * [ ]  Automatically create the Kubeflow component yaml
> * [ ]  Validate data types of input and output data
> 
> I resolved all related comments. Let's address the open comments still in this PR as soon as possible.

I added these on the Trello board so we can distribute and follow up

RobbeSneyders (approved):

PR #34, opened 2023-04-21
Implement Fondant pipeline class

PhilippeMoussalli:
This PR adds an FondantPipeline class that aims to compile the defined pipelines and validate if they are defined in the correct order (all subsets are present when needed based on the ordering of the components).

To do so,  We introduce a new data class `FondantComponentOp` which extends the `FondantComponent` class and includes the argument values that need to be passed to the components as well as optional configurations (GPU, nodepool assignment, ephemeral volume, ...).


.github/workflows/pipeline.yaml
```diff
@@ -22,7 +22,7 @@ jobs:
         run: |
           pip install --upgrade pip
           pip install poetry==1.4.0
-          poetry install
+          poetry install --extras pipelines
```
RobbeSneyders, line 25: We can change this to `--all-extras` so it's more future proof.

fondant/component_spec.py
```diff
@@ -277,6 +277,10 @@ def to_file(self, path: t.Union[str, Path]) -> None:
                 sort_keys=False,
             )
 
+    def to_text(self) -> str:
```
RobbeSneyders, line 280: Would prefer `to_string` as name.

fondant/pipeline.py
```diff
@@ -0,0 +1,311 @@
...
+    p_volumes: t.Optional[t.Dict[str, k8s_client.V1Volume]] = None
+    ephemeral_storage_size: t.Optional[str] = None
+
+    def __post_init__(self):
+        with open(self.component_spec_path, encoding="utf-8") as file_:
+            specification = yaml.safe_load(file_)
+        super().__init__(specification)
+
+
+class FondantPipeline:
```
RobbeSneyders, line 62: This class seems to be a combination of a `Client` and a `Pipeline`. I would either split it, or abstract the `Client` away even more.

Eg. the following methods don't really make sense to me, but could be converted:
- `pipeline.get_pipeline_id(pipeline_name)`  --> `pipeline.id`
- `pipeline.get_pipeline_version_ids(pipeline_id)` --> `pipeline.version_ids`
- `pipeline.compile_pipeline(..., components, ...)` --> `pipeline.compile(...)

It would be good to start from the user interface of the pipeline and work backwards.

RobbeSneyders (commented):
Thanks @PhilippeMoussalli 

I think we mainly need to check here which user interface we want to provide. I think that will give a better view on how we need to implement the `Pipeline` class.

fondant/pipeline.py
PhilippeMoussalli: Thanks for the feedback. I went with the approach of separating both classes since it's the one that makes the most sense. The class and examples are now updated

examples/pipelines/simple_pipeline/simple_pipeline_fondant.py
```diff
@@ -0,0 +1,39 @@
...
+)
+
+pipeline = FondantPipeline(pipeline_name=pipeline_name, base_path=pipeline_base_path)
+
+pipeline.add_op(load_from_hub_op)
+pipeline.add_op(image_filtering_op, dependency=load_from_hub_op)
+
+pipeline.compile_pipeline()
+
+client.upload_pipeline(pipeline_name=pipeline_name)
```
RobbeSneyders, line 39: I would expect us to pass the pipeline object her instead.

RobbeSneyders (commented):
Thanks for the changes @PhilippeMoussalli! 
The interface is looking a lot nicer already. I left some more comments on the example pipeline definition, but they are minor. I haven't yet looked into the implementation again.

examples/pipelines/simple_pipeline/simple_pipeline_fondant.py
```diff
@@ -0,0 +1,39 @@
...
+    arguments={
+        "min_width": ImageFilterConfig.MIN_WIDTH,
+        "min_height": ImageFilterConfig.MIN_HEIGHT,
+    },
+)
+
+pipeline = FondantPipeline(pipeline_name=pipeline_name, base_path=pipeline_base_path)
+
+pipeline.add_op(load_from_hub_op)
+pipeline.add_op(image_filtering_op, dependency=load_from_hub_op)
```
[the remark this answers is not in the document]
PhilippeMoussalli: This can be easily extensible to be a list of components (added it as a TODO in the pipeline class) but for now I kept it as one argument since the functionalities for parallel executions within Fondant are still missing (e.g. manifest evolution)

examples/pipelines/simple_pipeline/simple_pipeline_fondant.py
PhilippeMoussalli: good call, I'll change it

examples/pipelines/simple_pipeline/simple_pipeline_fondant.py
```diff
@@ -0,0 +1,39 @@
...
+
+pipeline.compile_pipeline()
```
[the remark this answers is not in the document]
PhilippeMoussalli: sure, makes more sense

examples/pipelines/simple_pipeline/components/image_filtering/Dockerfile
```diff
@@ -5,23 +5,13 @@ RUN apt-get update && \
...
-# Adding the package path to local
-ENV PATH $PATH:/usr/local/gcloud/google-cloud-sdk/bin
-
 # install requirements
 COPY requirements.txt /
 RUN pip3 install --no-cache-dir -r requirements.txt
 
 # Copy over src-files of the component
 COPY src /src
+COPY fondant_component.yaml /src
```
RobbeSneyders, line 14: We should keep the same directory structure in the docker image so we can run the same code both locally and in the docker image.

examples/pipelines/simple_pipeline/simple_pipeline_fondant.py
```diff
@@ -0,0 +1,42 @@
+"""Pipeline used to create a stable diffusion dataset from a set of given images."""
+import logging
+
+from config.general_config import KubeflowConfig
+from config.components_config import (
+    LoadFromHubConfig,
+    ImageFilterConfig,
+)
+
+from fondant.pipeline import FondantComponentOp, FondantPipeline, FondantClient
```
GeorgesLorre, line 10: Everything is called Fondant, might be a bit much

examples/pipelines/simple_pipeline/simple_pipeline_fondant.py
```diff
@@ -0,0 +1,42 @@
+"""Pipeline used to create a stable diffusion dataset from a set of given images."""
+import logging
+
+from config.general_config import KubeflowConfig
```
GeorgesLorre, line 4: There is also a `config.py` file that looks the same as the `general_config.py`

examples/pipelines/simple_pipeline/simple_pipeline_fondant.py
```diff
@@ -0,0 +1,42 @@
...
+# Define component ops
+load_from_hub_op = FondantComponentOp(
+    component_spec_path="components/load_from_hub/fondant_component.yaml",
+    arguments={"dataset_name": LoadFromHubConfig.DATASET_NAME},
+)
+
+image_filtering_op = FondantComponentOp(
+    component_spec_path="components/image_filtering/fondant_component.yaml",
+    arguments={
+        "min_width": ImageFilterConfig.MIN_WIDTH,
```
GeorgesLorre, line 32: Can we avoid having to pass individual args ?

fondant/pipeline.py
```diff
@@ -0,0 +1,447 @@
...
+            run_name: The name of the run. If not provided, the pipeline's name with 'run' appended
+             will be used.
+            experiment_name: The name of the experiment where the pipeline will be run.
+            Default is 'Default'.
+        """
+        pipeline.compile_pipeline()
+
+        try:
+            experiment = self.client.get_experiment(experiment_name=experiment_name)
+        except ValueError:
```
GeorgesLorre, line 434: Ideally the error would be more specific

GeorgesLorre (approved):
Left some small remarks

But nice work ! (love the tests and logging)

examples/pipelines/simple_pipeline/simple_pipeline_fondant.py
RobbeSneyders: Yes, would drop it everywhere (and maybe only keep an F). But we can do that in a separate PR.

fondant/pipeline.py
```diff
@@ -0,0 +1,447 @@
...
+            run_id: the run id of the component
+        """
+        load_component = True
+        load_component_name = list(self._graph.keys())[0]
+
+        # Create initial manifest
+        manifest = Manifest.create(
+            base_path=self.base_path, run_id=run_id, component_id=load_component_name
+        )
+        for _, operation_specs in self._graph.items():
```
RobbeSneyders, line 168: You can use `.values()` so you don't have the unused parameter. Also don't think `operation_specs` is the best name here. Maybe just `operation`.

fondant/pipeline.py
```diff
@@ -0,0 +1,447 @@
...
+                fondant_component_op.component_spec_path
+            )
+            manifest = manifest.evolve(component_spec)
+            if not load_component:
+                # Check subset exists
+                for (
+                    component_subset_name,
+                    component_subset,
+                ) in fondant_component_op.input_subsets.items():
+                    manifest_subset = manifest.subsets
```
RobbeSneyders, line 180: The assignment to a singular name is a bit confusing. But we can probably lose the assignment completely.

fondant/pipeline.py
```diff
@@ -0,0 +1,447 @@
...
+        # Create initial manifest
+        manifest = Manifest.create(
+            base_path=self.base_path, run_id=run_id, component_id=load_component_name
+        )
+        for _, operation_specs in self._graph.items():
+            fondant_component_op = operation_specs["fondant_component_op"]
+            component_spec = FondantComponentSpec.from_file(
+                fondant_component_op.component_spec_path
+            )
+            manifest = manifest.evolve(component_spec)
```
RobbeSneyders, line 173: I don't think we need to do this here already, as we'll be comparing the component spec input subsets to its output manifest below.

fondant/pipeline.py
```diff
@@ -0,0 +1,447 @@
...
+         associated fields match and are invoked in the correct order.
+
+        Raises:
+            InvalidPipelineDefinition: If a component is trying to invoke a subset that is not
+             defined or created in previous components, or if an invoked subset's schema does not
+              match the previously created subset definition.
+            base_path: the base path where to store the pipelines artifacts
+            run_id: the run id of the component
+        """
+        load_component = True
```
RobbeSneyders, line 161: This assumes that a pipeline always starts with a load component, which I don't think should be the case. What if you eg. want to run 2 fondant pipelines after each other? But we can keep this assumption for now and tackle it later.

fondant/pipeline.py
```diff
@@ -0,0 +1,447 @@
...
+                                f" field is currently defined with the following schema:\n"
+                                f"{manifest_fields[field_name]}\n"
+                                f"The current component to trying to invoke it with this schema:\n"
+                                f"{subset_field}"
+                            )
+            load_component = False
+
+        logger.info("All pipeline component specifications match.")
+
+    def compile_pipeline(self):
```
RobbeSneyders, line 218: Would propose to rename to `compile()`, so it's called as `pipeline.compile()`

fondant/pipeline.py
```diff
@@ -0,0 +1,447 @@
...
+        @dsl.pipeline(name=self.name, description=self.description)
+        def pipeline():
+            # TODO: check if we want to have the manifest path empty for loading component or remove
+            #  it completely from the loading component
+            # TODO: check if we want to have the metadata arg empty for transform component or
+            #  remove it completely from the transform component
+            manifest_path = ""
+            metadata = ""
+            previous_component_task = None
+            for _, operation_specs in self._graph.items():
```
RobbeSneyders, line 280: You can use `.values()` so you don't have the unused parameter. Also don't think `operation_specs` is the best name here. Maybe just `operation`.

fondant/pipeline.py
```diff
@@ -0,0 +1,447 @@
...
+            pipeline_name: The name of the pipeline.
+            pipeline_description: Optional description of the pipeline.
+        """
+        self.base_path = base_path
+        self.name = pipeline_name
+        self.description = pipeline_description
+        self.package_path = f"{pipeline_name}.tgz"
+        self._graph: t.OrderedDict[str, t.Any] = OrderedDict()
+
+    def add_op(
```
RobbeSneyders, line 90: What happens if more than one op is currently defined without dependencies?

fondant/pipeline.py
```diff
@@ -0,0 +1,447 @@
...
+
+            logger.info(
+                f"Pipeline {pipeline_name} already exists. Deleting old pipeline..."
+            )
+        else:
+            logger.info(f"No existing pipeline under `{pipeline_name}` name was found.")
+
+    def compile_and_upload(
+        self,
+        *,
```
RobbeSneyders, line 381: Nit, but I think it makes more sense to have `pipeline` as a positional argument.

tests/test_pipeline.py
```diff
@@ -0,0 +1,121 @@
...
+@pytest.mark.parametrize(
+    "valid_pipeline_example",
+    [
+        (
+            "example_1",
+            ["first_component.yaml", "second_component.yaml", "third_component.yaml"],
+        ),
+    ],
+)
+def test_valid_pipeline(default_pipeline_args, valid_pipeline_example, tmp_path):
```
RobbeSneyders, line 30: Your test here doesn't catch the issue that you're validating against the output manifest because each input subset is always propagated as an output subset. You might want to create an example where this is not the case.

RobbeSneyders (commented):
Thanks @PhilippeMoussalli 

Great work. Left some comments, but that's to be expected from this big of a feature :)

Did you get inspiration from an existing framework for the current interface to build the pipeline (the `.add_opp()` method etc.)? I see a lot of complex work on extending the internals in the future to support non-linear DAGs, so it would be great if we can get inspiration from the same framework for that.

examples/pipelines/simple_pipeline/components/image_filtering/src/main.py
```diff
@@ -22,26 +22,26 @@ def transform(
...
-            df: Dask dataframe
+            dataframe: Dask dataframe
             min_width: min width to filter on
             min_height: min height to filter on
 
         Returns:
             dataset
         """
-        logger.info("Length of the dataframe before filtering:", len(dataframe))
+        logger.info("Length of the dataframe before filtering: %s", len(dataframe))
```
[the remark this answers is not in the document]
PhilippeMoussalli: maybe let's keep this for now for the simple pipeline tests. I created a ticket for it [here](https://github.com/orgs/ml6team/projects/1/views/1?pane=issue&itemId=27796944)

examples/pipelines/simple_pipeline/simple_pipeline_fondant.py
PhilippeMoussalli: Makes sense indeed. Created a [ticket](https://github.com/orgs/ml6team/projects/1/views/1?pane=issue&itemId=27797215) for it

examples/pipelines/simple_pipeline/simple_pipeline_fondant.py
PhilippeMoussalli: I think it's better to keep the general configs in `config.py` at the `pipelines` directory level instead of replicating it inside each pipeline example since it contains cloud and cluster specific configs applicable to all pipelines. I'll make the changes

```
pipelines
--config.py
--pipeline_1
----components_config.py
--pipeline_2
----components_config.py
```

examples/pipelines/simple_pipeline/simple_pipeline_fondant.py
NielsRogge: Why not just call them `ComponentOp`, `Pipeline`, etc?

PhilippeMoussalli: I wanted to make it a bit more explicit to avoid confusion with KFP native objects since the Fondant pipeline is essentially a wrapper around it

RobbeSneyders:
I think we will need to be able to create a full non-linear DAG ourselves for features such as pipeline validation and caching / partial execution.

The repo you link could be an example, but it's not very popular. I'm wondering if we shouldn't just replicate the Kubeflow pipelines interface instead.

GeorgesLorre:
> 

I also want to mention the way airflow does it: https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/dags.html#task-dependencies

I had a lot of issues with airflow but defining complex graphs was always easy enough

PhilippeMoussalli:
> Wouldn't start overloading binary operators just yet, but the general Python interface might indeed make sense.
> 
> ```python
> second_task.set_upstream(first_task)
> joined_task.set_upstream(parallel_task1, parallel_task2)
> ```
> 
> is very similar to the one from KfP
> 
> ```python
> second_task.after(first_task)
> joined_task.after(parallel_task1, parallel_task2)
> ```
> 
> I like `after()` a bit more than `set_upstream()` since it's more clear (I had to look up if I needed `set_upstream()` or `set_downstream()` for Airflow).
> 
> But in any case, they are chaining tasks, not registering tasks on a pipeline object. Which seems to be the approach most often used by similar frameworks.

So currently we're able to build a non-linear DAG as follows: 
``` 
pipeline.add_op(first_component)
pipeline.add_op(parallel_component_1, depends_on=first_component)
pipeline.add_op(parallel_component_2, depends_on=first_component)
pipeline.add_op(second, depends_on=[parallel_component_1,parallel_component_2])
```

Should I then change the interface to do chain the tasks and then submit them to the pipeline?

RobbeSneyders:
I guess in essence, it boils down to the same. Let's keep it at least for now so we can work towards merging the PR, and then we can still see if we want to change it in a separate PR.

fondant/component_spec.py
```diff
@@ -276,13 +283,17 @@ def to_file(self, path: t.Union[str, Path]) -> None:
...
+        return json.dumps(self._specification)
+
     @property
     def input_arguments(self) -> t.Mapping[str, Argument]:
         """The input arguments of the component as an immutable mapping."""
         return types.MappingProxyType(
             {
                 info["name"]: Argument(
-                    name=info["name"],
+                    name=info["name"].replace("_", "-"),
```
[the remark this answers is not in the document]
PhilippeMoussalli: small nit but I noticed that the arguments were passed with a mix of underscore and upperscore (e.g. --input_manifest_path) so I changed this. However this might not be the best place to do this. I'll move the `replace` method to the `_add_and_parse_args` function

fondant/pipeline.py
PhilippeMoussalli: I think that's an edge case that I need to address

PhilippeMoussalli: addressed and added some tests

fondant/pipeline.py
PhilippeMoussalli: yeah indeed, I'v added a todo to tackle this later on but for now it should be fine

fondant/pipeline.py
PhilippeMoussalli: That's the only error that KFP throws when an experiment is not created, Added a logging to make it more specific to the user that a new experiment is created

tests/test_pipeline.py
PhilippeMoussalli: I'v adjusted the examples, now `captions` are outputted in the first component and are not propagated in the second one. I later invoke it in the third one

fondant/component_spec.py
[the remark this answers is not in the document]
RobbeSneyders: I'm not sure if we even need to do this at all. This might be surprising for the users, instead of just using the name that they provided.

fondant/pipeline.py
RobbeSneyders: No, we shouldn't evolve it yet for the loading component, as we'll then be checking the output manifest against the input subsets specified in the component spec.

tests/test_pipeline.py
RobbeSneyders: They are still propagated, and you will not catch the issue with this test. I'll add a commit with a test that fails to show the issue.

fondant/pipeline.py
PhilippeMoussalli: alright i'll move it below

fondant/component_spec.py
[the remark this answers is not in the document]
PhilippeMoussalli: The users still define argument with underscores in both the component function definition and specification. This only applies when running the component from the command line similar to [our boilerplate example](https://bitbucket.org/ml6team/nimbus/src/c798c261812e90c81e9aabab3b33ca603c09476f/src/_nimbus/boilerplates/mlpipelines/gcp/boilerplate/example_pipeline.py)

Not sure if there is a convention around this but I'm used to command line arguments with a mixture of hyphens and underscores always trip me up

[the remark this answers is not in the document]
RobbeSneyders: Agree that hyphens is nicer, but I think matching the arguments in the specification file is more important.

tests/example_pipelines/valid_pipeline/example_1/third_component.yaml
```diff
@@ -0,0 +1,30 @@
...
+  embeddings:
+    fields:
+      data:
+        type: binary
+
+output_subsets:
+  images:
+    fields:
+      data:
+        type: binary
```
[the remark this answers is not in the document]
RobbeSneyders: Perfect, thanks!

RobbeSneyders (approved):

PR #36, opened 2023-04-24
Extract kubeflow logic from ComponentSpec to KubeflowComponent

RobbeSneyders:
This PR extracts all the kubeflow logic from the `ComponentSpec` to the `KubeflowComponent` so the `ComponentSpec` is a minimal representation of the fondant component specification. I also brought the classes a bit more in line with the `Manifest` class.

This is preparation for the work to deduct the output manifest from the input manifest and component spec only.

### DIFF
--- a/fondant/component_spec.py
+++ b/fondant/component_spec.py
@@ -6,7 +6,7 @@ import pkgutil
 import types
 import typing as t
 from pathlib import Path
-from dataclasses import dataclass, asdict
+from dataclasses import dataclass
 
 import jsonschema.exceptions
 from jsonschema import Draft4Validator
@@ -38,7 +38,7 @@ kubeflow2python_type = {
 
 
 @dataclass
-class KubeflowInput:
+class Argument:
     """
     Kubeflow component input argument
     Args:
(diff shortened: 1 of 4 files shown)


fondant/component_spec.py
```diff
@@ -38,7 +38,7 @@
 
 
 @dataclass
-class KubeflowInput:
+class Argument:
     """
     Kubeflow component input argument
```
NielsRogge, line 43:
```suggestion
    Kubeflow component argument
```
I assume it represents both input and output arguments

fondant/component_spec.py
```diff
@@ -196,101 +103,41 @@ def _validate_spec(self) -> None:
...
-            inputs.append(
-                KubeflowInput(
-                    name=arg_name.strip(),
-                    description=arg_info["description"].strip(),
-                    type=python2kubeflow_type[arg_info["type"].strip()],
-                )
-            )
+    @classmethod
+    def from_file(cls, path: str) -> "ComponentSpec":
+        """Load the manifest from the file specified by the provided path"""
```
NielsRogge, line 112: Do you mean loading the component spec instead of the manifest here? As the manifest doesn't contain thiings like `input_subsets` and `output_subsets`

fondant/component_spec.py
```diff
@@ -196,101 +103,41 @@ def _validate_spec(self) -> None:
...
-            name=self.yaml_spec["name"],
-            description=self.yaml_spec["description"],
-            image=self.yaml_spec["image"],
-            inputs=inputs,
-            outputs=outputs,
-        )
-
-        return kubeflow_component.specification
+    def to_file(self, path) -> None:
+        """Dump the manifest to the file specified by the provided path"""
```
NielsRogge, line 118: Same comment here

fondant/component_spec.py
```diff
@@ -300,58 +147,146 @@ def output_subsets(self) -> t.Mapping[str, ComponentSubset]:
...
+
+    @property
+    def kubeflow_specification(self) -> "KubeflowComponent":
+        return KubeflowComponent.from_fondant_component(self)
+
+    def __repr__(self) -> str:
+        return f"{self.__class__.__name__}({self._specification!r}"
+
+
+class KubeflowComponent:
```
NielsRogge, line 181:
```suggestion
class KubeflowSpec:
```
To be consistent with `FondantSpec` ;)

fondant/component_spec.py
```diff
@@ -300,58 +147,146 @@ def output_subsets(self) -> t.Mapping[str, ComponentSubset]:
...
+
+    Args:
+        specification: The component specification as a Python dict
+    """
+
+    def __init__(self, specification: t.Optional[dict]) -> None:
+        self._specification = specification
+
+    @classmethod
+    def from_fondant_component(
```
NielsRogge, line 193:
```suggestion
    def from_fondant_spec(
```

fondant/component_spec.py
```diff
@@ -300,58 +147,146 @@ def output_subsets(self) -> t.Mapping[str, ComponentSubset]:
...
+        for arg in args:
+            arg_name = arg.name.replace("-", "_").strip()
+            arg_name_cmd = f'--{arg.name.replace("_", "-")}'.strip()
+
+            dumped_args.append(arg_name_cmd)
+            dumped_args.append({"inputValue": arg_name})
+
+        return dumped_args
+
+    def to_file(self, path: t.Union[str, Path]) -> None:
```
PhilippeMoussalli, line 252: I noticed that when compiling the kfp pipeline you can pass in directly a json string instead of the yaml file. So might be worth switching over to that approach. But this should be handled in separate PR related when automating the pipeline definition and compilation

PhilippeMoussalli (approved):
Really nice refactoring :+1:

Thanks!

NielsRogge (approved):
Thanks for improving!

fondant/component_spec.py
RobbeSneyders: Ok, the json is available under `_spec` already, so should be easy to expose.

PR #37, opened 2023-04-25
Remove cloud specific prefix

NielsRogge:
This PR removes the need for a specific prefix, as we can just include that in the base path of the configuration.

### DIFF
--- a/fondant/dataset.py
+++ b/fondant/dataset.py
@@ -6,6 +6,7 @@ It also defines the FondantComponent class, which uses the FondantDataset class 
 from abc import abstractmethod
 import argparse
 import json
+import logging
 from pathlib import Path
 from typing import List, Mapping
 
@@ -14,6 +15,8 @@ import dask.dataframe as dd
 from fondant.component_spec import FondantComponentSpec, kubeflow2python_type
 from fondant.manifest import Manifest
 from fondant.schema import Type, Field
+
+logger = logging.getLogger(__name__)
 
 
 class FondantDataset:
@@ -36,6 +39,8 @@ class FondantDataset:
         subset = self.manifest.subsets[name]
         # get remote path
         remote_path = subset.location
+
+        logger.info(f"Loading subset {name} with fields {fields}...")
 
         df = dd.read_parquet(
             remote_path,
--- a/fondant/dataset.py
+++ b/fondant/dataset.py
@@ -34,11 +34,11 @@ class FondantDataset:
     def _load_subset(self, name: str, fields: List[str]) -> dd.DataFrame:
         # get subset from the manifest
         subset = self.manifest.subsets[name]
-        # TODO remove prefix
-        location = "gcs://" + subset.location
+        # get remote path
+        remote_path = subset.location
 
         df = dd.read_parquet(
-            location,
+            remote_path,
             columns=fields,
         )
 
@@ -58,9 +58,8 @@ class FondantDataset:
         # return df
 
     def _upload_index(self, df: dd.DataFrame):
-        # get location
-        # TODO remove prefix and suffix
-        remote_path = "gcs://" + self.manifest.index.location
+        # get remote path
+        remote_path = self.manifest.index.location
 
         # upload to the cloud
         dd.to_parquet(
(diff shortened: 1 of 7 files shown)


fondant/pipeline_utils.py
```diff
@@ -29,15 +29,15 @@ def compile_and_upload_pipeline(
...
-            client.delete_pipeline(existing_pipeline.id)
+    # existing_pipelines = client.list_pipelines(page_size=100).pipelines
+    # for existing_pipeline in existing_pipelines:
+    #     if existing_pipeline.name == pipeline_name:
+    #         # Delete existing pipeline before uploading
+    #         logger.warning(
+    #             f"Pipeline {pipeline_name} already exists. Deleting old pipeline..."
+    #         )
+    #         client.delete_pipeline_version(existing_pipeline.default_version.id)
+    #         client.delete_pipeline(existing_pipeline.id)
```
NielsRogge, line 40: This is a temporary fix in order to compile the pipeline.

RobbeSneyders (approved):
Thanks @NielsRogge. LGTM.

fondant/dataset.py
```diff
@@ -31,36 +34,74 @@ def __init__(self, manifest: Manifest):
...
+        # add subset prefix to columns
+        df = df.rename(
+            columns={
+                col: name + "_" + col for col in df.columns if col not in index_fields
+            }
+        )
+
+        return df
+
+    def _load_index(self):
```
GeorgesLorre, line 72: Nice ! I can use this for 
image

fondant/dataset.py
```diff
@@ -31,36 +34,74 @@ def __init__(self, manifest: Manifest):
...
+        if index is None:
+            index_df = self._load_index()
+            ids = index_df["id"].compute()
+            sources = index_df["source"].compute()
+            df = df[df["id"].isin(ids) & df["source"].isin(sources)]
+
+        # add subset prefix to columns
+        df = df.rename(
+            columns={
+                col: name + "_" + col for col in df.columns if col not in index_fields
```
GeorgesLorre, line 66: For joining all the data into 1 dataframe we will need to join on the index which gets filtered out here.

GeorgesLorre (approved):

fondant/dataset.py
NielsRogge: The index is not filtered out, this operation only renames the non-index columns

fondant/dataset.py
```diff
@@ -31,36 +34,74 @@ def __init__(self, manifest: Manifest):
...
-            location,
+            remote_path,
             columns=fields,
         )
 
+        # filter on default index of manifest if no index is provided
+        if index is None:
+            index_df = self._load_index()
+            ids = index_df["id"].compute()
+            sources = index_df["source"].compute()
```
PhilippeMoussalli, line 60: So there is no workaround for having to bring those in memory before filtering?

fondant/dataset.py
```diff
@@ -31,36 +34,74 @@ def __init__(self, manifest: Manifest):
...
         for name, subset in spec.input_subsets.items():
             fields = list(subset.fields.keys())
             subset_df = self._load_subset(name, fields)
-            subsets.append(subset_df)
+            subset_dfs.append(subset_df)
 
         # TODO this method should return a single dataframe with column_names called subset_field
         # TODO add index
-        # df = concatenate_datasets(subsets)
+        df = dd.concat(subset_dfs)
```
PhilippeMoussalli, line 96: Should we join on index rather than concatenating? Are we guaranteed that all the subsets will have aligned index and source and that they are all of the same length?

PhilippeMoussalli (approved):
LGTM, left a few open questions that we might want to follow up on

fondant/dataset.py
NielsRogge: Yes this can be addressed in a separate PR (cc @GeorgesLorre for the "return single dataframe" ticket). We might want to leverage `dd.merge` here

fondant/dataset.py
NielsRogge: In order to filter a dataframe based on an index, I assume that you need to have the entire index present.

Correct me if I'm wrong @GeorgesLorre @RobbeSneyders

fondant/dataset.py
PhilippeMoussalli: alright thanks for clarifying :)

fondant/dataset.py
RobbeSneyders: I would have to look into Dask in more detail before I can give input on this.

PR #38, opened 2023-04-25
Update manifest schema and move test manifests to files

RobbeSneyders:
This PR includes 3 small changes:
- Add `run_id` and `component_id` as required metadata to the manifest schema
- Move test manifest to files similar to the component specs
- Leverage relative files in tests so they can be run from different working directories


NielsRogge (approved):
Thanks for adding

PR #41, opened 2023-04-25
Evolve manifest based on component spec

RobbeSneyders:
Don't be alarmed by the size of this PR, it contains a lot of test cases :sweat_smile: 

This PR adds functionality to generate the output manifest based on the input manifest and component specification alone. By deducting the output manifest this way, it can be used for validation of the output data in the component, and for static validation of the pipeline before execution.

I implemented this as an `.evolve()` method on the manifest, since the input manifest is always the starting point. This will also allow for nice chaining in the future. The method is quite long, but I don't think splitting it up will improve readability.

This PR only provides the functionality and tests. As a next step, we'll have to integrate this into the `Dataset` class to actually leverage the functionality for output validation.

### DIFF
--- a/fondant/component_spec.py
+++ b/fondant/component_spec.py
@@ -13,7 +13,7 @@ from jsonschema import Draft4Validator
 from jsonschema.validators import RefResolver
 
 from fondant.exceptions import InvalidComponentSpec
-from fondant.schema import Field
+from fondant.schema import Field, Type
 
 # TODO: Change after upgrading to kfp v2
 # :https://www.kubeflow.org/docs/components/pipelines/v2/data-types/parameters/
@@ -71,10 +71,14 @@ class ComponentSubset:
     def fields(self) -> t.Mapping[str, Field]:
         return types.MappingProxyType(
             {
-                name: Field(name=name, type=field["type"])
+                name: Field(name=name, type=Type[field["type"]])
                 for name, field in self._specification["fields"].items()
             }
         )
+
+    @property
+    def additional_fields(self) -> bool:
+        return self._specification.get("additionalFields", True)
 
 
 class FondantComponentSpec:
@@ -138,6 +142,7 @@ class FondantComponentSpec:
             {
                 name: ComponentSubset(subset)
                 for name, subset in self._specification["input_subsets"].items()
+                if name != "additionalSubsets"
             }
         )
 
@@ -148,6 +153,7 @@ class FondantComponentSpec:
             {
                 name: ComponentSubset(subset)
                 for name, subset in self._specification["output_subsets"].items()
+                if name != "additionalSubsets"
             }
         )
 
--- a/tests/example_specs/evolution_examples/7/output_manifest.json
+++ b/tests/example_specs/evolution_examples/7/output_manifest.json
@@ -1,0 +1,20 @@
+{
+  "metadata": {
+    "base_path": "gs://bucket",
+    "run_id": "12345",
+    "component_id": "67890"
+  },
+  "index": {
+    "location": "/index"
+  },
+  "subsets": {
+    "images": {
+      "location": "/images",
+      "fields": {
+        "data": {
+          "type": "utf8"
+        }
+      }
+    }
+  }
+}
--- a/tests/test_manifest_evolution.py
+++ b/tests/test_manifest_evolution.py
@@ -1,0 +1,35 @@
+import json
+from pathlib import Path
+
+import pytest
+import yaml
+
+from fondant.component_spec import FondantComponentSpec
+from fondant.manifest import Manifest
+
+
+examples_path = Path(__file__).parent / "example_specs/evolution_examples"
+
+
+@pytest.fixture
+def input_manifest():
+    with open(examples_path / "input_manifest.json") as f:
+        return json.load(f)
+
+
+def examples():
+    """Returns examples as tuples of component and expected output_manifest"""
+    for directory in (f for f in examples_path.iterdir() if f.is_dir()):
+        with open(directory / "component.yaml") as c, open(directory / "output_manifest.json") as o:
+            yield yaml.safe_load(c), json.load(o)
+
+
+@pytest.mark.parametrize("component_spec,output_manifest", list(examples()))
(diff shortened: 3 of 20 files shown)


tests/example_specs/evolution_examples/7/output_manifest.json
```diff
@@ -0,0 +1,20 @@
...
+    "base_path": "gs://bucket",
+    "run_id": "12345",
+    "component_id": "67890"
+  },
+  "index": {
+    "location": "/index"
+  },
+  "subsets": {
+    "images": {
+      "location": "/images",
```
RobbeSneyders, line 12: There is still some logic missing to update the locations of the index and subsets when they need to be rewritten. But I suggest we tackle this separately as part of https://trello.com/c/ksYMGyOw.

fondant/component_spec.py
```diff
@@ -71,11 +71,15 @@ def __repr__(self) -> str:
     def fields(self) -> t.Mapping[str, Field]:
         return types.MappingProxyType(
             {
-                name: Field(name=name, type=field["type"])
+                name: Field(name=name, type=Type[field["type"]])
```
NielsRogge, line 74: Was also wondering here, we return a dictionary with names and fields, but the fields themselves already contain a name. Is it worth having this twice?

fondant/component_spec.py
```diff
@@ -71,11 +71,15 @@ def __repr__(self) -> str:
...
                 for name, field in self._specification["fields"].items()
             }
         )
 
+    @property
+    def additional_fields(self) -> bool:
```
NielsRogge, line 80: What's the use case for this? 

(same question for AdditionalSubsets - why would a component want to deal with subsets it doesn't need?)

RobbeSneyders: Have a look at the test examples, hopefully they provide some clarity on this.

Some examples where this is useful:
- additionalSubsets
  - An embedder component adds an embedding subset based on images. It doesn't care if there are additional subsets (eg. captions), so it defines `additionalSubsets` to be `true` (the default)
  - A LAION retrieval component adds rows to the `images`, `captions`, and `embeddings` subsets. It should not allow any additional subsets, as they will no longer be available for every entry in the index. So it defines `additionalSubsets` to be `false`.
- additionalFields
  - A component calculating the `width` and `height` of an image doesn't care if there's other metadata as well, so it defines 'additionalFields` to be `true` (the default)
  - A component that transforms images might invalidate some of the metadata. So it can define `additionalFields` to be `false` and define all the metadata it generates explicitly.

The combination of `additionalSubsets` and `additionalFields` for both input and output provides enough flexibility to unambiguously define the output subsets and their fields.

PhilippeMoussalli: It's also not very clear to me why we would need `Additionalsubsets`. All subsets are going to have to be defined in the ComponentSpecs at some point. So subsets that are not specified should just not be loaded. What does `Additionalsubsets` further enable?

fondant/component_spec.py
RobbeSneyders: True. Returning a dict instead of a list allows lookup by name, which is definitely useful. Adding the name to the `Field`, makes it a complete representation of the field, which I believe is logical as well.

Currently we're not using the same approach for the `Subset`s though, so it would at least make sense to align on an approach.

fondant/component_spec.py
NielsRogge: Ok, but with `additionalSubsets` being `True` by default, it means that a lot more data gets loaded and provided to the component, which might slow things down? e.g. let's say you have 3 subsets in the manifest, "images", "captions" and "embeddings", and a component that only needs the "images" will get all embeddings as well returned

RobbeSneyders: No, we would only load subsets and fields explicitly defined in the component spec. `additionalSubsets` and `additionalFields` are only needed to unambiguously define the output manifest of the component.

RobbeSneyders: Have a look at test examples 1, 2, 3, or 4, 5, 6. Their component specs are the same except for differences in `additionalSubsets` and `additionalFields`. Without this information, how would you define which of the output manifests would be the correct one?

tests/test_manifest_evolution.py
```diff
@@ -0,0 +1,35 @@
...
+            yield yaml.safe_load(c), json.load(o)
+
+
+@pytest.mark.parametrize("component_spec,output_manifest", list(examples()))
+def test_evolution(input_manifest, component_spec, output_manifest):
+    manifest = Manifest(input_manifest)
+    component = FondantComponentSpec(component_spec)
+    evolved_manifest = manifest.evolve(component)
+
+    print(json.dumps(evolved_manifest._specification, indent=4))
```
NielsRogge, line 33: to be removed

fondant/component_spec.py
NielsRogge: Ok I see, so they define how the component expects the input and output manifest to look like. That makes sense. Thanks for clarifying!

PhilippeMoussalli: Thanks for clarifying.  Another question, if `additionalSubsets`is False (e.g. clip retrieval job) are we then required to populate every other subsets that's currently available and that's available for all other `sources`?

RobbeSneyders: > Ok I see, so they define how the component expects the input and output manifest to look like. That makes sense. Thanks for clarifying!

In a way. It also defines the behavior of the component of course, as the output manifest depends on this.

> Thanks for clarifying. Another question, if `additionalSubsets`is False (e.g. clip retrieval job) are we then required to populate every other subsets that's currently available for all other `sources`?

Not sure if I understand your question. But if `additionalSubsets` is False, subsets not explicitly defined in the component spec will not be propagated to the output manifest.

PhilippeMoussalli: Had another look at the example and I think it's all clear now. Thanks

PhilippeMoussalli (approved):
Thanks Robbe!

GeorgesLorre (approved):
Cool, excited to start using this.

Also the test setup is nice

PR #48, opened 2023-04-26
split fondant component into load and transform

PhilippeMoussalli:
PR that splits the Fondant Component into load and transform component. The user will now call `FondantLoaderComponent` or `FondantTransoromComponent` depending on the type of job that is run. 

A few notes: 

- Both components rely on the `FondantComponentSpec` which requires both `input_subsets` and defines an `input_manifest_path` as required argument (based on `KubeflowComponentSpec`). In practice, the loader component should not require both inputs. Wondering if it's worth splitting the Spec to loader and transform as well and initialize both `FondantLoaderComponent` and  `FondantTransformComponent` with those. 

- Currently only tested for the `FondantLoaderComponent`. I think the `FondantTransformComponent` still needs to be implemented but part of another PR (todos are still present in the code)


examples/pipelines/simple_pipeline/components/load_from_hub/src/main.py
```diff
@@ -11,7 +11,7 @@
 import numpy as np
 from PIL import Image
 
-from fondant.dataset import FondantComponent
+from fondant.dataset import FondantLoaderComponent
```
NielsRogge, line 14:
```suggestion
from fondant.dataset import FondantLoadComponent
```
Let's call them "load" and "transform"

fondant/dataset.py
```diff
@@ -166,82 +167,123 @@ def upload(self, save_path):
...
-            df = dataset.load_data(self.spec)
-            # provide this dataframe to the user
-            df = self.transform(
-                df=df,
-                args=args,
-            )
-            # TODO update index, potentially add new subsets
+        component_arguments = {}
+        kubeflow_component_spec = self.spec.kubeflow_specification
+        component_arguments.update(kubeflow_component_spec.input_arguments)
```
NielsRogge, line 192: just to confirm, the custom `args` of a component are part of the `input_arguments`?

PhilippeMoussalli: yes indeed the kubeflow input arguments are derived from the ComponentSpec of the component which contains the arguments

examples/pipelines/simple_pipeline/components/load_from_hub/src/main.py
PhilippeMoussalli: done :)

fondant/dataset.py
```diff
@@ -166,82 +167,123 @@ def upload(self, save_path):
...
+    @abstractmethod
+    def load(self, args: argparse.Namespace) -> dd.DataFrame:
+        """Abstract method that loads the initial dataframe"""
+
+    def run(self):
+        """
+        Runs the loading component.
+        """
+        # step 1: Add and parse arguments
+        args = self._add_and_parse_args()
```
RobbeSneyders, line 253: Can we lift these common steps to the base class?

fondant/dataset.py
```diff
@@ -166,82 +167,123 @@ def upload(self, save_path):
...
+
+    def __init__(self):
         self.spec = FondantComponentSpec.from_file("fondant_component.yaml")
-        self.type = type
 
-    def run(self) -> dd.DataFrame:
-        """
-        Parses input data, executes the transform, and creates the output manifest.
+    @abstractmethod
+    def _get_manifest(self, args: argparse.Namespace) -> Manifest:
```
RobbeSneyders, line 177: Not a fan of defining abstract methods which are not called by the base class. It imposes a structure on the subclasses without any advantage to them.

fondant/dataset.py
```diff
@@ -166,82 +167,123 @@ def upload(self, save_path):
...
+
+    @abstractmethod
+    def load(self, args: argparse.Namespace) -> dd.DataFrame:
+        """Abstract method that loads the initial dataframe"""
+
+    def run(self):
+        """
+        Runs the loading component.
+        """
+        # step 1: Add and parse arguments
```
RobbeSneyders, line 252: These comments are a bit redundant here. The code is quite self explanatory.

fondant/dataset.py
```diff
@@ -166,82 +167,123 @@ def upload(self, save_path):
...
+    """Abstract base class for a Fondant loader component"""
+
+    def _add_and_parse_args(self) -> argparse.Namespace:
         parser = argparse.ArgumentParser()
+        component_arguments = self.get_component_arguments()
 
-        kubeflow_component = self.spec.kubeflow_specification
+        for arg in component_arguments.values():
+            # Input manifest is not required for loading component
+            if arg.name == "input_manifest_path":
```
RobbeSneyders, line 210: I would expect this arg to not be defined for a Load component instead of having to handle it specifically here. Without this, we can lift this method completely to the base class.

RobbeSneyders (commented):
Thanks @PhilippeMoussalli!

fondant/dataset.py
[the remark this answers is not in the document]
RobbeSneyders: As mentioned below, I think lifting the code that calls this method to the base class is the best solution.

fondant/dataset.py
```diff
@@ -166,82 +167,123 @@ def upload(self, save_path):
...
+        parser.add_argument(
+            "--metadata",
+            type=str,
+            required=True,
+            help="The metadata associated with the pipeline run",
+        )
+
+        return parser.parse_args()
+
+    def _get_manifest(self, args: argparse.Namespace) -> Manifest:
```
NielsRogge, line 232:
```suggestion
    def _create_manifest(self, args: argparse.Namespace) -> Manifest:
```

fondant/dataset.py
PhilippeMoussalli: Fair point, can we just then keep it as Optional by default for both components. Otherwise we might need to rework the `KubeflowComponentSpec` to have different specification for both loader and transform. https://github.com/ml6team/fondant/blob/main/fondant/component_spec.py#:~:text=class-,KubeflowComponentSpec,-%3A
@NielsRogge wondering what's your take on this?

NielsRogge: I feel like the user would only need to be exposed to a single `FondantComponent` class actually. We could figure out that the first component in the pipeline is the loading one and that all others are transform components. We can internally split up the kubeflow spec for loading vs transformation, indeed.

PhilippeMoussalli: Good point. I think we can eventually tackle but maybe in a separate PR.

fondant/dataset.py
PhilippeMoussalli: This is an abstract function that might either get the manifest if it exists or create a new one 
perhaps `load_or_create_manifest` might be a better name

fondant/dataset.py
RobbeSneyders: The interface for both components is different, since we need to provide an input dataframe for the transform components and not for the load components. I don't think it makes sense to try to align these completely.

fondant/dataset.py
NielsRogge: The comments below are helpful to me

fondant/dataset.py
NielsRogge: Ok, we can keep the separation also for the user in that case.

fondant/dataset.py
NielsRogge: Ok, feel free to rename

fondant/dataset.py
RobbeSneyders: I really don't see the value of eg. these comments as the same information is clear from the lines of code. Let's only add inline comments when they provide additional information on the reasoning behind a piece of code or when the logic is too hard to follow in code.

```
# step 2: Get manifest
manifest = self._get_manifest(args)

# step 3: Create dataset
dataset = FondantDataset(manifest)
```

NielsRogge: Yes I meant the comments of the steps below step 3

fondant/dataset.py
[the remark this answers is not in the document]
PhilippeMoussalli: Thanks for the suggestion. Indeed it's much nicer to enforce and define where those abstract methods should be used in the base class. I updated the PR accordingly

fondant/dataset.py
PhilippeMoussalli: Done :)

fondant/dataset.py
PhilippeMoussalli: I removed some of the redundant comments and left the ones that could be useful in providing more clarity

RobbeSneyders (approved):
Thanks @PhilippeMoussalli!

Issue #46, opened 2023-04-25
Add typing check in Github actions


RobbeSneyders:
The Fondant code is typed, but the typing is not checked. This is an issue if users want to leverage the Fondant typing, as there is no guarantee that it is correct. We should add a typing check to our Github actions to validate this. This will require fixing the types, and might have to be done in multiple steps.

PR #50, opened 2023-04-26
Add typing check

PhilippeMoussalli:
This PR adds type checking into the CICD pipelines of Fondant. We're leveraging the latest version of mypy. 
#46


fondant/component_spec.py
```diff
@@ -239,7 +244,7 @@ def from_fondant_component_spec(
     @staticmethod
     def _dump_args(args: t.List[Argument]) -> t.List[t.Union[str, t.Dict[str, str]]]:
         """Dump Fondant specification arguments to kfp command arguments."""
-        dumped_args = []
+        dumped_args: t.List[t.Union[str, t.Dict[str, str]]] = []
```
NielsRogge, line 247: Tbh I find these typing hints incredibly hard to read and I'm not sure they're helpful for us. They seem useful for computers but not for humans
 
`t.List[t.Union[str, t.Dict[str, str]]]` => I mean just look at this

Just my personal opinion :D

PhilippeMoussalli: I twas getting errors from `mypy` as follows 
 Incompatible return value type (got "List[str]", expected "List[Union[str, Dict[str, str]]]") which went away when I explicitly defined the type.  
However, the args of kubeflow does required to have a list that can contains either a string or a dictionary so in that sense I think the type hint in the function's return is valid.

I'll just add a mypy exception for that one

RobbeSneyders (approved):
Thanks @PhilippeMoussalli!

I would revert ignoring the args type, but otherwise LGTM.

fondant/component_spec.py
PhilippeMoussalli: alright I think that's a better approach indeed :) thanks for the suggestion. Added

PR #53, opened 2023-04-26
Make input and output subsets in component spec optional

### LINKED ISSUE #52: Make input_subsets and output_subsets no longer required in component spec schema
Not every component needs input and output subsets. A data loading component can start without inputs, while some transform components like filtering might not write new output subsets. We should make these sections optional in the component spec.

RobbeSneyders:
Fixes #52

### DIFF
--- a/fondant/component_spec.py
+++ b/fondant/component_spec.py
@@ -141,7 +141,7 @@ class FondantComponentSpec:
         return types.MappingProxyType(
             {
                 name: ComponentSubset(subset)
-                for name, subset in self._specification["input_subsets"].items()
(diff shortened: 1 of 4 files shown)


NielsRogge (approved):
Thanks for adding!

PhilippeMoussalli (approved):
Thanks for adding this! I think we can now rely on one component spec for both Loader and and transform component

NielsRogge:
@RobbeSneyders we may also want to enforce that a loading component always needs to define `output_subsets`, and a transform component always needs to define `input_subsets`

PR #54, opened 2023-04-27
Load subset into dataframe

GeorgesLorre:
This PR implements a very basic way of merging all subsets into 1 dataframe to be used in the component.

Still to do:

- [x]  add tests
- [x] add more tests

Out of scope of this PR:
- [x] De we add the subset name to the column when loading or when writing (when creating the subset)?
- [x] No indices are used when merging, it is a hash merge right now
- [x] Writing out the resulting dataframe into subsets


GeorgesLorre:
This was branched from https://github.com/ml6team/fondant/pull/48 and merges back into it, not sure why the file diff looks at main

fondant/dataset.py
```diff
@@ -74,26 +64,19 @@ def _load_index(self):
...
         for name, subset in spec.input_subsets.items():
             fields = list(subset.fields.keys())
             subset_df = self._load_subset(name, fields)
-            subset_dfs.append(subset_df)
-
-        # return a single dataframe with column_names called subset_field
-        # TODO perhaps leverage dd.merge here instead
-        df = dd.concat(subset_dfs)
+            # left joins -> filter on index
+            df = dd.merge(df, subset_df, how="left")
```
PhilippeMoussalli, line 79: does it automatically know which columns names to join on?

PhilippeMoussalli: might be more efficient if we use the `_set_index()` method on both datasets before joining: 
https://docs.dask.org/en/stable/dataframe-best-practices.html

GeorgesLorre: The `set_index()` is persisted when writing and then reading to/from parquet. So Ideally this is already done upstream when the subset is created.

GeorgesLorre: > does it automatically know which columns names to join on?

It joins on columns with the same name, an optimisation is definitely to join on the index.

RobbeSneyders: Yeah we need to make sure that all subsets are always aligned. Since we might not be reading all subsets here, we should do this when originally writing the subsets, all the way back to the loading component, which should set the index for the different subsets it generates.

RobbeSneyders: Is there a reason to not define here that it should join on the index?

RobbeSneyders:
Nice, the partitions persisting when writing to and reading from parquet will be really beneficial to us.

tests/test_dataset.py
```diff
@@ -0,0 +1,47 @@
...
+from pathlib import Path
+
+from fondant.manifest import Manifest
+from fondant.dataset import FondantDataset
+
+manifest_path = Path(__file__).parent / "example_data/manifest.json"
+dataset_path = Path(__file__).parent / "example_data/151"
+
+@pytest.fixture
+def subset_factory(tmp_path_factory):
```
RobbeSneyders, line 13: Is there a reason to prepare the data dynamically instead of storing it in the expected state?

GeorgesLorre: I removed the dynamic part

pyproject.toml
```diff
@@ -41,7 +41,7 @@ classifiers = [
 [tool.poetry.dependencies]
 python = "^3.8"
 jsonschema = "^4.17.3"
-dask = "^2022.2.0"
+dask = "^2023.4.0"
```
RobbeSneyders, line 44: Why don't we need the `dataframe` extra here?

fondant/dataset.py
```diff
@@ -49,19 +47,12 @@ def _load_subset(
...
-            sources = index_df["source"].compute()
-            df = df[df["id"].isin(ids) & df["source"].isin(sources)]
-
         # add subset prefix to columns
-        df = df.rename(
-            columns={
-                col: name + "_" + col for col in df.columns if col not in index_fields
-            }
-        )
+        # df = df.rename(
```
RobbeSneyders, line 51: Can we reenable this?

RobbeSneyders (commented):
Thanks @GeorgesLorre, left some final comments.

tests/test_dataset.py
```diff
@@ -0,0 +1,33 @@
...
+def test_load_index(manifest):
+    fds = FondantDataset(manifest)
+    assert len(fds._load_index()) == 151
+
+
+def test_merge_subsets(manifest, component_spec):
+    fds = FondantDataset(manifest=manifest)
+    df = fds.load_dataframe(spec=component_spec)
+    assert len(df) == 151
+    assert list(df.columns) == ["id", "source", "Name", "HP", "Type 1", "Type 2"]
```
NielsRogge, line 33: The column names should be subset underscore field name

tests/example_data/raw/split.py
```diff
@@ -0,0 +1,39 @@
+"""
+This is a small script to split the raw data into different subsets to be used while testing.
+
+The data is the 151 first pokemon and the following field are available:
```
NielsRogge, line 4: Not a fan of adding Parquet files to the Github repository, and as they stay in the commit history they stay there forever so just removing them later won't make it faster.

I'd host the Parquet files on the hub and load them using the `hf_hub_download` [method](https://huggingface.co/docs/huggingface_hub/guides/download)

pyproject.toml
```diff
@@ -54,6 +54,8 @@ pipelines = ["kfp", "kubernetes"]
 liccheck = "^0.7.3"
 pre-commit = "^3.1.1"
 pytest = "^7.2.2"
+dask = {extras = ["dataframe"], version = "^2023.4.1"}
+pyarrow = "^11.0.0"
```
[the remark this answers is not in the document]
GeorgesLorre: https://docs.dask.org/en/stable/dataframe-parquet.html#engine

Used by dask but not installed by dask.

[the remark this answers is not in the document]
RobbeSneyders: Thanks, I would have expected them to at least provide an extra for it.

tests/example_data/raw/split.py
RobbeSneyders: As long as we stay in the range of kilobytes, I don't see an issue with this. Adding data files for the tests is fine as long as the goal is to validate format handling and they contain minimal data.

Moving the files out of the repo decouples them from the versioning and adds complexity.

RobbeSneyders: Moving the files out of the repo also makes our tests dependent on an internet connection, which will only bring pain.

Some best practices for tests:
https://web.archive.org/web/20210510024513/http://beyondcoding.net/articles/firstprinciples.html

NielsRogge (approved):
Thanks for working on this!

PR #56, opened 2023-04-28
Use evolve in Component class

### LINKED ISSUE #45: Use manifest evolution in `Dataset`

NielsRogge:
This PR uses the `evolve` functionality (which updates the manifest based on a component spec) in the `Component` class.

Fixes #45


fondant/test.py
RobbeSneyders, line 1: I guess this file shouldn't be included?

fondant/component.py
```diff
@@ -92,15 +92,19 @@ class FondantLoadComponent(FondantComponent):
...
         #  run_id based on args.output_manifest_path
+        metadata = json.loads(self.args.metadata)
         manifest = Manifest.create(
             base_path=metadata["base_path"],
             run_id=metadata["run_id"],
             component_id=metadata["component_id"],
         )
 
+        # evolve manifest based on component spec
+        manifest = manifest.evolve(self.spec)
```
RobbeSneyders, line 106: I don't think we need to evolve the manifest here already, as we need the input manifest first to load the data.

fondant/component.py
```diff
@@ -153,6 +158,10 @@ def _process_dataset(self, dataset: FondantDataset) -> FondantDataset:
         """
         df = dataset.load_dataframe(self.spec)
         df = self.transform(args=self.args, dataframe=df)
+        # evolve manifest
+        dataset.manifest = dataset.manifest.evolve(self.spec)
```
RobbeSneyders, line 162: I would rather remove all manifest manipulation in the `Dataset` completely, use the manifest evolution instead, and validate the output data against the evolved manifest. So regard the manifest as immutable in the `Dataset` class.

It's also not completely clear yet how we want to leverage the `Dataset` class. Currently we mutate it, but it might be interesting to have it immutable and just use it to translate between the user dataframe and the stored data.
stored data -> `Dataset(manifest)` -> dataframe
dataframe -> `Dataset(manifest)` -> stored data

RobbeSneyders (commented):
Thanks @NielsRogge.

I would propose a different approach for the `Dataset` class, see my comments.

fondant/component.py
NielsRogge: This is a loading component, so there's no data to be loaded here. For a transform component we can indeed only evolve after we have provided the dataframe to the user (see below)

fondant/component.py
NielsRogge: So that means we would have to create 2 Dataset instances for transform components?
- one to wrap the input manifest and load the data from it to provide to the user
- one to wrap the output manifest and upload data to the cloud

fondant/component.py
RobbeSneyders: Ah ok, makes sense.

I think it still might be more logical to always return the input manifest from this method though, and evolve it as a next step. Just so it's more consistent across the components.

fondant/component.py
RobbeSneyders: Yes. I just think the mutable part is done by:
- The user manipulating the dataframe
- The manifest evolution

So I don't think we need it in the dataset as well. If we follow the proposed approach, the `Dataset` class has a clear goal, which is translating between stored data and the user dataframe based on the manifest. This can include things such as data and type validation as well.

RobbeSneyders: CC: @PhilippeMoussalli @GeorgesLorre since this is about higher level design.

fondant/manifest.py
```diff
@@ -188,12 +188,18 @@ def remove_subset(self, name: str) -> None:
...
+        self, metadata: dict, component_spec: FondantComponentSpec
+    ) -> "Manifest":
+        """Evolve the manifest based on the provided metadata and component spec. The resulting
+        manifest is the expected result if the current manifest is provided to the component defined
+        by the component spec."""
         evolved_manifest = self.copy()
 
+        # Update metadata
+        for key, value in metadata.items():
+            evolved_manifest.add_metadata(key, value)
```
NielsRogge, line 201: @RobbeSneyders I've added updating the `metadata` to the evolve method.

RobbeSneyders: Do we need to pass in the metadata? We should be able to get the `component_id` from the component spec, and I guess we can get the `run_id` from the input manifest.

I'ts just not clear to me yet how we can define the version of the component as well, since this depends on the commit hash of the component, not necessarily of the repo running the pipeline.

RobbeSneyders: Even without discussing if we need to pass in metadata to the component, we don't need to pass it to the `evolve` method here if we can infer it, right? The goal is that this method can be called from outside a component as well for eg. static analysis.

GeorgesLorre:
I see some potential conflict with the join and merge subset PR's, how would you advice we proceed ?

fondant/manifest.py
NielsRogge: Yes I've [removed](https://github.com/ml6team/fondant/pull/56/commits/cd0e23a7b4c5ff02a9436a12778318061d703959) `metadata` from the `evolve` method, it now only updates the component_id based on the name property of the component spec.

fondant/manifest.py
```diff
@@ -188,12 +188,22 @@ def remove_subset(self, name: str) -> None:
...
+    ) -> "Manifest":
+        """Evolve the manifest based on the component spec. The resulting
+        manifest is the expected result if the current manifest is provided to the component defined
+        by the component spec."""
         evolved_manifest = self.copy()
 
+        # Update `component_id` of the metadata
+        evolved_manifest.add_metadata(key="component_id", value=component_id)
+
+        # Update index location as this is currently always rewritten
```
NielsRogge, line 202: @RobbeSneyders I was wondering whether the decoupling between the `evolve` method and the `Dataset` class is convenient, given that the `evolve` method defines new locations of files without actually uploading any data. e.g. here I need to update the location of the index, since the `Dataset` class will be used to write a new index.

Wouldn't it make more sense to just have a single place where both a location is defined + data is written to the cloud (i.e. both in the `Dataset` class)?

RobbeSneyders: > Otherwise we will need a strict coupling between the Manifest and Dataset classes to make sure that the latter will upload new data to new locations defined by the evolve method.

That's indeed what I'm proposing. There's a couple of advantages of splitting the `evolve` functionality out of the `Dataset` class:
- Separation of concerns, which makes it easier to follow the logic.
- The `evolve` method works purely based on a `Manifest` and a `ComponentSpec` which allows you to precalculate the manifests for a complete pipeline statically.

I see no reason we cannot include the location update in the `evolve` method.

fondant/component.py
NielsRogge: > The Dataset class should be able to check the output subsets of its manifest, which are the only ones it needs to upload.

A manifest only consists of subsets, it's the component spec that lists the output subsets, so I guess we need that to check which ones to upload. So then we're checking output subsets twice? One time in the `evolve` method and one time in the `Dataset` class

RobbeSneyders: Sorry, you're right. The output subsets should be checked in the component spec, which the `Dataset` also needs to define which data to load.

So a `Dataset` can either
- load data based on an input manifest and a component specification
- write data based on an output manifest and a component specification

To be seen if it's logical to keep both in a single class or if we should split them.

NielsRogge:
@GeorgesLorre this PR can be merged after the join and merge PRs

fondant/component.py
```diff
@@ -74,27 +74,37 @@ def _load_or_create_manifest(self) -> Manifest:
...
-        manifest = self._load_or_create_manifest()
-        dataset = FondantDataset(manifest)
-        dataset = self._process_dataset(dataset)
+        input_manifest = self._load_or_create_manifest()
+        dataset = FondantDataset(input_manifest)
+
+        df = self._process_dataset(dataset)
+
+        output_manifest = input_manifest.evolve(component_spec=self.spec)
+        dataset = FondantDataset(output_manifest)
```
GeorgesLorre, line 91: Nit but the can we choose better names for the datasets here ? since we create 2 ?

fondant/test.py
```diff
@@ -0,0 +1,26 @@
+import yaml
```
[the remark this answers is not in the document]
GeorgesLorre: still needs to be removed

fondant/manifest.py
```diff
@@ -189,11 +189,20 @@ def remove_subset(self, name: str) -> None:
...
-        """Evolve the manifest based on the provided component spec. The resulting manifest is
-        the expected result if the current manifest is provided to the component defined by the
-        component spec."""
+        """Evolve the manifest based on the component spec. The resulting
+        manifest is the expected result if the current manifest is provided to the component defined
+        by the component spec."""
         evolved_manifest = self.copy()
 
+        # Update `component_id` of the metadata
+        component_id = component_spec.name.lower().replace(" ", "_")
```
PhilippeMoussalli, line 198: can we take this from the `image` field directly?

fondant/test.py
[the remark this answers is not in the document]
RobbeSneyders: Still to be removed :)

fondant/manifest.py
```diff
@@ -189,11 +189,20 @@ def remove_subset(self, name: str) -> None:
         del self._specification["subsets"][name]
 
     def evolve(self, component_spec: FondantComponentSpec) -> "Manifest":
-        """Evolve the manifest based on the provided component spec. The resulting manifest is
-        the expected result if the current manifest is provided to the component defined by the
-        component spec."""
+        """Evolve the manifest based on the component spec. The resulting
```
RobbeSneyders, line 192: Can be reformatted.

fondant/component.py
```diff
@@ -74,27 +74,37 @@ def _load_or_create_manifest(self) -> Manifest:
...
         with the new or loaded subsets"""
 
     def run(self):
         """
         Runs the component.
         """
-        manifest = self._load_or_create_manifest()
-        dataset = FondantDataset(manifest)
-        dataset = self._process_dataset(dataset)
+        input_manifest = self._load_or_create_manifest()
```
PhilippeMoussalli, line 85: Nice :) I think this is exactly in line with what we discussed this morning

fondant/component.py
```diff
@@ -107,22 +117,19 @@ def _load_or_create_manifest(self) -> Manifest:
     def load(self, args: argparse.Namespace) -> dd.DataFrame:
         """Abstract method that loads the initial dataframe"""
 
-    def _process_dataset(self, dataset: FondantDataset) -> FondantDataset:
+    def _process_dataset(self, dataset: FondantDataset) -> dd.DataFrame:
         """This function takes in a FondantDataset object and processes the initial input dataframe
```
PhilippeMoussalli, line 121: I think some of the docstrings need to be updated here. Currently this is just used to load in the initial dataframe. Adding the index and subsets is currently handled by the `run` method

tests/test_manifest_evolution.py
```diff
@@ -24,10 +24,13 @@ def examples():
...
-@pytest.mark.parametrize("component_spec,output_manifest", list(examples()))
+@pytest.mark.parametrize("component_spec, output_manifest", list(examples()))
 def test_evolution(input_manifest, component_spec, output_manifest):
     manifest = Manifest(input_manifest)
-    component = FondantComponentSpec(component_spec)
-    evolved_manifest = manifest.evolve(component)
+    component_spec = FondantComponentSpec(component_spec)
+    evolved_manifest = manifest.evolve(component_id="100", component_spec=component_spec)
+
+    print("Spec:", evolved_manifest._specification)
```
PhilippeMoussalli, line 33: I think we can remove those print statements

examples/pipelines/simple_pipeline/config/components_config.py
```diff
@@ -23,8 +23,8 @@ class ImageFilterConfig:
         MIN_WIDTH (int): Minimum width for each image.
     """
 
-    MIN_HEIGHT = 200
-    MIN_WIDTH = 200
+    MIN_HEIGHT = 600
```
PhilippeMoussalli, line 26: what's the reason for bumping this up? we might be filtering everything in this case.

fondant/component.py
```diff
@@ -74,27 +74,37 @@ def _load_or_create_manifest(self) -> Manifest:
...
+        df = self._process_dataset(dataset)
+
+        output_manifest = input_manifest.evolve(component_spec=self.spec)
+        dataset = FondantDataset(output_manifest)
+
+        # write index and output subsets to remote storage
+        dataset.write_index(df)
+        dataset.write_subsets(df, self.spec)
+
         dataset.upload(save_path=self.args.output_manifest_path)
```
RobbeSneyders, line 97: I would lift saving the manifest from the `Dataset` to the `Component` class.

RobbeSneyders (commented):
Thanks @NielsRogge! Some minor comments left.

fondant/component.py
```diff
@@ -74,27 +74,37 @@ def _load_or_create_manifest(self) -> Manifest:
         """Abstract method that returns the dataset manifest"""
 
     @abstractmethod
-    def _process_dataset(self, dataset: FondantDataset) -> FondantDataset:
+    def _process_dataset(self, dataset: FondantDataset) -> dd.DataFrame:
```
PhilippeMoussalli, line 77: docstring of this function need to be updated and moved to the `run` method

fondant/component.py
```diff
@@ -74,27 +74,37 @@ def _load_or_create_manifest(self) -> Manifest:
...
         """Abstract method that processes the input dataframe and updates the Fondant Dataset
         with the new or loaded subsets"""
 
     def run(self):
```
PhilippeMoussalli, line 81: Would be nice to have a long descriptive docstring describing in-details what the `run()` method is responsible for. Alternatively, you can include this doc at the class definition level (`FondantComponent(ABC)`) to provide a more comprehensive explanation of the component's functionality.

fondant/component.py
NielsRogge: This is fixed now

PhilippeMoussalli (commented):
Thanks for taking this up Niels! Left a few comments mainly related to docstrings but the overall logic of handling the manifest evolve and component spec in the Component seems to be valid.

fondant/manifest.py
NielsRogge: the `image` field currently isn't used, and should point to a URL like `ghcr.io/ml6team/image_filtering:latest`, right?

In that case, we could split the string to get the component ID from that. For now I just used the name property.

examples/pipelines/simple_pipeline/config/components_config.py
NielsRogge: The image filtering component wasn't filtering out any images, so I just set this higher to make sure the resulting dataframe contains less images than the original one, to see whether the index is properly updated

NielsRogge: In this, out of the 833 images, we are left with 551 images.

fondant/manifest.py
PhilippeMoussalli: We use it to build the `kubeflow_component.yaml` [example](https://github.com/ml6team/fondant/blob/main/examples/pipelines/simple_pipeline/components/load_from_hub/kubeflow_component.yaml#:~:text=container%3A-,image,-%3A%20ghcr.io) the FondantSpec so it should be properly defined for each component and pointing to a valid url. 
Splitting it should be easy though so would rather do it now or add a TODO

fondant/component.py
```diff
@@ -92,15 +92,19 @@ class FondantLoadComponent(FondantComponent):
...
         #  run_id based on args.output_manifest_path
+        metadata = json.loads(self.args.metadata)
         manifest = Manifest.create(
             base_path=metadata["base_path"],
             run_id=metadata["run_id"],
             component_id=metadata["component_id"],
         )
 
+        # evolve manifest based on component spec
+        manifest = manifest.evolve(metadata=metadata, component_spec=self.spec)
```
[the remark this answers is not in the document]
NielsRogge: This is outdated, is updated now

examples/pipelines/simple_pipeline/config/components_config.py
PhilippeMoussalli: alright now it's more clear. Thanks!

examples/pipelines/simple_pipeline/components/image_filtering/src/main.py
```diff
@@ -29,7 +29,7 @@ def transform(self, df: dd.DataFrame, args: Dict) -> dd.DataFrame:
         """
         logger.info("Filtering dataset...")
         min_width, min_height = args.min_width, args.min_height
-        filtered_df = df[(df["images_width"] > min_width) & (df["images_height"] > min_height)]
+        filtered_df = dataframe[(dataframe["images_width"] > min_width) & (dataframe["images_height"] > min_height)]
```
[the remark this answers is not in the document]
NielsRogge: The superclass defined the dataframe argument as `dataframe` instead of `df`.

fondant/manifest.py
NielsRogge: Updating the index location is now part of the `evolve` method

fondant/component.py
NielsRogge: Closing this discussion as it's now incorporated

fondant/component.py
NielsRogge: Done!

fondant/manifest.py
NielsRogge: What benefit would using the `image` provide over using the `name`?

NielsRogge:
Feel free to approve the PR, I've addressed all comments.

GeorgesLorre (approved):
Nice work

fondant/manifest.py
PhilippeMoussalli: you could include the docker tag of the images that you've used to run a certain a component in the manifest which can be used for versioning. (Figuring out what component version generated which manifest)

RobbeSneyders (approved):

PR #58, opened 2023-04-28
Split dataframe to subsets

PhilippeMoussalli:
This PR optimizes the process of executing dask transformations and writing by creating delayed tasks that are executed in parallel. This is mainly to optimize the process of writing many different subsets at the same time.

I also added more documentation for the `Dataset` class. The changes did require some refactoring


fondant/dataset.py
```diff
@@ -95,71 +114,145 @@ def load_dataframe(self, spec: FondantComponentSpec) -> dd.DataFrame:
...
+            subset_columns = verify_subset_columns(subset_name, subset.fields, df)
+
+            # Create a new dataframe with only the columns needed for the output subset
+            subset_df = create_subset_dataframe(subset_name, subset_columns, df)
+
+            # Add the output subset to the manifest
+            manifest_fields = [
+                (field.name, Type[field.type.name]) for field in subset.fields.values()
+            ]
+            self.manifest.add_subset(subset_name, fields=manifest_fields)
```
NielsRogge, line 231: The `Dataset` class should be a wrapper around an immutable `Manifest` (see discussion [here](https://github.com/ml6team/fondant/pull/56#discussion_r1180159365)).

The `evolve` method currently takes care of updating the manifest (adding subsets based on the output subsets of the component spec). cc @RobbeSneyders

fondant/dataset.py
```diff
@@ -95,71 +114,145 @@ def load_dataframe(self, spec: FondantComponentSpec) -> dd.DataFrame:
...
-        manifest_fields = [
-            (field.name, Type[field.type.name]) for field in fields.values()
-        ]
-        self.manifest.add_subset(name, fields=manifest_fields)
+        return upload_index_task
 
-        # create expected schema
-        expected_schema = {field.name: field.type.name for field in fields.values()}
-        expected_schema.update(self.index_schema)
+    def get_upload_subsets_task(
```
NielsRogge, line 159:
```suggestion
    def get_upload_subsets_tasks(
```
Seems like this method returns a list of tasks

fondant/dataset.py
```diff
@@ -31,56 +31,75 @@ def __init__(self, manifest: Manifest):
...
     ) -> dd.DataFrame:
-        # get subset from the manifest
-        subset = self.manifest.subsets[name]
-        # get remote path
+        """
+        Function that loads the subset
+        Args:
+            subset_name: the name of the subset to load
+            fields: the fields to load from the subset
+            index: the manifest index to filter the subset on.
```
NielsRogge, line 41:
```suggestion
            index: the index to filter the subset on. If not provided, the default manifest index is used.
```

fondant/dataset.py
```diff
@@ -31,56 +31,75 @@ def __init__(self, manifest: Manifest):
...
         fields = index_fields + fields
 
-        logger.info(f"Loading subset {name} with fields {fields}...")
+        logger.info(f"Loading subset {subset_name} with fields {fields}...")
 
-        df = dd.read_parquet(
-            remote_path,
-            columns=fields,
-        )
+        subset_df = dd.read_parquet(remote_path, columns=fields)
```
NielsRogge, line 53: Is this also a Dask task?

PhilippeMoussalli: It returns a dask dataframe object (lazily loaded). The `to_parquet` method defines a task to write to a parquet file so both are separate.

fondant/component.py
```diff
@@ -118,9 +118,16 @@ def _process_dataset(self, dataset: FondantDataset) -> FondantDataset:
         """
         # Load the dataframe according to the custom function provided to the user
         df = self.load(self.args)
+
         # Add index and specified subsets and write them to remote storage
-        dataset.add_index(df)
-        dataset.add_subsets(df, self.spec)
+        index_task = dataset.get_upload_index_task(df)
+        subset_tasks = dataset.get_upload_subsets_task(df, self.spec)
```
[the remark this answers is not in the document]
PhilippeMoussalli: Makes sense, I'v updated it accordingly.Only small downside is that now we're handling both writing of index and all of the subsets separately but that shouldn't introduce a major performance bump

GeorgesLorre:
For testing you can use https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html to write out the subsets temporarily.

RobbeSneyders:
#56 has been merged. Can you rebase or merge so the conflicts are resolved? That will make it easier to review.

fondant/dataset.py
```diff
@@ -27,43 +26,73 @@ def __init__(self, manifest: Manifest):
...
-        # get remote path
+    def _load_subset(
+        self, subset_name: str, fields: t.List[str], index: t.Optional[Index] = None
+    ) -> dd.DataFrame:
+        """
+        Function that loads the subset
+        Args:
+            subset_name: the name of the subset to load
+            fields: the fields to load from the subset
+            index: optional index to filter the subset on. If not provided, the default manifest
```
RobbeSneyders, line 37: Shouldn't this always be the manifest index? I don't see this parameter currently being used.

fondant/dataset.py
```diff
@@ -76,60 +105,124 @@ def load_dataframe(self, spec: FondantComponentSpec) -> dd.DataFrame:
...
+            df: The DataFrame to be uploaded.
+            remote_path: the location to upload the subset to
+            schema: the schema of the dataframe to write
+        Returns:
+             A delayed Dask task that uploads the DataFrame to the remote storage location when
+              executed.
+        """
+
+        # Define task to upload index to remote storage
+        upload_index_task = dd.to_parquet(
```
RobbeSneyders, line 125: This naming should not be index specific.

fondant/dataset.py
```diff
@@ -76,60 +105,124 @@ def load_dataframe(self, spec: FondantComponentSpec) -> dd.DataFrame:
...
-            subset_columns = [f"{name}_{field}" for field in fields]
+        """
+        Write subset tasks to a remote location
+
+        Args:
+            df (dask.dataframe.DataFrame): The input Dask dataframe.
+            spec (FondantComponentSpec): The specification of the output subsets.
+
+        Raises:
+            ValueError: If a field defined in an output subset is not present in the input
```
RobbeSneyders, line 158: "input dataframe" is a bit confusing here, as it could refer to the dataframe we provide as the input of the component as well.

fondant/dataset.py
```diff
@@ -76,60 +105,124 @@ def load_dataframe(self, spec: FondantComponentSpec) -> dd.DataFrame:
...
+             dataframe.
+        """
+
+        def verify_subset_columns(subset_name, subset_fields, df):
+            """
+            Verify that all the fields defined in the output subset are present in
+            the input dataframe
+            """
+
+            field_names = list(subset_fields.keys())
```
RobbeSneyders, line 168: Either the `list()` or `.keys()` call can be removed here.

Or you can even combine it with the list comprehension below:
```python
subset_columns = [f"{subset_name}_{field}" for field in subset_fields]
```
which is quite readable I think.

fondant/dataset.py
```diff
@@ -76,60 +105,124 @@ def load_dataframe(self, spec: FondantComponentSpec) -> dd.DataFrame:
...
+        Args:
+            df (dask.dataframe.DataFrame): The input Dask dataframe.
+            spec (FondantComponentSpec): The specification of the output subsets.
+
+        Raises:
+            ValueError: If a field defined in an output subset is not present in the input
+             dataframe.
+        """
+
+        def verify_subset_columns(subset_name, subset_fields, df):
```
RobbeSneyders, line 162: Can you add types here?

fondant/dataset.py
```diff
@@ -27,43 +26,73 @@ def __init__(self, manifest: Manifest):
...
+    def _load_subset(
+        self, subset_name: str, fields: t.List[str], index: t.Optional[Index] = None
+    ) -> dd.DataFrame:
+        """
+        Function that loads the subset
+        Args:
+            subset_name: the name of the subset to load
+            fields: the fields to load from the subset
+            index: optional index to filter the subset on. If not provided, the default manifest
+             index is used.
```
NielsRogge, line 38: I don't think this argument is implemented correctly at the moment (see code below). We can leave it out for now

fondant/dataset.py
```diff
@@ -76,60 +105,124 @@ def load_dataframe(self, spec: FondantComponentSpec) -> dd.DataFrame:
...
+
+        def verify_subset_columns(subset_name, subset_fields, df):
+            """
+            Verify that all the fields defined in the output subset are present in
+            the input dataframe
+            """
+
+            field_names = list(subset_fields.keys())
+            subset_columns = [f"{subset_name}_{field}" for field in field_names]
             subset_columns.extend(self.mandatory_subset_columns)
```
RobbeSneyders, line 170: Can we rename `mandatory_subset_columns` to `index_fields` or something? Then it's more clear what is happening here just from the code.

fondant/dataset.py
```diff
@@ -76,60 +105,124 @@ def load_dataframe(self, spec: FondantComponentSpec) -> dd.DataFrame:
...
+
+            return subset_df
+
+        upload_subsets_tasks = []
+
+        # Loop through each output subset defined in the spec
+        for subset_name, subset in spec.output_subsets.items():
+            # Verify that all the fields defined in the output subset are present in the
+            # input dataframe
+            subset_columns = verify_subset_columns(subset_name, subset.fields, df)
```
RobbeSneyders, line 207: I think we'll need some logic here for `additionalFields` as well, but it's fine to pick this up in a separate PR. I added a ticket to the project for this.

fondant/dataset.py
```diff
@@ -27,43 +26,73 @@ def __init__(self, manifest: Manifest):
...
 
-    def _load_subset(self, name: str, fields: t.List[str]) -> dd.DataFrame:
-        # get subset from the manifest
-        subset = self.manifest.subsets[name]
-        # get remote path
+    def _load_subset(
+        self, subset_name: str, fields: t.List[str], index: t.Optional[Index] = None
+    ) -> dd.DataFrame:
+        """
+        Function that loads the subset
```
NielsRogge, line 33:
```suggestion
        Function that loads a subset as a Dask dataframe.
```

fondant/dataset.py
```diff
@@ -76,60 +105,124 @@ def load_dataframe(self, spec: FondantComponentSpec) -> dd.DataFrame:
...
                     raise ValueError(
-                        f"Field {col} defined in output subset {name} but not found in dataframe"
+                        f"Field {col} defined in output subset {subset_name} "
+                        f"but not found in dataframe"
                     )
 
-            # load subset dataframe
+            return subset_columns
+
+        def create_subset_dataframe(subset_name, subset_columns, df):
```
RobbeSneyders, line 181: Could you add types here?

fondant/dataset.py
```diff
@@ -27,43 +26,73 @@ def __init__(self, manifest: Manifest):
...
+        subset_df = dd.read_parquet(remote_path, columns=fields)
+
+        # filter on default index of manifest if no index is provided
+        if index is None:
+            index_df = self._load_index()
+            ids = index_df["id"].compute()
+            sources = index_df["source"].compute()
+            subset_df = subset_df[
+                subset_df["id"].isin(ids) & subset_df["source"].isin(sources)
+            ]
```
NielsRogge, line 59: Can be removed for now

fondant/dataset.py
```diff
@@ -76,60 +105,124 @@ def load_dataframe(self, spec: FondantComponentSpec) -> dd.DataFrame:
...
-            self._upload_subset(name, subset.fields, subset_df)
+
+            return subset_df
+
+        upload_subsets_tasks = []
+
+        # Loop through each output subset defined in the spec
+        for subset_name, subset in spec.output_subsets.items():
+            # Verify that all the fields defined in the output subset are present in the
+            # input dataframe
```
RobbeSneyders, line 206: "input dataframe" is again a bit confusing here. I think "output dataframe
 is more clear.

fondant/dataset.py
```diff
@@ -27,43 +26,73 @@ def __init__(self, manifest: Manifest):
...
             }
         )
 
-        return df
+        return subset_df
 
-    def _load_index(self):
+    def _load_index(self) -> dd.DataFrame:
+        """
+        Function that loads the index dataframe from the manifest
```
NielsRogge, line 74:
```suggestion
        Function that loads the index from the manifest as a Dask dataframe.
```

fondant/dataset.py
```diff
@@ -27,43 +26,73 @@ def __init__(self, manifest: Manifest):
...
         # load index from parquet, expecting id and source columns
-        df = dd.read_parquet(remote_path, columns=["id", "source"])
+        index_df = dd.read_parquet(remote_path, columns=["id", "source"])
 
-        return df
+        return index_df
 
     def load_dataframe(self, spec: FondantComponentSpec) -> dd.DataFrame:
+        """
+        Function that loads the subsets defined in the component spec as a single dask dataframe
```
NielsRogge, line 90:
```suggestion
        Function that loads the subsets defined in the component spec as a single dask dataframe
        
```

fondant/dataset.py
```diff
@@ -76,60 +105,124 @@ def load_dataframe(self, spec: FondantComponentSpec) -> dd.DataFrame:
...
+            remote_path: the location to upload the subset to
+            schema: the schema of the dataframe to write
+        Returns:
+             A delayed Dask task that uploads the DataFrame to the remote storage location when
+              executed.
+        """
+
+        # Define task to upload index to remote storage
+        upload_index_task = dd.to_parquet(
+            df, remote_path, schema=schema, overwrite=True, compute=False
```
RobbeSneyders, line 126: Should we set `overwrite=True`? I think we should assume that the directory we are writing to is empty and otherwise fail?

fondant/dataset.py
```diff
@@ -76,60 +105,124 @@ def load_dataframe(self, spec: FondantComponentSpec) -> dd.DataFrame:
...
+
+            # Create a new dataframe with only the columns needed for the output subset
+            subset_df = create_subset_dataframe(subset_name, subset_columns, df)
+
+            # Get the remote path where the output subset should be uploaded
+            remote_path = self.manifest.subsets[subset_name].location
+
+            # Create the expected schema for the output subset
+            expected_schema = {
+                field.name: field.type.name for field in subset.fields.values()
```
RobbeSneyders, line 217: Don't we need to convert these types to `PyArrow` types?

RobbeSneyders (commented):
Thanks @PhilippeMoussalli! General approach looks good to me. Left some comments :)

fondant/dataset.py
```diff
@@ -27,43 +26,73 @@ def __init__(self, manifest: Manifest):
...
 
-        return df
+        return index_df
 
     def load_dataframe(self, spec: FondantComponentSpec) -> dd.DataFrame:
+        """
+        Function that loads the subsets defined in the component spec as a single dask dataframe
+        Args:
+            spec: the fondant component spec
+        Returns:
```
NielsRogge, line 93:
```suggestion

        Returns:
```
Same here

fondant/dataset.py
```diff
@@ -27,43 +26,73 @@ def __init__(self, manifest: Manifest):
...
+            The dask dataframe with the field columns in the format (<subset>_<column_name>)
```
NielsRogge, line 94:
```suggestion
            The dask dataframe with the field columns in the format (<subset>_<column_name>) as well as the index columns.
```

fondant/dataset.py
```diff
@@ -76,60 +105,124 @@ def load_dataframe(self, spec: FondantComponentSpec) -> dd.DataFrame:
...
-            schema=self.index_schema,
-            overwrite=True,
+    @staticmethod
+    def _create_write_dataframe_task(
+        *, df: dd.DataFrame, remote_path: str, schema: t.Dict[str, str]
+    ) -> dd.core.Scalar:
+        """
+        Creates a delayed Dask task to upload the given DataFrame to the remote storage location
+         specified in the manifest.
+        Args:
```
NielsRogge, line 115:
```suggestion

        Args:
```

fondant/dataset.py
PhilippeMoussalli: yeah indeed, this was easier for testing purposes but I think it can be removed

fondant/dataset.py
```diff
@@ -76,60 +105,124 @@ def load_dataframe(self, spec: FondantComponentSpec) -> dd.DataFrame:
...
+        # Write index
+        dd.compute(upload_index_task)
 
     def write_subsets(self, df: dd.DataFrame, spec: FondantComponentSpec):
-        for name, subset in spec.output_subsets.items():
-            fields = list(subset.fields.keys())
-            # verify fields are present in the output dataframe
-            subset_columns = [f"{name}_{field}" for field in fields]
+        """
+        Write subset tasks to a remote location
```
NielsRogge, line 151:
```suggestion
        Write all subsets of the Dask dataframe to a remote location.
```

fondant/dataset.py
PhilippeMoussalli: I will just call it `dataframe` maybe to avoid confusion

fondant/dataset.py
PhilippeMoussalli: not necessarily, it works fine with general parquet types (similar to how it's currently defined for the `index_schema`)

fondant/dataset.py
```diff
@@ -76,60 +105,124 @@ def load_dataframe(self, spec: FondantComponentSpec) -> dd.DataFrame:
...
+
+        Args:
+            df (dask.dataframe.DataFrame): The input Dask dataframe.
```
NielsRogge, line 154:
```suggestion
            df (dask.dataframe.DataFrame): The Dask dataframe returned by the user.
```

fondant/dataset.py
```diff
@@ -76,60 +105,124 @@ def load_dataframe(self, spec: FondantComponentSpec) -> dd.DataFrame:
...
+            spec (FondantComponentSpec): The specification of the output subsets.
```
NielsRogge, line 155:
```suggestion
            spec (FondantComponentSpec): The component specification.
```

fondant/dataset.py
PhilippeMoussalli: or `output dataframe` as you specified below might be better

fondant/dataset.py
RobbeSneyders: Ok, not sure if the `field.type.name` completely align with `parquet` types though.

RobbeSneyders: They currently don't align completely with the `Arrow` data types either. If those are the ones we need, we can update our types to match the `Arrow` types completely. But we can do this in a separate PR. I'll add a ticket.

PhilippeMoussalli: At first glance it seems like they align no? I'm comparing both of these: https://github.com/ml6team/fondant/blob/main/fondant/schema.py
https://arrow.apache.org/docs/python/api/datatypes.html#api-types

RobbeSneyders: At least `categorical` is not aligned, and we are missing some.

tests/example_data/subsets/index/part.0.parquet
GeorgesLorre, line 1: What exactly changed in the parquet files ?

fondant/dataset.py
```diff
@@ -76,60 +99,129 @@ def load_dataframe(self, spec: FondantComponentSpec) -> dd.DataFrame:
...
+        """
+        Write the index dataframe to a remote location.
+        Args:
+            df: The output Dask dataframe returned by the user.
+        """
+        remote_path = self.manifest.index.location
         index_columns = list(self.manifest.index.fields.keys())
 
         # load index dataframe
         index_df = df[index_columns]
```
GeorgesLorre, line 136: Does this `index_df` have an index ? Maybe we need to call `.set_index()` ?

fondant/dataset.py
```diff
@@ -76,60 +99,129 @@ def load_dataframe(self, spec: FondantComponentSpec) -> dd.DataFrame:
...
+        """
+
+        def verify_subset_columns(
+            subset_name: str, subset_fields: t.Mapping[str, Field], df: dd.DataFrame
+        ) -> t.List[str]:
+            """
+            Verify that all the fields defined in the output subset are present in
+            the output dataframe
+            """
+            # TODO: add logic for `additional fields`
```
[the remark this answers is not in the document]
RobbeSneyders: Yes

fondant/dataset.py
PhilippeMoussalli: we're using both the `id` and `source` as index. It does not seem like dask supports multiindex though https://dask.discourse.group/t/everything-about-multiindex-in-dask/593/2

tests/example_data/subsets/index/part.0.parquet
PhilippeMoussalli: I think the operation you defined returns a lazy dataframe so no need to define a task there. 

I changed the `source` data type from integer to string since that's what we currently expect

fondant/dataset.py
GeorgesLorre: We could just index `id` since it is better then no index (this is also what happens in the test data see split.py). I'll create a ticket to solve multi-index.

PR #71, opened 2023-05-04
Example/alpha pipeline

jamesbraniganml6:


examples/pipelines/alpha_pipeline/components/retrieve_images/src/main.py
```diff
@@ -0,0 +1,56 @@
...
+    Returns:
+        df (Dask DataFrame): a Dask DataFrame with an additional column
+                             "image_urls", where each row contains a list of
+                             URLs of images related to the query in that row
+    """
+
+    client = ClipClient(
+        url="https://knn.laion.ai/knn-service",
+        indice_name="laion5B-L-14",
+        num_images=1000,  # TODO: include as argument and increase for
```
NielsRogge, line 44: @PhilippeMoussalli we might need to instead download the entire LAION-5B dataset in order to query more images

PhilippeMoussalli: I think we can upgrade this once you have the end-to-end pipeline. Working with the publicly hosted service should be much easier for testing

NielsRogge: Ok makes sense

examples/pipelines/alpha_pipeline/components/generate_prompts/src/main.py
```diff
@@ -0,0 +1,130 @@
...
+            Dask dataframe
+        """
+        room_tuples = list(itertools.product(rooms, interior_prefix, interior_styles))
+        prompts = list(map(lambda x: make_interior_prompt(*x), room_tuples))
+
+        df = pd.DataFrame(prompts, columns=["prompts_text"])
+
+        df = dd.from_pandas(
+            df, npartitions=1
+        )  # need to decide how npartitions should be set
```
NielsRogge, line 117: This can be in a single line

examples/pipelines/alpha_pipeline/components/add_captions/src/main.py
```diff
@@ -0,0 +1,53 @@
...
+    Returns:
+        str: The generated caption for the input image bytes, or None if
+             image_bytes is None.
+    """
+    if image_bytes is None:
+        return None
+
+    image = Image.open(io.BytesIO(image_bytes))
+    pixel_values = processor(images=image, return_tensors="pt").pixel_values
+    generated_ids = model.generate(pixel_values=pixel_values, max_length=50)
```
NielsRogge, line 27: This is not very efficient, would be much better to generate in batches of images as the generate method accepts that.

See https://examples.dask.org/machine-learning/torch-prediction.html. We may need to use Dask.delayed for that

examples/pipelines/alpha_pipeline/components/segment_images/src/main.py
```diff
@@ -0,0 +1,56 @@
...
+    Returns:
+        seg: A segmentation map containing class labels for each pixel in the
+             input image.
+    """
+    if image_bytes is None:
+        return None
+
+    image = Image.open(BytesIO(image_bytes))
+    pixel_values = processor(images=image, return_tensors="pt").pixel_values
+    outputs = model(pixel_values)
```
NielsRogge, line 29: Same comment here, this is not very efficient given that neural networks are a lot faster when sending batches through them. We will need to leverage Dask.delayed here

examples/pipelines/alpha_pipeline/components/add_captions/src/main.py
```diff
@@ -0,0 +1,53 @@
+import dask.dataframe as dd
+from transformers import AutoProcessor, AutoModelForCausalLM
+from PIL import Image
+import io
+
+
+processor = AutoProcessor.from_pretrained("microsoft/git-base-coco")
+model = AutoModelForCausalLM.from_pretrained("microsoft/git-base-coco")
```
NielsRogge, line 8: We will need to place the model on GPU else it's gonna be very slow

RobbeSneyders:
This has been split into separate PRs.

PR #73, opened 2023-05-04
Redesign transform interface

### LINKED ISSUE #67: Redesign `component.transform` interface
Currently the interface of the `component.transform` method is:

```python
def transform(
    self, args: argparse.Namespace, dataframe: dd.DataFrame
) -> dd.DataFrame:
```

Since the `dataframe` is the main argument, I would put it first. And instead of providing an `argparse.Namespace`, I would provide the arguments as keyword arguments.

```python
def transform(
    self, dataframe: dd.DataFrame, **kwargs
) -> dd.DataFrame:
```

This way users can define the keyword arguments they are expecting in the method signature instead of having to unpack an `argparse.Namespace`. Eg. for a component that filters on image dimensions:

```python
def transform(
    self, dataframe: dd.DataFrame, *, height: int, width: int
) -> dd.DataFrame:
```

NielsRogge:
Fixes #67


fondant/component.py
```diff
@@ -142,9 +153,7 @@ def _load_or_create_manifest(self) -> Manifest:
         return Manifest.from_file(self.args.input_manifest_path)
 
     @abstractmethod
-    def transform(
-        self, args: argparse.Namespace, dataframe: dd.DataFrame
-    ) -> dd.DataFrame:
+    def transform(self, dataframe: dd.DataFrame, **kwargs) -> dd.DataFrame:
         """Abstract method for applying data transformations to the input dataframe"""
```
PhilippeMoussalli, line 157: Could you add an extensive docstring since this is a user facing function. You could describe the dataframe object and that the user can include additional key-word arguments

fondant/component.py
```diff
@@ -120,7 +130,7 @@ def _load_or_create_manifest(self) -> Manifest:
         return manifest
 
     @abstractmethod
-    def load(self, args: argparse.Namespace) -> dd.DataFrame:
+    def load(self, **kwargs) -> dd.DataFrame:
```
PhilippeMoussalli, line 133: Same comments as fro the `transform` method (my bad should have included this in my PR)

examples/pipelines/simple_pipeline/components/image_filtering/src/main.py
```diff
@@ -18,19 +18,19 @@ class ImageFilterComponent(FondantTransformComponent):
     Component that filters images based on height and width.
     """
 
-    def transform(self, dataframe: dd.DataFrame, args: Dict) -> dd.DataFrame:
+    def transform(self, dataframe: dd.DataFrame, *, min_width: int, min_height: int) -> dd.DataFrame:
         """
         Args:
             df: Dask dataframe
```
PhilippeMoussalli, line 24:
```suggestion
            df: The Dask dataframe to load as defined in the component specification 
```
Also make sure to include this doc in the `transform` method of the component class (see below)

examples/pipelines/simple_pipeline/components/image_filtering/src/main.py
```diff
@@ -18,19 +18,19 @@ class ImageFilterComponent(FondantTransformComponent):
...
-            args: args to pass to the function
+            min_width: minimum width of the resulting images
+            min_height: minimum height of the resulting images
         
         Returns:
-            dataset
+            Dask dataframe
```
PhilippeMoussalli, line 29:
```suggestion
            The transformed Dask dataframe to write 
```

PhilippeMoussalli (approved):
Nice work! Left a few comments but should be good to go

RobbeSneyders:
I opened #74 to address the same issue. Please check the assignee before tackling an issue and assign it to yourself if you start to work on something.

NielsRogge:
Will close this PR in favor of #74

PR #74, opened 2023-05-04
Provide unpacked keyword arguments to `component.transform` method

### LINKED ISSUE #67: Redesign `component.transform` interface

RobbeSneyders:
Fixes #67


examples/pipelines/simple_pipeline/components/image_filtering/requirements.txt
```diff
@@ -1,3 +1,3 @@
-git+https://github.com/ml6team/fondant.git@d697be951fa33faa0baa3625cf3cb04d6800834a
+git+https://github.com/ml6team/fondant.git
```
RobbeSneyders, line 1: Since we squash merge, any commit hash part of a PR will actually not become part of `main`, so it's not ideal to use them as a reference.

I think we should always keep these examples in line with main, so I would propose to unfreeze the fondant version in their requirements.

fondant/component.py
```diff
@@ -70,6 +80,14 @@ def _add_and_parse_args(self) -> argparse.Namespace:
 
         return parser.parse_args()
 
+    @property
+    def user_arguments(self) -> t.Dict[str, t.Any]:
```
NielsRogge, line 84: Maybe you can add a docstring here to specify that this is about the custom arguments that the user defined (excluding the Kubeflow arguments)

tests/test_component.py
```diff
@@ -0,0 +1,52 @@
...
+    """Test that arguments are passed correctly to `Component.transform` method."""
+
+    class EarlyStopException(Exception):
+        """Used to stop execution early instead of mocking all later functionality"""
+
+    # Mock `Dataset.load_dataframe` so no actual data is loaded
+    def mocked_load_dataframe(self, spec):
+        return dd.from_dict({"a": [1, 2, 3]}, npartitions=1)
+
+    monkeypatch.setattr(FondantDataset, "load_dataframe", mocked_load_dataframe)
```
NielsRogge, line 24: Just a question, would it be cleaner to use the `mock.patch` decorator function?

https://docs.python.org/3/library/unittest.mock.html

tests/test_component.py
```diff
@@ -0,0 +1,52 @@
...
+
+    # Implemented Component class
+    class MyComponent(FondantTransformComponent):
+
+        def transform(self, dataframe, *, flag, value):
+            assert flag == "success"
+            assert value == 1
+            raise EarlyStopException()
+
+    # Mock CLI arguments
```
NielsRogge, line 39: In #72 I used the `mock.patch` annotator for mocking the argparse arguments:
```
@mock.patch('argparse.ArgumentParser.parse_args',
            return_value=argparse.Namespace(input_manifest_path=".", output_manifest_path="result.parquet", metadata=metadata))
```
could be used as an alternative to `sys.argv`

NielsRogge (approved):
Thanks for improving!

examples/pipelines/simple_pipeline/components/image_filtering/requirements.txt
PhilippeMoussalli: I think those are mainly used for testing the functionality of your branch code. Right now the workflow is a bit annoying to do so: you have to push your code and then reference the hash so that the component can rebuild with the latest version. Wondering if there isn't a better way to do. @NielsRogge any thoughts?

RobbeSneyders: Once we 'actually open source', the examples on main should always work though. For components to share, the version can still be fixed.

NielsRogge: Yes for now we probably need to include the commit hash when trying out new functionality, but when merging a PR it should always be updated to main.

tests/test_component.py
RobbeSneyders: There doesn't seem to be a clear choice: https://github.com/pytest-dev/pytest/issues/4576

I like the fact that with `monkeypatch`, the patched code is included inside of your test function, which is not the case with the `mock.patch` decorator.

NielsRogge:
Note that ideally also the simple pipeline gets updated and tested: https://github.com/ml6team/fondant/blob/1ac0fa0d2ed8196ebf4255b148823aab135e8353/examples/pipelines/simple_pipeline/components/image_filtering/src/main.py#L21

Cause we had updates to the Fondant code base in the past which turned out needed fixes to run the pipeline successfully on KubeFlow

PR #76, opened 2023-05-05
Remove legacy files

NielsRogge:
This PR removes all legacy files, namely the example components and the `hf_dataset_pipeline`.


RobbeSneyders (approved):
Thanks!

PR #77, opened 2023-05-05
Run black on tests and all examples

RobbeSneyders:
I will do the same for Ruff as well as a next step.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -40,6 +40,7 @@ repos:
                 examples/pipelines/finetune_stable_diffusion/components/image_filter_component|
                 examples/pipelines/finetune_stable_diffusion/components/embedding_component|
                 examples/pipelines/finetune_stable_diffusion/dataset_creation_pipeline.py|
+                tests/.*|
             )$
 
   - repo: https://github.com/pre-commit/mirrors-mypy
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -35,11 +35,7 @@ repos:
         files:  |
             (?x)^(
                 fondant/.*|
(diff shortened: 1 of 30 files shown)


.pre-commit-config.yaml
```diff
@@ -35,11 +35,8 @@ repos:
         files:  |
             (?x)^(
                 fondant/.*|
-                examples/pipelines/hf_dataset_pipeline|
-                examples/pipelines/finetune_stable_diffusion/components/load_from_hub_component|
-                examples/pipelines/finetune_stable_diffusion/components/image_filter_component|
-                examples/pipelines/finetune_stable_diffusion/components/embedding_component|
-                examples/pipelines/finetune_stable_diffusion/dataset_creation_pipeline.py|
+                examples/.*|
```
PhilippeMoussalli, line 38: Some of the examples from `finetune_stable_diffusion` are still from the initial merge so I'm not sure if we should go over all of them. Either we move them to an archive path or exclude them here. 

I think it's fine for black formatting but maybe when you start linting there will be a few things that will need to be fixed that might not be worth the effort since we're gonna re-implement the components with Fondant.

RobbeSneyders: Yes, you're right. @NielsRogge already removed some outdated examples and I won't spend any time fixing the outdated ones.

But for black it should indeed not matter.

PhilippeMoussalli (approved):
LGTM! thanks for adding this

GeorgesLorre:
for vscode I use: https://marketplace.visualstudio.com/items?itemName=ms-python.black-formatter

which formats my code on save

GeorgesLorre (approved):
Nice, I like black so far

PR #82, opened 2023-05-05
Move from caret to inequality requirements

RobbeSneyders:
I would have expected poetry to convert the caret requirements into something generic that other tools such as `pip` could handle as well, but apparently this is not the case.

Best practice for frameworks / libraries is also to not pin any upper limits unless there are clear incompatibilities, so I switched all the caret versioning to `>=`.

[Relevant discussion on the Poetry repo](https://github.com/orgs/python-poetry/discussions/3757#discussioncomment-435337).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,14 +39,14 @@ classifiers = [
 ]
 
 [tool.poetry.dependencies]
-python = "^3.8"
-jsonschema = "^4.17.3"
-dask = {extras = ["dataframe"], version = "^2023.4.1"}
+python = ">= 3.8"
+jsonschema = ">= 4.17.3"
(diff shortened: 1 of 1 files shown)


RobbeSneyders:
You can specify the version when adding a requirement:
```bash
poetry add "pendulum>=2.0.5"
```
Although I never use the `add` command and always edit the `pyproject.toml` manually :slightly_smiling_face:

GeorgesLorre:
> You can specify the version when adding a requirement:
> 
> ```shell
> poetry add "pendulum>=2.0.5"
> ```
> 
> Although I never use the `add` command and always edit the `pyproject.toml` manually 🙂

Ok, something to look out for when reviewing PR's then

GeorgesLorre (approved):

PR #84, opened 2023-05-08
[Alpha pipeline] Add prompt generating component

RobbeSneyders:
Fixes #84

Split this from #71 so we can review this component by component.

This PR contains the first component, which generates captions to query LAION-5B with. This is also the only component that is specific to this example, so I would place all other components in a `components/` directory.

I also think we can put examples directly under the `examples/` directory instead of `examples/pipelines`, as the example pipelines contain example components, and the reusable components in the future `components/` directory can serve as examples as well.


examples/controlnet-interior-design/components/generate_prompts/src/main.py
```diff
@@ -0,0 +1,130 @@
...
+
+    Returns:
+        str: prompt for the interior design model
+    """
+    # TODO: add more variation in the prompt, multiple variants, ...
+    return f"{prefix.lower()} {room.lower()}, {style.lower()} interior design"
+
+
+class LoadPromptsComponent(FondantLoadComponent):
+    def load(self, dataset_name):
```
RobbeSneyders, line 103: The `dataset_name` parameter is unused and can be removed.

examples/controlnet-interior-design/components/generate_prompts/src/main.py
```diff
@@ -0,0 +1,130 @@
...
+        prompts = list(map(lambda x: make_interior_prompt(*x), room_tuples))
+
+        df = pd.DataFrame(prompts, columns=["prompts_text"])
+
+        df = dd.from_pandas(
+            df, npartitions=1
+        )  # need to decide how npartitions should be set
+
+        index_list = [idx for idx in range(len(df))]
+        source_list = ["hub" for _ in range(len(df))]
```
RobbeSneyders, line 120: We're not using the hub as a source here.

examples/controlnet-interior-design/components/generate_prompts/src/main.py
```diff
@@ -0,0 +1,130 @@
...
+        """
+        Generate a set of initial prompts that will be used to retrieve images from the LAION-5B dataset.
+
+        Returns:
+            Dask dataframe
+        """
+        room_tuples = list(itertools.product(rooms, interior_prefix, interior_styles))
+        prompts = list(map(lambda x: make_interior_prompt(*x), room_tuples))
+
+        df = pd.DataFrame(prompts, columns=["prompts_text"])
```
RobbeSneyders, line 113: This column name doesn't match the name in the component specification.

examples/controlnet-interior-design/components/generate_prompts/src/main.py
```diff
@@ -0,0 +1,130 @@
...
+
+        df = dd.from_pandas(
+            df, npartitions=1
+        )  # need to decide how npartitions should be set
```
RobbeSneyders, line 117: Should be based on the amount of cores available and data size.
https://stackoverflow.com/a/46646366/4098821

 I think we should handle this as Fondant because:
- This is the complexity we want to take out of the user's hands
- The `dask` partitioning translates to `parquet` partitioning and so is also relevant for downstream components

We can do this after the user returns the dataframe leveraging [`repartition`](https://docs.dask.org/en/stable/generated/dask.dataframe.DataFrame.repartition.html), but then the user still needs to provide some placeholder value in the meantime.

Another option is that we provide some utility functionality for this.

CC: @GeorgesLorre

RobbeSneyders (commented):
Thanks @jamesbraniganml6 and @NielsRogge!

Left some comments.

examples/controlnet-interior-design/components/generate_prompts/src/main.py
GeorgesLorre: I agree we should handle this in Fondant. the argument is optional so I would just not use it here.

PhilippeMoussalli (commented):
Looking good so far :) thanks! 
Left a few coments

examples/controlnet-interior-design/components/generate_prompts/src/main.py
RobbeSneyders: It is optional, but only if you specify `chunk_size` instead.
https://github.com/dask/dask/blob/980fd92a46f101f161d21e0c4f486024bafbfad3/dask/dataframe/io/io.py#L263-L264

examples/controlnet-interior-design/components/generate_prompts/src/main.py
```diff
@@ -0,0 +1,130 @@
...
+        source_list = ["hub" for _ in range(len(df))]
+
+        df["id"] = da.array(index_list)
+        df["source"] = da.array(source_list)
+
+        return df
+
+
+if __name__ == "__main__":
+    component = LoadPromptsComponent.from_file()
```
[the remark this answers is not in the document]
RobbeSneyders: It uses the default location, which is `../fondant_component.yaml`.

examples/controlnet-interior-design/components/generate_prompts/src/main.py
```diff
@@ -0,0 +1,130 @@
...
+import pandas as pd
+
+from fondant.component import FondantLoadComponent
+from fondant.logger import configure_logging
+
+configure_logging()
+logger = logging.getLogger(__name__)
+
+
+interior_styles = [
```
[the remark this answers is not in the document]
RobbeSneyders: Then the question is if we even need this component, or if you shouldn't just pass your prompts to the LAION retriever directly.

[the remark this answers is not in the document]
PhilippeMoussalli: I would still create a parquet file out of them to potentially link the retrieved images with a given prompt based on ID

examples/controlnet-interior-design/components/generate_prompts/fondant_component.yaml
```diff
@@ -0,0 +1,14 @@
...
+image: generate_prompts::latest
+
+output_subsets:
+  prompts:
+    fields:
+      text:
+        type: utf8
+
+args:
+  dataset_name:
```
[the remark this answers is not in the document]
NielsRogge: Currently it's not possible to not include any args in the component spec. I created a ticket for it as this needs to be addressed first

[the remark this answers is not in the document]
NielsRogge: Update, it's now possible thanks to #87 so can indeed be removed

examples/controlnet-interior-design/components/generate_prompts/src/main.py
NielsRogge: What's the current way of creating the index? Just as 2 separate columns?

cc @GeorgesLorre

examples/controlnet-interior-design/components/generate_prompts/src/main.py
```diff
@@ -0,0 +1,130 @@
...
+
+    Args:
+        room (str): room name
+        prefix (str): prefix for the room
+        style (str): interior style
+
+    Returns:
+        str: prompt for the interior design model
+    """
+    # TODO: add more variation in the prompt, multiple variants, ...
```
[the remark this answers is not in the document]
NielsRogge: Can be addressed later on. Would focus on the prompts we have right now.

examples/controlnet-interior-design/components/generate_prompts/src/main.py
NielsRogge: Could you clarify? We have "prompts" as output subset with field "text"

NielsRogge (approved):
LGTM after remaining comments are addressed

examples/controlnet-interior-design/components/generate_prompts/src/main.py
RobbeSneyders: Might have been a mistake from my side. The field was called `data` in the use case doc. But indeed looks correct here.

examples/controlnet-interior-design/components/generate_prompts/fondant_component.yaml
```diff
@@ -0,0 +1,14 @@
+name: Generate prompts
```
NielsRogge, line 1: Also a kubeflow component spec is missing for now, but I assume the pipeline wrapper of #34 is going to add that

RobbeSneyders:
> LGTM after remaining comments are addressed

Can you pick this up @NielsRogge? I just split this PR from the other branch.

> Normally all components of a given pipeline should be placed in a components folder, otherwise it would become hard to see which components a certain pipeline consists of. I thought the plan was to have native components directly available in the Fondant SDK, not in the examples folder?

Indeed, I would add all other components as native / reusable components. So we should get the following directory structure:
```
- examples
  |- controlnet-interior-design
     |- components
        |- generate_prompts
     |- pipeline.py
- components
  |- prompt-based-LAION-retriever
  |- image-downloader
  |- image-resolution-filter
  ...
```

And the pipeline should use the reusable components.

examples/controlnet-interior-design/components/generate_prompts/fondant_component.yaml
RobbeSneyders: Yes, we shouldn't have to add one to the code, it should be handled by Fondant.

examples/controlnet-interior-design/components/generate_prompts/Dockerfile
```diff
@@ -0,0 +1,26 @@
+FROM --platform=linux/amd64 python:3.8-slim
+
+## System dependencies
+RUN apt-get update && \
+    apt-get upgrade -y && \
+    apt-get install git curl -y
+
+# Installing the package
+RUN mkdir -p /usr/local/gcloud \
```
PhilippeMoussalli, line 9: I think there are still some gcloud dependencies lingering here

Would remove all of 
```
# Installing the package
RUN mkdir -p /usr/local/gcloud \
&& tar -C /usr/local/gcloud -xvf /tmp/google-cloud-sdk.tar.gz \
&& /usr/local/gcloud/google-cloud-sdk/install.sh

# Adding the package path to local
ENV PATH $PATH:/usr/local/gcloud/google-cloud-sdk/bin
```

PhilippeMoussalli (commented):
Thanks! Left a few open comments to address

examples/pipelines/controlnet-interior-design/components/generate_prompts/fondant_component.yaml
```diff
@@ -0,0 +1,11 @@
...
+description: Component that generates a set of seed prompts
+image: generate_prompts::latest
+
+output_subsets:
+  prompts:
+    fields:
+      text:
+        type: utf8
+
+args:
```
PhilippeMoussalli, line 11: You should be able to remove this completely ... I think

examples/pipelines/controlnet-interior-design/controlnet_pipeline:dev.yaml
```diff
@@ -0,0 +1,57 @@
+apiVersion: argoproj.io/v1alpha1
```
PhilippeMoussalli, line 1: This file can be removed

PhilippeMoussalli (approved):
Thanks for adding this Niels :+1: ! One small comment but otherwise good to go

fondant/component_spec.py
RobbeSneyders, line 1: This is duplicate of #107. Merging that one first, will fix the coverage pipeline here as well I believe.

RobbeSneyders (commented):
Thanks @NielsRogge! 
LGTM (Can't approve my own PR)

PR #103, opened 2023-05-09
Add manifest documentation

### LINKED ISSUE #62: Create 'Manifest' documentation

RobbeSneyders:
Fixes #62

Please check if there's anything I forgot to explain :)

### DIFF
--- a/docs/manifest.md
+++ b/docs/manifest.md
@@ -1,0 +1,104 @@
+# Manifest
+
+A Fondant manifest describes the contents of a dataset and is used as a reference to the dataset 
+to be passed between Fondant components.
+
+Creating the manifest is handled by Fondant, as each component will automatically generate the 
+manifest that represents its output dataset. While you should not have to create the manifest 
+yourself, it is still useful to understand how it works.
+
+## Contents
+
+A manifest consists of the following sections
+
+```json
+{
+  "metadata": {
+    ...
+  },
+  "index": {
+    ...
+  },
+  "subsets": {
+    ...
+  }
+}
+```
+
+### Metadata
+
+The metadata section tracks metadata about the dataset. Currently, it tracks the location where 
+the data is stored and references to the pipeline and component that were used to create it.
+
+```json
+{
+  "metadata": {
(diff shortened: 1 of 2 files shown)


docs/manifest.md
```diff
@@ -0,0 +1,104 @@
...
+{
+  "index": {
+    "location": "/index"
+  }
+}
+```
+
+### Subsets
+
+Each subset is a different view on the data contained in the dataset. Different subsets could
```
NielsRogge, line 64: Wouldn't say it's a different view, it's just a part of the dataset

docs/manifest.md
```diff
@@ -0,0 +1,104 @@
...
+        "data": {
+          "type": "binary"
+        }
+      }
+    }
+  }
+}
+```
+
+All subsets should contain at least the data points defined by the index.
```
PhilippeMoussalli, line 104: Question regarding this point, say that we have an image subset and we gathered different subsets linked to it (captions, aesthetic score, ...) where each subset was probably obtained in a separate component. Then we extend the images subset (e.g. laion retrieval) and the index along with it.

seed_images (image_subset)-> caption_component (add caption subset) -> aesthetic_filter (add aesthetic score subset) -> **image retrieval (add index and extend image_subset)** 

How do we handle the missing entries in the other subsets (captions, aesthetic score, ...) if we only plan on including them using subsequent components? 

... -> **image retrieval (add index and extend image_subset)** -> caption_component -> filter_component

I think this is not possible with the current framework? This question might have already been addressed and I could have missed.

docs/manifest.md
RobbeSneyders: Reworded into "different features".

docs/manifest.md
RobbeSneyders: Indeed, this is not supported. There's two ways around this:
- Reorder the components in your pipeline (eg. in your example, move the image retrieval to the 2nd place)
- Work with non-linear DAGs when this is supported in the future (eg. in your example, the 2nd part of the pipeline could branch off and later merge again)

PR #108, opened 2023-05-09
Handle kfp types

PhilippeMoussalli:
This PR enables the user to pass different `bool`, `lists` and `dict` to a component. Kubeflow typically handles those arguments by serializing them as a string . For this reason, they need to be de-serialized again within the component in order for them to be properly handled. 

This might go away once we move to V2. 

References to the issue: 
https://github.com/kubeflow/pipelines/issues/7457
https://github.com/kubeflow/pipelines/issues/7719


RobbeSneyders:
Do we only need this for the user arguments?

We could solve this cleanly as well by registering [custom types](https://docs.python.org/3/library/argparse.html#type) in the `kubeflow2python_type` dict.

PhilippeMoussalli:
> Do we only need this for the user arguments?
> 
> We could solve this cleanly as well by registering [custom types](https://docs.python.org/3/library/argparse.html#type) in the `kubeflow2python_type` dict.

Didn't know about custom types :) I just implemented it and indeed it's neater than the previous approach :) thanks for suggesting

PR #110, opened 2023-05-10
[Alpha pipeline] Add image retrieval component

### LINKED ISSUE #95: Prompt-based LAION retrieval component

NielsRogge:
This PR adds the image retrieval component, which retrieves image URLs using clip-retrieval from LAION-5B.

Fixes #95


examples/pipelines/controlnet-interior-design/components/generate_prompts/src/main.py
```diff
@@ -114,6 +114,17 @@ def load(self) -> dd.DataFrame:
...
+        # TODO remove, just use a tiny df for testing purposes
+        data = {
+            "prompts_text": [
+                "comfortable bathroom, art deco interior design",
+                "comfortable bathroom, bauhaus interior design",
+            ]
+        }
+        pandas_df = pd.DataFrame.from_dict(data)
+        df = dd.from_pandas(pandas_df, npartitions=1)
+        # end of TODO
```
NielsRogge, line 126: Would keep this as long as we're building the pipeline. Would remove once the pipeline is finished

examples/pipelines/controlnet-interior-design/components/retrieve_images/fondant_component.yaml
```diff
@@ -0,0 +1,27 @@
...
+
+args:
+  num_images:
+    description: Number of images to retrieve for each prompt
+    type: int
+  aesthetic_score:
+    description: Ranking score for aesthetic, higher is prettier
+    type: int
+  aesthetic_weight:
+    description: Weight of the aesthetic score, between 0 and 1
```
PhilippeMoussalli, line 26: what's the difference between the weight and the score?

examples/pipelines/controlnet-interior-design/components/retrieve_images/fondant_component.yaml
```diff
@@ -0,0 +1,27 @@
...
+      url:
+        type: utf8
+  additionalSubsets: false
+
+args:
+  num_images:
+    description: Number of images to retrieve for each prompt
+    type: int
+  aesthetic_score:
+    description: Ranking score for aesthetic, higher is prettier
```
PhilippeMoussalli, line 23: could you indicate the upper limit here (I guess its 9)

examples/pipelines/controlnet-interior-design/components/retrieve_images/src/main.py
```diff
@@ -0,0 +1,89 @@
...
+        dataframe["images_url"] = dataframe["prompts_text"].map_partitions(
+            lambda example: example.apply(query_clip_client, args=(client,)),
+            meta=("images_url", "str"),
+        )
+
+        # unpack list of urls
+        dataframe = dataframe.explode("images_url")
+
+        # add index
+        index_list = [idx for idx in range(len(dataframe))]
```
PhilippeMoussalli, line 78: For this one you can do as follows: 
```
dask_df['id'] = dask_df.assign(id=1).row_num.cumsum()
```

Here, we take a temporary column and assign it a value of 1, then we use `cumsum()` to calculative the cumulative sum of all previous values (emulating your `range(len)`). 

You will get the same results but without any array/list in memory. 

I don't expect users to figure this one right off the bat so either we make it very explicit with documentation or we handle this internally

examples/pipelines/controlnet-interior-design/components/retrieve_images/src/main.py
```diff
@@ -0,0 +1,89 @@
...
+        source_list = ["seed" for _ in range(len(dataframe))]
+
+        dataframe["id"] = da.array(index_list)
+        dataframe["source"] = da.array(source_list)
+
+        return dataframe
+
+
+if __name__ == "__main__":
+    component = RetrieveImagesComponent.from_file("../fondant_component.yaml")
```
PhilippeMoussalli, line 88: you can remove `"../fondant_component.yaml"` since it's the default

examples/pipelines/controlnet-interior-design/config/components_config.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+@dataclass
+class RetrieveImagesConfig:
+    """
+    Configs for the retrieve images component
+
+    Args:
+        NUM_IMAGES (str): Number of images to retrieve for each prompt.
+        AESTHETIC_SCORE (int): Ranking score for aesthetic, higher is prettier.
```
PhilippeMoussalli, line 11: same comment for the score

examples/pipelines/controlnet-interior-design/components/retrieve_images/src/main.py
```diff
@@ -0,0 +1,89 @@
...
+            modality=Modality.IMAGE,
+        )
+
+        dataframe["images_url"] = dataframe["prompts_text"].map_partitions(
+            lambda example: example.apply(query_clip_client, args=(client,)),
+            meta=("images_url", "str"),
+        )
+
+        # unpack list of urls
+        dataframe = dataframe.explode("images_url")
```
PhilippeMoussalli, line 75: just for clarity, this is used because each row in the dataframe stores a list of urls right?

PhilippeMoussalli (commented):
Thanks Niels :) Looks good! I left a few comments.  
How did you manage to resolve the issues with slow retrieval eventually?

examples/pipelines/controlnet-interior-design/components/retrieve_images/src/main.py
```diff
@@ -0,0 +1,89 @@
...
+    Returns:
+        results: a list of strings, each representing a URL of an image related to the query
+    """
+    results = client.query(text=text)
+    results = [i["url"] for i in results]
+
+    return results
+
+
+class RetrieveImagesComponent(FondantTransformComponent):
```
jamesbraniganml6, line 37: Perhaps LAIONRetrievalComponent (or something to that effect) might be a better name here, as the component is quite specific to that dataset. Let me know what you think.

fondant/data_io.py
```diff
@@ -140,7 +141,9 @@ def write_index(self, df: dd.DataFrame):
         )
 
         # Write index
-        dd.compute(upload_index_task)
+        with ProgressBar():
```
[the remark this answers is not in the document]
NielsRogge: It looks like this:

Screenshot 2023-05-10 at 17 38 17

It's not ideal as it seems to replicate the progress bars across cores? Not sure. Do we know how many cores the machines on Kubeflow have?

examples/pipelines/controlnet-interior-design/components/retrieve_images/fondant_component.yaml
```diff
@@ -0,0 +1,27 @@
+name: LAION retrieval
+description: A component that retrieves image URLs from LAION-5B based on a set of seed prompts
+image: laion_retrieval::latest
```
RobbeSneyders, line 3:
```suggestion
image: laion_retrieval:latest
```

fondant/data_io.py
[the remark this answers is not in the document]
RobbeSneyders: Will depend on the machine defined for the component, but often quite a few I think. And also multiple machines in the future.

RobbeSneyders (commented):
Thanks @NielsRogge!

RobbeSneyders:
As discussed before, I would make these reusable components available in a `components` directory directly under the root directory. We can merge this one as is though, as we will have to provide a mechanism to easily load the reusable components. We can implement this using this component as an example in a separate PR.

examples/pipelines/controlnet-interior-design/components/retrieve_images/src/main.py
PhilippeMoussalli: what I meant by emulating `range(len)` is that you can set the index incrementally with `cumsum()` but without loading anything in memory. There is some computation that happens but should be lightweight

RobbeSneyders: Yes, your solution is even better, just as a side note :slightly_smiling_face: 

But shouldn't we actually leverage the LAION ids here?

examples/pipelines/controlnet-interior-design/components/retrieve_images/fondant_component.yaml
NielsRogge: CLIP-retrieval uses [this model](https://github.com/LAION-AI/aesthetic-predictor) to do the following operation:
```
aesthetic_embedding = clip_resource.aesthetic_embeddings[aesthetic_score]
query = query + aesthetic_embedding * aesthetic_weight
query = query / np.linalg.norm(query)
```

examples/pipelines/controlnet-interior-design/components/retrieve_images/src/main.py
```diff
@@ -0,0 +1,89 @@
...
+        client = ClipClient(
+            url="https://knn.laion.ai/knn-service",
+            indice_name="laion5B-L-14",
+            num_images=num_images,
+            aesthetic_score=aesthetic_score,
+            aesthetic_weight=aesthetic_weight,
+            modality=Modality.IMAGE,
+        )
+
+        dataframe["images_url"] = dataframe["prompts_text"].map_partitions(
```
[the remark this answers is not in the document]
NielsRogge: @jamesbraniganml6 defined `map_partitions` to do the `apply` operation on each partition, and hence leverage parallellism. However I think I can remove this as Fondant will handle the optimizations

examples/pipelines/controlnet-interior-design/components/retrieve_images/src/main.py
NielsRogge: Yes, each row stores a list of URLs so we want to flatten them to several rows, see https://docs.dask.org/en/stable/generated/dask.dataframe.DataFrame.explode.html

examples/pipelines/controlnet-interior-design/components/retrieve_images/src/main.py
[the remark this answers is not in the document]
RobbeSneyders: `.apply()` should already leverage parallelism, so I don't think we win anything with the `.map_partitions()`.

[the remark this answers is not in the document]
NielsRogge: Ok, will remove

examples/pipelines/controlnet-interior-design/components/retrieve_images/src/main.py
NielsRogge: @PhilippeMoussalli this gives me: `AttributeError: 'DataFrame' object has no attribute 'row_num'`

PhilippeMoussalli: sorry it should be `id` instead of `row_num`

```
dask_df['id'] = dask_df.assign(id=1).id.cumsum()

```

examples/pipelines/controlnet-interior-design/components/generate_prompts/src/main.py
```diff
@@ -114,11 +114,23 @@ def load(self) -> dd.DataFrame:
...
+
+        # add id and source columns
+        df["id"] = df.assign(id=1).id.cumsum()
```
RobbeSneyders, line 129: We should use the LAION ids here.

NielsRogge: Can I leave this for a follow-up PR? Would like to first get all components get up and running

PR #111, opened 2023-05-10
[Alpha pipeline] Add captioning component

### LINKED ISSUE #98: Captioning component

NielsRogge:
This PR adds a generic image captioning component.

Fixes #98


examples/pipelines/controlnet-interior-design/components/caption_images/src/main.py
```diff
@@ -0,0 +1,106 @@
...
+        # join lists into a single Dask delayed object
+        captions = flatten(captions)
+        delayed_series = dd.from_delayed(captions, meta=pd.Series(dtype="str"))
+        captions_df = delayed_series.to_frame(name="captions_text")
+
+        return captions_df
+
+
+if __name__ == "__main__":
+    component = CaptionImagesComponent.from_file("../fondant_component.yaml")
```
RobbeSneyders, line 105:
```suggestion
    component = CaptionImagesComponent.from_file()
```

examples/pipelines/controlnet-interior-design/components/caption_images/src/main.py
RobbeSneyders, line 1: Would love to see if we can translate this to the dataframe interface in the future, but for now this is fine for me.

RobbeSneyders (commented):
Thanks @NielsRogge!

LGTM apart from some small comments. Would like to get @PhilippeMoussalli's review on the Dask implementation as well though.

examples/pipelines/controlnet-interior-design/components/caption_images/src/main.py
```diff
@@ -0,0 +1,106 @@
...
+import toolz
+
+import dask
+import dask.dataframe as dd
+import pandas as pd
+
+from fondant.component import FondantTransformComponent
+from fondant.logger import configure_logging
+
+from transformers import AutoProcessor, AutoModelForCausalLM
```
PhilippeMoussalli, line 16: Can you group the 3rd party imports together?

examples/pipelines/controlnet-interior-design/components/caption_images/src/main.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+@dask.delayed
+def collate(examples):
+    return torch.cat([ex["pixel_values"] for ex in examples])
+
+
+@dask.delayed
+def caption(batch, model, processor):
+    generated_ids = model.generate(pixel_values=batch, max_length=50)
```
PhilippeMoussalli, line 46: I think the `max_length` should also be a component parameter

examples/pipelines/controlnet-interior-design/components/caption_images/src/main.py
```diff
@@ -0,0 +1,106 @@
...
+        ]
+
+        # batch images together
+        batches = [
+            collate(batch)
+            for batch in toolz.partition_all(batch_size, transformed_images)
+        ]
+
+        # caption images
+        gpu_model = dask.delayed(model.to(device))
```
PhilippeMoussalli, line 93: does this also need to be delayed? Have you tried passing it without?

examples/pipelines/controlnet-interior-design/components/caption_images/src/main.py
```diff
@@ -0,0 +1,106 @@
...
+        captions = [caption(batch, gpu_model) for batch in batches]
+
+        # join lists into a single Dask delayed object
+        captions = flatten(captions)
+        delayed_series = dd.from_delayed(captions, meta=pd.Series(dtype="str"))
+        captions_df = delayed_series.to_frame(name="captions_text")
+
+        return captions_df
```
PhilippeMoussalli, line 101: Does this dataframe contain the ids?

PhilippeMoussalli (commented):
Thanks Niels! This will be a nice example for future components that involve inferencing with a model.
Left a few comments

RobbeSneyders:
Do we actually need this component for the ControlNet pipeline? We already have captions from LAION, no?

examples/pipelines/controlnet-interior-design/components/caption_images/src/main.py
```diff
@@ -0,0 +1,115 @@
...
+            Dask dataframe
+        """
+        device = "cuda" if torch.cuda.is_available() else "cpu"
+
+        processor = AutoProcessor.from_pretrained(model_id)
+        model = AutoModelForCausalLM.from_pretrained(model_id)
+
+        # load and transform the images
+        images = dataframe["images_data"]
+        loaded_images = [load(image) for image in images]
```
bertChristiaens1, line 86: Have we tested this with larger amounts of data? It seems that e.g. naively loading 1M+ images into memory could use a lot of RAM

examples/pipelines/controlnet-interior-design/components/caption_images/src/main.py
```diff
@@ -0,0 +1,115 @@
...
+        transformed_images = [
+            transform(image, processor, device) for image in loaded_images
```
bertChristiaens1, line 88: Also here, moving all the images to the device could be problematic (certainly if device='cuda'). This would give CUDA Out of memory

examples/pipelines/controlnet-interior-design/run_locally.sh
RobbeSneyders, line 1: Nice, but I assume this isn't meant to be included with your hard-coded arguments?

Would be great if the pipeline could do this, then we don't even need a kfp cluster locally. Not sure how this would work with hardware requirements though.

components/prompt_based_laion_retrieval/src/main.py
```diff
@@ -29,7 +29,10 @@ def query_clip_client(text: str, client: ClipClient) -> List[str]:
         urls: a list of strings, each representing a URL of an image related to the query
     """
     results = client.query(text=text)
-    urls = [i["url"] for i in results]
+    try:
+        urls = [i["url"] for i in results]
+    except:
+        urls = ["" for _ in range(len(results))]
```
RobbeSneyders, line 35: Why is this try except needed? Would prefer to at least not have a bare `except` (and I'm surprised ruff doesn't complain about this)

components/prompt_based_laion_retrieval/src/main.py
```diff
@@ -82,6 +88,11 @@ def transform(
         # reorder columns
         dataframe = dataframe[["id", "source", "images_url"]]
 
+        dataframe = dataframe.astype({'id': 'string', 'source': 'string'})
+
+        print("Final dataframe:", dataframe.head(4))
+        dataframe = dataframe.reset_index(drop=True)
```
RobbeSneyders, line 94: Why are you resetting the index here?

examples/pipelines/controlnet-interior-design/components/caption_images/src/main.py
```diff
@@ -0,0 +1,126 @@
...
+
+        Returns:
+            Dask dataframe
+        """
+        device = "cuda" if torch.cuda.is_available() else "cpu"
+
+        processor = AutoProcessor.from_pretrained(model_id)
+        model = AutoModelForCausalLM.from_pretrained(model_id)
+
+        print("Length of the input dataframe:", len(dataframe))
```
RobbeSneyders, line 85: These `len()` calls trigger a compute, which will have a high cost when working with larger than memory data. We probably want to include logging the input and output size lazily in the component class.

RobbeSneyders (commented):
Thanks @NielsRogge!

components/prompt_based_laion_retrieval/src/main.py
```diff
@@ -66,6 +69,9 @@ def transform(
             modality=Modality.IMAGE,
         )
 
+        print("Input dataframe:", dataframe.head(2))
+
+        logger.info("Retrieving URLs...")
```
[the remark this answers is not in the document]
NielsRogge: Yes, although logging complains when providing non-string objects like `dataframe.head(2)` to see the first few rows

components/prompt_based_laion_retrieval/src/main.py
NielsRogge: Because @GeorgesLorre asked me to do so :D

examples/pipelines/controlnet-interior-design/components/caption_images/src/main.py
NielsRogge: Yes I know, its mainly for debugging

examples/pipelines/controlnet-interior-design/components/caption_images/src/main.py
```diff
@@ -0,0 +1,128 @@
...
+        Returns:
+            Dask dataframe
+        """
+        device = "cuda" if torch.cuda.is_available() else "cpu"
+        logger.info("Device:", device)
+
+        processor = AutoProcessor.from_pretrained(model_id)
+        model = BlipForConditionalGeneration.from_pretrained(model_id)
+
+        print("Length of the input dataframe:", len(dataframe))
```
PhilippeMoussalli, line 86: would remove those print statements before merging

examples/pipelines/controlnet-interior-design/components/caption_images/src/main.py
```diff
@@ -0,0 +1,128 @@
...
+        delayed_series = dd.from_delayed(captions, meta=pd.Series(dtype="str"))
+        captions_df = delayed_series.to_frame(name="captions_text")
+
+        # add index columns
+        captions_df["id"] = dataframe["id"].reset_index(drop=True)
+        captions_df["source"] = dataframe["source"].reset_index(drop=True)
+
+        captions_df = captions_df.reset_index(drop=True)
+
+        print("Final dataframe:", captions_df.head(4))
```
PhilippeMoussalli, line 120: same here

examples/pipelines/controlnet-interior-design/components/caption_images/src/main.py
```diff
@@ -0,0 +1,128 @@
...
+        # join lists into a single Dask delayed object
+        captions = flatten(captions)
+        delayed_series = dd.from_delayed(captions, meta=pd.Series(dtype="str"))
+        captions_df = delayed_series.to_frame(name="captions_text")
+
+        # add index columns
+        captions_df["id"] = dataframe["id"].reset_index(drop=True)
+        captions_df["source"] = dataframe["source"].reset_index(drop=True)
+
+        captions_df = captions_df.reset_index(drop=True)
```
GeorgesLorre, line 118: the number of rows does not change in this component right ? We only add columns ?

examples/pipelines/controlnet-interior-design/components/caption_images/src/main.py
GeorgesLorre, line 86: Could you remove the print statements or convert them to logs?

examples/pipelines/controlnet-interior-design/pipeline.py
```diff
@@ -42,11 +42,22 @@
         "max_aspect_ratio": 2.5,
     },
 )
+caption_images_op = ComponentOp(
+    component_spec_path="components/caption_images/fondant_component.yaml",
+    arguments={
+        "model_id": "Salesforce/blip-image-captioning-base",
+        "batch_size": 2,
```
PhilippeMoussalli, line 49: would make the batch larger before merging

examples/pipelines/controlnet-interior-design/run_locally.sh
```diff
@@ -0,0 +1,19 @@
+export METADATA='{"run_id":"test_niels", "base_path":"/Users/nielsrogge/Documents/fondant_artifacts"}'
```
GeorgesLorre, line 1: I would not merge this into main but it is a nice script we can start improving to speed up local development

PhilippeMoussalli, line 1: I'm a bit conflicted about having this merged on the main branch.
Do we want to set an example/way of working on running things locally like this @GeorgesLorre ? 
Maybe leave out your name from it and just keep some defaults

PhilippeMoussalli (approved):
Approved! Thanks for implementing this :) 
Left a few minor comments

examples/pipelines/controlnet-interior-design/components/caption_images/src/main.py
GeorgesLorre: Then we should not have to drop the index in this case since it does not change

PR #114, opened 2023-05-11
Add workflow to publish package to PyPI

### LINKED ISSUE #112: Add pypi release pipeline in github actions
We need a release pipeline to release the `fondant` package to PyPI.

RobbeSneyders:
Fixes #112 

Adapted from the [connexion pipeline](https://github.com/spec-first/connexion/blob/main/.github/workflows/release.yml) to leverage "trusted publishing" using OpenID connect.

We leverage the reusable [gh-action-pypi-publish](https://github.com/pypa/gh-action-pypi-publish) action.

I don't think there's away around merging this to test it.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,44 @@
+name: Publish Python 🐍 distributions 📦 to PyPI and TestPyPI
+on:
+  push:
+    tags:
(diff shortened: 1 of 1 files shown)


PhilippeMoussalli (approved):
Thanks for adding this!

PR #115, opened 2023-05-11
Modify controlnet pipeline to use new pipeline structure

PhilippeMoussalli:
PR to switch over the controlnet pipeline to use the new Fondant interface.


examples/pipelines/controlnet-interior-design/pipeline.py
```diff
@@ -1,58 +1,38 @@
...
-)
-laion_retrieval_op = comp.load_component(
-    "components/laion_retrieval/kubeflow_component.yaml"
+generate_prompts_op = FondantComponentOp(
+    component_spec_path="components/generate_prompts/fondant_component.yaml"
 )
 
+laion_retrieval_op = FondantComponentOp(
+    component_spec_path="components/laion_retrieval/fondant_component.yaml",
+    arguments={"num_images": 2, "aesthetic_score": 9, "aesthetic_weight": 0.5},
```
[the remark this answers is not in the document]
PhilippeMoussalli: I think they are defined quite explicitly here so don't see any clear advantage of having to define them in one place and call them in another.

NielsRogge (approved):
Thanks, looks a lot simpler now

examples/pipelines/controlnet-interior-design/pipeline.py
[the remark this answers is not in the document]
NielsRogge: Ok, fine for me

examples/pipelines/controlnet-interior-design/pipeline.py
```diff
@@ -1,58 +1,38 @@
...
+import sys
 
-from kfp import components as comp
-from kfp import dsl
+sys.path.append("../")
 
-from config.general_config import KubeflowConfig
-from config.components_config import RetrieveImagesConfig
-from fondant.pipeline_utils import compile_and_upload_pipeline
+from pipeline_configs import PipelineConfigs
```
RobbeSneyders, line 8: Where is this imported from?

PhilippeMoussalli: It's here https://github.com/ml6team/fondant/blob/main/examples/pipelines/pipeline_configs.py at the root level, not sure why it does not show up on the github UI though

RobbeSneyders: Ok thanks, I was looking for a new file.

PR #118, opened 2023-05-12
Add mechanism for reusable components

### LINKED ISSUE #117: Importing mechanism for reusable components
We should provide an importing mechanism for our reusable components so they can easily be included in user pipelines.

Loading a local component is done as follows:
```Python
from fondant.pipeline import FondantComponentOp

component_op = FondantComponentOp(
    component_spec_path="...",
    arguments={...},
)
```
where the `component_spec_path` is a local path.

We no longer want the user to have to specify the `component_spec_path`, but still want them to be able to provide custom arguments.

We can provide a single `ReusableComponent` class which takes the name of the component to load
```Python
from fondant.components import ReusableComponentOp

component_op = ReusableComponentOp(
    name="...",
    arguments={...},
)
```

Which we can optionally still wrap as specific classes:
```Python
from fondant.components import CaptioningComponentOp

captioning_op = CaptioningComponent(
    arguments={...},
)
```

### LINKED ISSUE #124: Add release pipeline for building docker images

RobbeSneyders:
Fixes #117 
Fixes #124 

This PR does a couple of things:
- Move reusable components into the top-level `components` directory
- Add the `ReusableComponentOp` class which can load a reusable component based on its name
- Build the component images as part of the release workflow


components/prompt_based_laion_retrieval/src/clip_client.py
```diff
@@ -7,6 +7,8 @@
 Copyright (c) 2021 Romain Beaumont
 """
 
+# mypy: ignore-errors
```
RobbeSneyders, line 10: Ignore errors since this is a copied file.

components/prompt_based_laion_retrieval/requirements.txt
```diff
@@ -0,0 +1,2 @@
+fondant
```
RobbeSneyders, line 1: This uses the latest version by default, but the released version is frozen during the release process.

fondant/component.py
```diff
@@ -179,8 +179,14 @@ def _load_or_create_manifest(self) -> Manifest:
         return Manifest.from_file(self.args.input_manifest_path)
 
     @abstractmethod
-    def transform(self, *, dataframe: dd.DataFrame, **kwargs) -> dd.DataFrame:
-        """Abstract method for applying data transformations to the input dataframe."""
+    def transform(self, *args, **kwargs) -> dd.DataFrame:
```
RobbeSneyders, line 182: This led to typing issues before since the signature in the subclasses contain specific arguments instead of `**kwargs`. mypy has a [mechanism](https://github.com/python/mypy/issues/5876) to enable this, but it only works if you specify both `*args` and `**kwargs`.

fondant/pipeline.py
```diff
@@ -57,6 +58,51 @@ def component_spec(self) -> FondantComponentSpec:
         return FondantComponentSpec.from_file(self.component_spec_path)
 
 
+class ReusableComponentOp:
```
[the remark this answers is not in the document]
RobbeSneyders: Good idea. Updated it.

components/prompt_based_laion_retrieval/requirements.txt
GeorgesLorre: nice

PhilippeMoussalli (approved):
LGTM! Thanks for implementing this :)

RobbeSneyders:
Let's see if the pipeline works on main :)

fondant/pipeline.py
```diff
@@ -45,17 +48,68 @@ class FondantComponentOp:
...
+        """
+        component_spec_path = (
+            files("fondant") / f"components/{name}/fondant_component.yaml"
+        )
+        component_spec_path = t.cast(Path, component_spec_path)
+
+        if not (component_spec_path.exists() and component_spec_path.is_file()):
+            raise ValueError(f"No reusable component with name {name} found.")
+
+        return FondantComponentOp(
```
GeorgesLorre, line 104: Cool

GeorgesLorre (approved):

PR #119, opened 2023-05-12
Add a better index

GeorgesLorre:
Note that if you set a column as index it stops being a column: https://docs.dask.org/en/stable/generated/dask.dataframe.DataFrame.set_index.html


fondant/data_io.py
```diff
@@ -134,7 +132,7 @@ def write_index(self, df: dd.DataFrame):
         index_df = df[index_columns]
 
         # set index
-        index_df.set_index("id")
+        index_df = index_df.set_index("id")
```
PhilippeMoussalli, line 135: Getting 
```
 "Data has no column 'id': use any column of ['source']"
```
when trying to run `simple_pipeline`. 

Could you try replicating the issue: 
inside `examples/pipelines/simple_pipeline/components/load_from_hub/src`

you can run 
```
python3 main.py  --output_manifest_path a.txt --metadata '{"run_id":"test", "base_path":"soy-audio-379412-kfp-output"}' --dataset_name lambdalabs/pokemon-blip-captions
```

fondant/data_io.py
```diff
@@ -209,7 +207,11 @@ def create_subset_dataframe(
             subset_df = create_subset_dataframe(subset_name, subset_columns, df)
 
             # set index
-            subset_df.set_index("id")
+            if subset_df.index.name != "uid":
+                subset_df["uid"] = (
+                    subset_df["id"].astype("str") + "_" + subset_df["source"]
```
PhilippeMoussalli, line 212: General question, do we always expect the users to add an ID + source when creating/adding new subsets? I guess there is no way around it. We just have to make sure it's well documented

PhilippeMoussalli (approved):
Thanks Georges! Tried it out and it works like a charm

fondant/data_io.py
GeorgesLorre: Well is is loaded in the dataframe we supply to the user so it should already be there for adding columns/subsets for adding rows they will need to add ID's and sources.

fondant/data_io.py
```diff
@@ -14,11 +14,7 @@ class DataIO:
     def __init__(self, manifest: Manifest):
         self.manifest = manifest
         self.index_fields = ["id", "source"]
-        self.index_schema = {
-            "source": "string",
-            "id": "int64",
-            "__null_dask_index__": "int64",
-        }
+        self.index_schema = {"source": "string", "id": "int64"}
```
RobbeSneyders, line 17: I know this was already in here, but shouldn't `id` be a `string` as well?

fondant/data_io.py
RobbeSneyders, line 212: I think concatenating `{source}_{id}` instead of `{id}_{source}` might be less costly, since chances are that the current order of the records resembles this more already.

fondant/data_io.py
```diff
@@ -209,7 +207,11 @@ def create_subset_dataframe(
             subset_df = create_subset_dataframe(subset_name, subset_columns, df)
 
             # set index
-            subset_df.set_index("id")
+            if subset_df.index.name != "uid":
+                subset_df["uid"] = (
+                    subset_df["id"].astype("str") + "_" + subset_df["source"]
+                )
+                subset_df = subset_df.set_index("uid")
```
RobbeSneyders, line 214: Shouldn't we pass in the divisions so they are the same across index and all subsets?

tests/test_data_io.py
```diff
@@ -70,6 +74,22 @@ def test_write_index(tmp_path_factory, dataframe, manifest):
         df = dd.read_parquet(fn / "index")
         assert len(df) == NUMBER_OF_TEST_ROWS
         assert list(df.columns) == ["id", "source"]
+        assert df.index.name == "uid"
+
+
+def test_rewrite_index(tmp_path_factory, dataframe, manifest):
+    """Test writing out the index subset without a dask index."""
```
RobbeSneyders, line 81: It took me a while to figure out what this test was doing. If I understand correctly, it checks that the index is created correctly when writing, right? Can you add some explanation on this in the docstring?

Same for the subset test below.

RobbeSneyders (commented):
Thanks Georges.

Left some comments while trying to wrap my head around this :sweat_smile:

GeorgesLorre:
@RobbeSneyders I addressed your comments

fondant/data_io.py
```diff
@@ -134,7 +130,9 @@ def write_index(self, df: dd.DataFrame):
         index_df = df[index_columns]
 
         # set index
-        index_df.set_index("id")
+        if index_df.index.name != "uid":
+            index_df["uid"] = index_df["source"] + "_" + index_df["id"].astype("str")
```
RobbeSneyders, line 134: I think we can drop the `astype("str")` here now. Below as well.

fondant/data_io.py
RobbeSneyders: Any input on this @GeorgesLorre?

GeorgesLorre: I'm looking into this, can I take it up in a different PR ?

RobbeSneyders: Ok, I think we should still align them between the index and subsets as well, but let's do that as a follow up.

PR #136, opened 2023-05-15
Rename input_subsets/output_subsets to consumes/produces

RobbeSneyders:
I was working on the component spec documentation and became less and less of a fan of the `input_subsets` and `output_subsets` naming. I think they can more clearly be described using `consumes` and `produces`.
- `consumes` describes which subsets a component actually "consumes", while allowing freedom to "accept" additional input subsets using the `additionalSubsets` flag.
- `produces` describes which subsets a component actually produces, while allowing freedom to "pass along" additional output subsets using the `additionalSubsets` flag.

Once we release, it would be hard or even impossible to change his.

### DIFF
--- a/fondant/component_spec.py
+++ b/fondant/component_spec.py
@@ -140,40 +140,34 @@ class ComponentSpec:
         return self._specification["image"]
 
     @property
-    def input_subsets(self) -> t.Mapping[str, ComponentSubset]:
-        """The input subsets of the component as an immutable mapping."""
+    def consumes(self) -> t.Mapping[str, ComponentSubset]:
+        """The subsets consumed by the component as an immutable mapping."""
         return types.MappingProxyType(
             {
                 name: ComponentSubset(subset)
-                for name, subset in self._specification.get("input_subsets", {}).items()
+                for name, subset in self._specification.get("consumes", {}).items()
                 if name != "additionalSubsets"
             }
         )
 
     @property
-    def output_subsets(self) -> t.Mapping[str, ComponentSubset]:
-        """The output subsets of the component as an immutable mapping."""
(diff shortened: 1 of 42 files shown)


fondant/component_spec.py
```diff
@@ -140,40 +140,34 @@ def image(self):
         return self._specification["image"]
 
     @property
-    def input_subsets(self) -> t.Mapping[str, ComponentSubset]:
+    def consumes(self) -> t.Mapping[str, ComponentSubset]:
         """The input subsets of the component as an immutable mapping."""
```
PhilippeMoussalli, line 144: Do we then still describe the `consumed subsets` as `input subsets`? or is input subsets just indicative of the additional subsets defined with the `additionalSubsets` flag. If we're going for this notation switch would rather remove the term input subsets completely to avoid confusion

fondant/manifest.py
```diff
@@ -209,19 +209,19 @@ def evolve(  # noqa : PLR0912 (too many branches)
...
         # If additionalSubsets is False in component output subsets,
         # Remove all subsets from the manifest that are not listed
-        if not component_spec.produces_additional_subsets:
+        if not component_spec.outputs_additional_subsets:
             for subset_name in evolved_manifest.subsets:
-                if subset_name not in component_spec.output_subsets:
+                if subset_name not in component_spec.produces:
                     evolved_manifest.remove_subset(subset_name)
 
         # If additionalFields is False for input subsets,
```
PhilippeMoussalli, line 222: Those comments need to be updated if we want to remove the term `input subsets`

PhilippeMoussalli (approved):
Thanks Robbe! Left a few comments. 
Would just double check that there are no hanging `input subsets` `output subsets` left in the repo with a search and replace tool

PR #137, opened 2023-05-16
Add symlink to prompt_based_laion_retrieval in ControlNet example

RobbeSneyders:
This PR adds a temporary symlink from the ControlNet example pipeline to the `prompt_based_laion_retrieval` component to be able to develop the example pipeline more easily while we develop Fondant. We can now just work as if the component was still located in the example directory.

The symlink should be removed before alpha release.


NielsRogge (commented):
Thanks, could you confirm that running `sh build_images.sh --component prompt_based_laion_retrieval` from the pipeline directory works?

PhilippeMoussalli (approved):
Thanks Robbe, 
Could you let us know how to this? I will need to do it for all other shared components

PR #140, opened 2023-05-16
[Alpha pipeline] Write to hub component

PhilippeMoussalli:
PR that adds the write to hub component. 

Right now we're running `to_parquet` method twice, once within the component itself to write to the hub and another time in fondant to write the dataset. We should work on implementing a `DataWriter` class huggingface/huggingface_hub#138. 

Another thing that we need to configure is how to properly pass secrets huggingface/huggingface_hub#139. Currently the hf token is passed as an argument that is visible in kubeflow to all users who have access to it. 

Linking also the issue for properly representing submitted images on the hub (right now they are represented as byte strings: 
https://github.com/huggingface/datasets/issues/5869


examples/pipelines/simple_pipeline/components/write_to_hub/src/main.py
```diff
@@ -0,0 +1,57 @@
+"""
+This component write an images dataset to the hub.
```
RobbeSneyders, line 2:
```suggestion
This component writes an images dataset to the hub.
```

PhilippeMoussalli:
> Thanks @PhilippeMoussalli!
> 
> Now this component isn't really generic though. Would be great if we could automatically define which data to write based on the component spec instead of hardcoded.

I think for this we will need to implement the `DataWriter` class. Don't see how this could be done within the component if we don't have access to the subsets/index columns

RobbeSneyders:
We do have access to the component spec as an attribute on the component class. Agree that we'll want to refactor this into a `DataWriter` class though.

For me it's fine to implement this as a custom component first as well, so it works in the pipeline, and we can do the work to make it reusable later.

examples/pipelines/simple_pipeline/components/write_to_hub/src/main.py
```diff
@@ -0,0 +1,57 @@
...
+            username: The username under which to upload the dataset
+            dataset_name: The name of the dataset to upload
+
+        Returns:
+            dataset
+        """
+        # login
+        huggingface_hub.login(token=hf_token)
+
+        # Create hub
```
NielsRogge, line 38:
```suggestion
        # Create repository
```

examples/pipelines/simple_pipeline/pipeline.py
```diff
@@ -31,9 +31,14 @@
     },
 )
 
-# TODO: ADD Arguments for embedding component later on
-# MODEL_ID = "openai/clip-vit-large-patch14"
-# BATCH_SIZE = 10
+write_to_hub_op = FondantComponentOp(
```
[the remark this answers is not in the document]
PhilippeMoussalli: yeah indeed, gonna change the simple pipeline to the sd one and then i'll connect all the components together

examples/pipelines/simple_pipeline/components/write_to_hub/src/main.py
```diff
@@ -0,0 +1,57 @@
...
+
+configure_logging()
+logger = logging.getLogger(__name__)
+
+
+class WriteToHubComponent(FondantTransformComponent):
+    def transform(
+        self,
+        *,
+        dataframe: dd.DataFrame,
```
NielsRogge, line 20:
```suggestion
        dataframe: dd.DataFrame,
        *,
```
It might be that you'll need to change this to this to run the component successfully

PhilippeMoussalli: I think it should be fine since the FondantTransformComponent is currently defined like this 
  ```
  def transform(self, *args, **kwargs) -> dd.DataFrame
  ```

RobbeSneyders (approved):
Thanks @PhilippeMoussalli!

PR #142, opened 2023-05-16
Add first working for gcp

GeorgesLorre:


deployments/README.md
```diff
@@ -0,0 +1,49 @@
+# 
+
+In order to run fondant properly we'll need a couple of things:
```
RobbeSneyders, line 3:
```suggestion
Fondant is built on top of Kubeflow Pipelines, so to run it we'll need:
```

deployments/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Fondant on GCP
+
+1. If you don't already have a google cloud project ready you can follow this [guide](https://v1-5-branch.kubeflow.org/docs/distributions/gke/deploy/project-setup/) to set one up, you will need to have set up billing.
+
+2. Make sure you have the [gcloud cli](https://cloud.google.com/sdk/docs/install) installed (and it is the latest version) and that you have it configured to use your project by using `gcloud init`.
+
+3. Setup [Default compute Region and Zone](https://cloud.google.com/compute/docs/gcloud-compute#default-region-zone)
+
+3. Install [kubectl](https://kubernetes.io/docs/tasks/tools/)
```
RobbeSneyders, line 27: Could we include this in the Makefile as well?

RobbeSneyders (commented):
Thanks @GeorgesLorre!

Will we use a Makefile for local deployment as well?

deployments/Makefile
```diff
@@ -0,0 +1,35 @@
...
+     --machine-type "e2-standard-2" \
+     --scopes "cloud-platform"
+
+authenticate-gcp-cluster:
+	gcloud container clusters get-credentials ${CLUSTER_NAME} \
+
+install-kfp:
+	kubectl apply -k "github.com/kubeflow/pipelines/manifests/kustomize/cluster-scoped-resources?ref=${KUBEFLOW_PIPELINE_VERSION}"
+	kubectl wait --for condition=established --timeout=60s crd/applications.app.k8s.io
+	kubectl apply -k "github.com/kubeflow/pipelines/manifests/kustomize/env/dev?ref=${KUBEFLOW_PIPELINE_VERSION}"
```
PhilippeMoussalli, line 32: The installation here is using the standard `dev`environment. We might want to switch to the GCP one to be able to store the kfp artifact (manifests in this case) in a persistent storage. There are other advantages as well such as persistent storage of pipeline runs on SQL but we can skip that. 
Kubeflow pipelines also had custom enviroments for different clouds: 
https://github.com/kubeflow/pipelines/tree/sdk/release-1.8/manifests/kustomize/env

Links: 
https://github.com/ml6team/Express-infra/tree/main/kfp-setup
https://www.kubeflow.org/docs/components/pipelines/v1/installation/standalone-deployment/#:~:text=to%20do%20anything.-,Customizing%20Kubeflow%20Pipelines,-Kubeflow%20Pipelines%20can

PhilippeMoussalli (commented):
Thanks @GeorgesLorre :) !
Left a small comment

deployments/README.md
GeorgesLorre: checking if it is installed ? or installing it (or both) ?

The installation is OS and CPU architecture specific so it might be difficult to automate

deployments/Makefile
GeorgesLorre: So I decided against it since you need to setup a SQL database which is not cheap. I did add a link to this so the user can still go for custom deployments if they want to.

docs/infrastructure.md
```diff
@@ -0,0 +1,155 @@
...
+## Introduction
+Fondant is built on top of [Kubeflow Pipelines](https://www.kubeflow.org/docs/components/pipelines/v1/introduction/), so to run it we'll need:
+
+- A kubernetes cluster
+- Kubeflow pipelines installed on the cluster
+- A registry to store custom component images like (docker hub, Github Container Registry, etc)
+
+This can be on any kubernetes cluster, if you don't have access to a setup like this or you feel uncomfortable to setup your own we have provided some basic scripts to get you started on GCP or on a small scale locally.
+
+**IMPORTANT: These script serve just a kickstart to help you start using fondant fast, these are not production ready environments.**
```
RobbeSneyders, line 19: Would be nice to use [Admonitions](https://squidfunk.github.io/mkdocs-material/reference/admonitions/) here.

README.md
```diff
@@ -117,7 +117,9 @@ pip install git+https://github.com/ml6team/fondant.git
 
 ### Deploying Fondant
 
-TODO
+Fondant is flexible and works by leveraging [kubeflow pipelines](https://www.kubeflow.org/docs/components/pipelines/v1/introduction/) on any kubernetes cluster. All fondant needs is an url pointing to the kubeflow pipeline host and a Object Storage provider (S3, GCS, etc) to store data produced in the pipeline between steps.
+
+We have compiled some references and created some scripts to [get you started](docs/infrastructure.md) with setting up the required infrastructure.
```
RobbeSneyders, line 122:
```suggestion
We have compiled some references and created some scripts to [get you started](https://fondant.readthedocs.io/en/latest/manifest/) with setting up the required infrastructure.
```

RobbeSneyders (commented):
Thanks @GeorgesLorre!
Looks great! 2 minor comments.

README.md
GeorgesLorre: I guess we still want to ref infrastructure and not manifest ?

docs/infrastructure.md
GeorgesLorre: Thx! I'll take a look

README.md
RobbeSneyders: Woops, yes :)

RobbeSneyders:
Maybe move the make files to the `scripts` folder

README.md
GeorgesLorre: I would expect it to work based on:
https://www.mkdocs.org/user-guide/writing-your-docs/#linking-to-pages

Unless I'm misunderstanding

GeorgesLorre: Just tested locally with `mkdocs serve` and just referencing the .md file works because mkdocs will convert it to the right url.

RobbeSneyders: Ok, it didn't work on the test build on readthedocs.

RobbeSneyders (approved):
Thanks! If you resolve the conflict, we can merge.

README.md
GeorgesLorre: because I used `docs/infrastructure.md` but it should just be `infrastructure.md`

GeorgesLorre: well I did not think about that. I'll make the change

PR #146, opened 2023-05-17
Allow handling nested list types

PhilippeMoussalli:
PR that adds missing data types required for defining needed nested data types for the embedding component

Changes:
* The string values of the enum types have been changed to pyarrow types to make it easier to define complex schema
* utf8 types defined in the components have been changed to strings to make them more intuitive 

We will need to make more changes in the future to handle different nested data types as suggested by @GeorgesLorre 
https://swagger.io/docs/specification/data-models/data-types/#:~:text=the%20null%20value.-,Arrays,-Arrays%20are%20defined

Enums allow us to define nice constants that are typed but we'll need to define many in order to accommodate for all different types of nested structures. We might need to move away to dynamically typed data types with a dictionary but this will require quite some changes to the json schemas and the code so better leave it for later


fondant/schema.py
```diff
@@ -16,44 +18,76 @@ class Type(enum.Enum):
...
 class Field(t.NamedTuple):
     """Class representing a single field or column in a Fondant subset."""
 
     name: str
     type: Type
+
+
+Field("a", Type.string_list)
+
+a = 2
```
RobbeSneyders, line 93: I guess this should not be in here?

RobbeSneyders (commented):
Thanks @PhilippeMoussalli 

I don't think all of these make sense to add though. If we want to support all possible nested data types, we should build a proper mechanism for it.

Since we just want to unlock usage of vectors, I would just add the one(s) needed to support this.

fondant/schema.py
PhilippeMoussalli: indeed, forgot to remove it

PhilippeMoussalli:
> Thanks @PhilippeMoussalli
> 
> I don't think all of these make sense to add though. If we want to support all possible nested data types, we should build a proper mechanism for it.
> 
> Since we just want to unlock usage of vectors, I would just add the one(s) needed to support this.

That makes sense, I removed the unwanted ones and kept the ones that are needed or might be needed in the near future

RobbeSneyders (approved):
Thanks!

PR #147, opened 2023-05-17
Component/image border crop

ChristiaensBert:
This PR contains the Image Cropping component.

The component looks for the most common color in the border. It uses this color to calculate how much of the image border can be cropped out. If the crop is not square, it will paste a border on the shortest side to make it square again.

d4e35776-3ce1-4157-ac1f-5b2f18ff2ad4
82eeae2d-c63c-42cb-881c-3707971d043c


examples/pipelines/simple_pipeline/components/image_cropping/fondant_component.yaml
```diff
@@ -0,0 +1,30 @@
...
+image: ghcr.io/ml6team/image_cropping:latest
+
+input_subsets:
+  images:
+    fields:
+      data:
+        type: binary
+
+output_subsets:
+  images_crop:
```
RobbeSneyders, line 12: I think we can overwrite the original `images` subset.

examples/pipelines/simple_pipeline/components/image_cropping/src/main.py
```diff
@@ -0,0 +1,78 @@
...
+
+    Args:
+        image_bytes (bytes): input image in bytes
+
+    Returns:
+        np.int16: height of the image
+    """
+    return np.int16(
+        Image.open(io.BytesIO(image_bytes)).size[1]
+    )
```
RobbeSneyders: Can we combine these so we don't need to open the image twice?

We could also expect these as part of the input dataset, but I guess that would make this component less reusable, since then it would only work on datasets where this information is already available.

examples/pipelines/simple_pipeline/components/image_cropping/fondant_component.yaml
```diff
@@ -0,0 +1,30 @@
...
+      data:
+        type: binary
+      width:
+        type: int16
+      height:
+        type: int16
+
+args:
+  scale:
+    description: Scale parameter used for detecting borders
```
RobbeSneyders, line 23: These arguments are not completely clear to me. Can we explain them in a bit more detail? Either here or in the readme.

RobbeSneyders (commented):
Thanks @ChristiaensBert!

Could you add a small README as well? Including the images from your PR description.

We should do this for every reusable component :)

examples/pipelines/simple_pipeline/components/image_cropping/fondant_component.yaml
ChristiaensBert: done!

examples/pipelines/simple_pipeline/components/image_cropping/fondant_component.yaml
ChristiaensBert: I removed the first argument since it's not that important and added some information about the offset

examples/pipelines/simple_pipeline/components/image_cropping/fondant_component.yaml
```diff
@@ -0,0 +1,27 @@
...
+        type: int16
+      height:
+        type: int16
+
+args:
+  cropping_threshold:
+    description: Threshold parameter used for detecting borders. A lower (negative) parameter results in a more performant border detection, but can cause overcropping. Default is -30
+    type: int
+  padding:
+    description: Padding for the image cropping
```
PhilippeMoussalli, line 26: Could you also clarify how padding changes the algorithm's behaviour in the description?

examples/pipelines/simple_pipeline/components/image_cropping/src/image_crop.py
```diff
@@ -0,0 +1,106 @@
...
+            )
+        else:
+            padding = int((image_crop.size[1] - image_crop.size[0]) / 2)
+            image_crop = ImageOps.expand(
+                image_crop, border=(padding, 0), fill=color_common
+            )
+
+    # serialize image to JPEG
+    crop_bytes = io.BytesIO()
+    image_crop.save(crop_bytes, format="JPEG")
```
PhilippeMoussalli, line 104: probably best to omit the format to keep it as the default one. Perhaps user start with a png image and want to keep it that way

examples/pipelines/simple_pipeline/components/image_cropping/src/main.py
```diff
@@ -0,0 +1,74 @@
...
+            lambda x: remove_borders(x, cropping_threshold, padding),
+            meta=("images_data", "bytes"),
+        )
+
+        # extract width and height
+        dataframe[["images_width", "images_height"]] = dataframe[
+            [
+                "images_data",
+            ]
+        ].apply(extract_dimensions, axis=1, result_type="expand", meta={0: int, 1: int})
```
PhilippeMoussalli, line 67: can we define the meta return types explicitly and not by index? 
```
meta={"images_width": int, "images_height": int}
```

PhilippeMoussalli (approved):
Thanks @ChristiaensBert :) 

Left a few small comments but should be good to ship

.gitignore
```diff
@@ -81,6 +81,10 @@ MANIFEST
 
 # Poetry
 *.lock
+development/
 
 # kubeflow artifacts
-pipeline.tgz
\ No newline at end of file
+pipeline.tgz
```
NielsRogge, line 87: The `gitignore` on main was updated to contain *.tgz so this can be removed

examples/pipelines/simple_pipeline/components/image_cropping/src/main.py
ChristiaensBert: I think I tried it but it didn't work. But by having the "images_width" and height in the dataframe it gives the same results. It's weird

RobbeSneyders (approved):
Thanks Bert!
If you resolve the merge conflict, you can merge.

PR #152, opened 2023-05-22
Add embedding based LAION retrieval component

ChristiaensBert:
This branch is based on the image-embedding branch which has a lot of changes. I would suggest to first merge that PR which will give a much smaller Diff here.

This component implements the LAION image retrieval component which uses CLIP embeddings from the input subset to query the LAION database. It returns an images subset with URLs, similar to the other prompt based Clip Retrieval component. These URLs should then be downloaded by the already-made image-downloading component.


RobbeSneyders:
Can you rebase this branch @ChristiaensBert?

PhilippeMoussalli (approved):
Rebased it since I'm a bit more familiar with what changed are needed. 
LGTM overall

PR #155, opened 2023-05-22
Remove id and source column, optimize data writing & components

RobbeSneyders:
This PR implements a couple of optimizations. Probably too many at once. We can split them if we want to take the time for this.

- Refactor the `DataWriter` class to only call `compute()` once.
- Remove the `id` and `source` column and let the user work with the dataframe index directly. The separate columns were giving too many issues without benefits. It was not clear when the `uid` column should be rebuild and reset, which led to every component both resetting and rebuilding it.
- Make querying the clip client concurrent. This was taking a long time before, since each request was handled sequentially. We should do the same for the image downloading component.
- Use LAION ids.
- Pass the ids in the captioning component so the resulting captions are linked to the original ids. I'm not sure that the captions were being linked to the correct images in the previous implementation. Not too happy with my current implementation though, this should be cleaned up.


components/prompt_based_laion_retrieval/src/main.py
```diff
@@ -70,33 +82,20 @@ def transform(
...
 
-        print("Input dataframe:", dataframe.head(2))
+        meta_df = pd.DataFrame(columns=["id", "url"], dtype="string")
+        meta_df = meta_df.set_index("id")
 
-        logger.info("Retrieving URLs...")
-        dataframe["images_url"] = dataframe["prompts_text"].apply(
-            lambda example: query_clip_client(example, client),
-            meta=("images_url", "str"),
+        dataframe = dataframe.map_partitions(
```
ChristiaensBert, line 88: Maybe it's worth it calculating the cost of deploying LAION ourself in an endpoint somewhere. If the cost is reasonable it would avoid the timeouts of the LAION-hosted dataset

RobbeSneyders: Yes, or if we could help host the LAION one. Something to look into.

BTW, this PR is outdated.
We already split these parts into separate PRs:

- Refactor the DataWriter class to only call compute() once.
- Remove the id and source column and let the user work with the dataframe index directly. The separate columns were giving too many issues without benefits. It was not clear when the uid column should be rebuild and reset, which led to every component both resetting and rebuilding it.

Still need to do the rest.

RobbeSneyders:
Split into #157 #159 #168

PR #156, opened 2023-05-22
Move reusable components to registry

RobbeSneyders:


components/image_cropping/src/main.py
```diff
@@ -17,7 +18,7 @@
 logger = logging.getLogger(__name__)
 
 
-def extract_dimensions(image_df: dd.DataFrame) -> [np.int16, np.int16]:
+def extract_dimensions(image_df: dd.DataFrame) -> t.Union[np.int16, np.int16]:
```
PhilippeMoussalli, line 21: Doesn't `union` suggest that it's either one or the other?

Suggestion: 
```
def extract_dimensions(image_df: dd.DataFrame) -> t.Tuple[np.int16, np.int16]:
    """Extract the width and height of an image

    Args:
        image_df (dd.DataFrame): input dataframe with images_data column

    Returns:
       Tuple containing the width and height of the images
    """
    image = Image.open(io.BytesIO(image_df["images_data"]))

    return np.int16(image.size[0]), np.int16(image.size[1])
```

examples/pipelines/finetune_stable_diffusion/pipeline.py
```diff
@@ -18,43 +18,55 @@
...
 pipeline = Pipeline(pipeline_name=pipeline_name, base_path=PipelineConfigs.BASE_PATH)
 
 pipeline.add_op(load_from_hub_op)
-pipeline.add_op(image_filtering_op, dependencies=load_from_hub_op)
 pipeline.add_op(image_embedding_op, dependencies=load_from_hub_op)
-# pipeline.add_op(image_filtering_op, dependencies=load_from_hub_op)
-pipeline.add_op(write_to_hub_op, dependencies=image_embedding_op)
+pipeline.add_op(laion_retrieval_op, dependencies=image_embedding_op)
+pipeline.add_op(download_images_op, dependencies=laion_retrieval_op)
+pipeline.add_op(caption_images_op, dependencies=download_images_op)
```
PhilippeMoussalli, line 70: I can add the write to hub in a later PR. Any reason for removing the `image_filtering`?

PhilippeMoussalli (approved):
Thanks Robbe for the hard work! :)

components/image_cropping/src/main.py
RobbeSneyders: Yes, wasn't paying attention apparently :sweat_smile:

examples/pipelines/finetune_stable_diffusion/pipeline.py
RobbeSneyders: Yes, I removed the write to hub since it still needs to be duplicated as discussed.

I removed the image filtering, because this is already part of the downloading component, which I added together with the LAION retrieval component. Unless I missed a reason that we still need it?

I haven't tested this pipeline yet with these steps btw.

components/image_cropping/src/main.py
PhilippeMoussalli: You were doing a lot of heavy lifting yesterday so you're good :P

examples/pipelines/finetune_stable_diffusion/pipeline.py
PhilippeMoussalli: No in this case we don't need a separate component for it. I can test everything once we merge this PR and the one containing the dask tasks optimizations

NielsRogge (approved):
Thanks, have you checked that the alpha pipeline still runs fine?

RobbeSneyders:
Yes :rocket:

PR #159, opened 2023-05-23
Drop id and source columns

RobbeSneyders:
Remove the `id` and `source` column and let the user work with the dataframe index directly. The separate columns were giving too many issues without benefits. It was not clear when the `uid` column should be rebuild and reset, which led to every component both resetting and rebuilding it.


PhilippeMoussalli (approved):

fondant/data_io.py
```diff
@@ -91,16 +91,10 @@ class DaskDataWriter(DataIO):
...
-        # set index
-        if dataframe.index.name != "uid":
-            dataframe["uid"] = dataframe["source"] + "_" + dataframe["id"]
-            dataframe = dataframe.set_index("uid", divisions=divisions)
-
-        # load index dataframe
-        index_columns = list(self.manifest.index.fields.keys())
-        index_df = dataframe[index_columns]
+        # Turn index into an empty dataframe so we can write it
+        index_df = dataframe.index.to_frame().drop(columns=["id"])
```
GeorgesLorre, line 97: why are we dropping the id column here ? It does not exist anymore ?

components/prompt_based_laion_retrieval/src/main.py
```diff
@@ -78,17 +78,6 @@ def transform(
...
-        # TODO use LAION id instead
-        dataframe["id"] = dataframe.assign(id=1).id.cumsum()
-        dataframe["source"] = "laion"
-
-        # reorder columns
-        dataframe = dataframe[["id", "source", "images_url"]]
-
-        dataframe = dataframe.astype({'id': 'string', 'source': 'string'})
-
         dataframe = dataframe.reset_index(drop=True)
```
GeorgesLorre, line 81: This creates the default index, could we not do it dynamically when adding rows ?

fondant/data_io.py
```diff
@@ -91,16 +91,10 @@ class DaskDataWriter(DataIO):
     def write_dataframe(self, dataframe: dd.DataFrame) -> None:
         write_tasks = []
 
-        divisions = list(dataframe.divisions) if dataframe.known_divisions else None
+        dataframe.index = dataframe.index.rename("id").astype("string")
```
GeorgesLorre, line 94: So we expect that every dataframe coming from a component has an index set.

PR #167, opened 2023-05-23
add multi modal dataset example

ChristiaensBert:
Small PR to add a good example image of a multi-modal dataset to the Pipeline documentation


RobbeSneyders (approved):
Thanks Bert!

PR #168, opened 2023-05-24
Make LAION retrieval concurrent and use LAION ids

RobbeSneyders:
This PR tackles 2 changes:

1. It makes the LAION retrieval components concurrent, which speeds them up significantly. To really use this at scale, we should set up our own backend though.

As part of this change, I also included the LAION ids instead of a range index, which leads to the second change:

2. Update the embedding, caption, and segmentation components to track the ids correctly. Previously a new range index was created at the end, which only coincidentally matched the input index. To track the ids, I refactored these components to leverage the dataframe instead of delayed API from Dask.

All changes are completely contained in there component, so the components can be reviewed individually.


components/caption_images/src/main.py
```diff
@@ -22,37 +19,67 @@
...
+    images: pd.Series,
+    *,
+    model: BlipForConditionalGeneration,
+    processor: BlipProcessor,
+    batch_size: int,
+    max_new_tokens: int,
+    device: str,
+):
+    """Caption a pandas series of images"""
+    images = images.apply(process_image, processor=processor, device=device)
```
NielsRogge, line 70: So we're not leveraging parallellism for prepocessing the images?

As this seems to be just Pandas

components/segment_images/src/main.py
```diff
@@ -109,32 +125,22 @@ def transform(
         device = "cuda" if torch.cuda.is_available() else "cpu"
         logger.info(f"Device: {device}")
 
-        processor = AutoImageProcessor.from_pretrained(model_id)
+        processor = SegformerImageProcessor.from_pretrained(model_id)
```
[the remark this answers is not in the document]
RobbeSneyders: Based on [this](https://github.com/huggingface/transformers/blob/28aa438cd2759cfdf1da9326ea6cde4496d1be23/src/transformers/models/auto/image_processing_auto.py#L97), I think UperNet does use the `SegformerImageProcessor`? But I'm not as familiar with this as you :sweat_smile: 

Using the `AutoImageProcessor` is nice in theory, but since the different ImageProcessors don't implement the same interface, I actually think this component only supports models using the `SegformerImageProcessor` anyway? Also for typing, it helps to know which `Processor` will be selected.

components/caption_images/src/main.py
```diff
@@ -22,37 +19,67 @@
...
+
+def caption_images(
+    images: pd.Series,
+    *,
+    model: BlipForConditionalGeneration,
+    processor: BlipProcessor,
+    batch_size: int,
+    max_new_tokens: int,
+    device: str,
+):
```
PhilippeMoussalli, line 68: Could you add a return type hint?

components/embedding_based_laion_retrieval/src/main.py
```diff
@@ -17,23 +18,35 @@
...
     """
-    try:
-        results = client.query(embedding_input=embedding_input.tolist())
-        urls = [i["url"] for i in results]
-    except ConnectionError as e:
-        urls = ["" for _ in range(len(results))]
 
-    return urls
+    async def async_query():
+        with concurrent.futures.ThreadPoolExecutor(max_workers=20) as executor:
```
PhilippeMoussalli, line 32: should we make the `max_workers` a component argument?

components/caption_images/src/main.py
```diff
@@ -80,35 +107,21 @@ def transform(
...
+        dataframe = dataframe["images_data"].map_partitions(
+            caption_images,
+            model=model,
+            processor=processor,
+            batch_size=batch_size,
+            max_new_tokens=max_new_tokens,
+            device=device,
+            meta={0: str}
+        )
+        dataframe.columns = ["captions_text"]
```
PhilippeMoussalli, line 122: Could you follow a similar for defining the `meta` dataframe as you did in the the laion component
```
meta_df = pd.DataFrame(columns=["id", "captions_text"], dtype="string")
meta_df = meta_df.set_index("id")

        dataframe = dataframe["images_data"].map_partitions(
            caption_images,
            model=model,
            processor=processor,
            batch_size=batch_size,
            max_new_tokens=max_new_tokens,
            device=device,
            meta=meta_df
        )
```
Haven't tested this but I think it makes it a bit more explicit what dataframe you're returning.

components/prompt_based_laion_retrieval/src/main.py
```diff
@@ -17,25 +18,35 @@
...
     """
-    results = client.query(text=text)
-    try:
-        urls = [i["url"] for i in results]
-    except ConnectionError as e:
-        urls = ["" for _ in range(len(results))]
 
-    return urls
+    async def async_query():
+        with concurrent.futures.ThreadPoolExecutor(max_workers=20) as executor:
```
PhilippeMoussalli, line 32: Same comment as above

components/segment_images/src/main.py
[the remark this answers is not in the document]
PhilippeMoussalli: Indeed would be nice to have a generic component that can leverage any HF segmentation model and users can change the model based on a model ID. Makes them more re-usable. But if the models have different interfaces that won't be possible

components/segment_images/src/main.py
```diff
@@ -109,32 +125,22 @@ def transform(
...
-        segmentations_df = delayed_series.to_frame(name="segmentations_data")
-
-        return segmentations_df
+        dataframe = dataframe["images_data"].map_partitions(
+            segment_images,
+            model=model,
+            processor=processor,
+            batch_size=batch_size,
+            device=device,
+            meta={0: object}
```
PhilippeMoussalli, line 137: Same comment here as above regarding explicitly defining the dataframe

components/segment_images/src/main.py
```diff
@@ -109,32 +125,22 @@ def transform(
...
+        )
+        dataframe.columns = ["segmentations_data"]
+
+        print(dataframe.head())
```
PhilippeMoussalli, line 141: This can be removed

PhilippeMoussalli (approved):
Thanks Robbe! This will improve the speed quite a bit and the interface looks cleaner as well with dataframes. 

Left a few comments

components/caption_images/src/main.py
RobbeSneyders: I'm not doing this here since I didn't want to set the column names inside the `caption_images`. I kept those names on the "component level", since that's where they are specified in the specification as well.

components/embedding_based_laion_retrieval/src/main.py
RobbeSneyders: Sounds like a good idea indeed, but we currently don't have a default mechanism for the arguments. I would add that first.

PR #172, opened 2023-05-25
Update discord badge to show logo

RobbeSneyders:
Received some feedback that the Discord should be more prominent. I think this does the trick :)


PhilippeMoussalli (approved):
Thanks :)

PR #175, opened 2023-05-31
Add Starcoder example pipeline + base components

NielsRogge:
This PR:

- adds a set of components aimed to process text/code datasets to the component registry. These components can be used to 1) filter code based on code to comments ratio 2) filter code based on line length 3) detect and redact (replace) PII or personal identifiable information from code.
- adds an example pipeline that adds one specific component called `load_from_hub_stack` that loads a code dataset from the hub. Next, it uses the 3 components in sequence to process and filter the code dataset.


components/pii_redaction/src/main.py
```diff
@@ -0,0 +1,72 @@
+"""A component that detects and redacts Personal Identifiable Information (PII) in code."""
```
ChristiaensBert, line 1: This docstring is styled different than in `filter_metadata/src/main.py`. Which one should we use?

components/pii_redaction/src/pii_detection.py
```diff
@@ -0,0 +1,42 @@
+import json
```
ChristiaensBert, line 1: add docstring

components/pii_redaction/src/main.py
```diff
@@ -0,0 +1,74 @@
...
+
+import json
+import logging
+
+import dask.dataframe as dd
+
+from fondant.component import TransformComponent
+from fondant.logger import configure_logging
+
+from pii_detection import scan_pii
```
PhilippeMoussalli, line 13:
```suggestion
from pii_detection import scan_pii, redact_pii
```

components/pii_redaction/src/pii_detection.py
PhilippeMoussalli, line 1: could you add typehints and module description?

examples/pipelines/starcoder/components/filter_comments
```diff
@@ -0,0 +1 @@
+/Users/nielsrogge/Documents/python_projects/fondant/components/filter_comments
```
PhilippeMoussalli, line 1: can be removed

examples/pipelines/starcoder/components/filter_line_length
```diff
@@ -0,0 +1 @@
+/Users/nielsrogge/Documents/python_projects/fondant/components/filter_line_length
```
PhilippeMoussalli, line 1: can be removed

PhilippeMoussalli (commented):
Thanks Niels! 

Left a few comments, I think there are a few things still GCP specific that need to be removed. 

I;m guessing this code and components will only work for our personally created ML6 version of the stack dataset (loading it from json and dumping it to a parquet under a new dataset repo). What if the user wants to work with the big stack datasets? Can we generalize well with those components?

components/pii_redaction/src/pii_redaction.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+def random_replacements(n=10):
+    """Build dictionaries of random replacements for PII (key, email, IP address)
+
+    Emails: replace with one of n [random string of 5 characters + @example.com]
+    IP addresses: replace with one of n synthetic private IP addresses (IPv4 or IPv6)
+    Keys: replace with one of n [sequence of 32 random characters/digits]
+
+    TODO: add IPv6 and IPv4 separation
```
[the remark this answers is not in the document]
NielsRogge: This was taken from the BigCode project

components/pii_redaction/src/main.py
```diff
@@ -0,0 +1,74 @@
...
+        # detect PII
+        result = dataframe.apply(
+            lambda example: scan_pii(text=example.code_content),
+            axis=1,
+            result_type="expand",
+            meta={0: object, 1: bool, 2: int},
+        )
+        result.columns = ["code_secrets", "code_has_secrets", "code_number_secrets"]
+
+        dataframe = dataframe.merge(result, left_index=True, right_index=True)
```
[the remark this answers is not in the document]
NielsRogge: We are merging since Dask does not support multiple columns assignment

One of the many reasons to stay away from Dask :D

examples/pipelines/starcoder/components/filter_comments
NielsRogge: I needed the symbolic links in order to run the pipeline locally

components/pii_redaction/src/main.py
PhilippeMoussalli: just a matter of personal preference but I think both are valid :)

components/pii_redaction/src/pii_redaction.py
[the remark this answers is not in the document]
PhilippeMoussalli: maybe not for this project but it would be interesting to have a way of working on updating components like this where the source code is taken from another repo. Right now it's just a snapshot

examples/pipelines/starcoder/components/filter_comments
RobbeSneyders: Only with the script right. I would not include this in git. The local runner will solve this.

examples/pipelines/starcoder/run_locally.sh
RobbeSneyders, line 1: Let's not include this, as the local runner will replace this.

components/pii_redaction/src/main.py
```diff
@@ -0,0 +1,74 @@
...
+                replacements=replacements,
+            ),
+            axis=1,
+            meta=(None, "str"),
+        )
+        dataframe = dataframe.drop(
+            ["code_secrets", "code_has_secrets", "code_number_secrets"], axis=1
+        )
+
+        print("Final dataframe:", dataframe.head(2))
```
RobbeSneyders, line 67: Should be removed

components/pii_redaction/src/main.py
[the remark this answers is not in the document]
RobbeSneyders: Yes, in pandas this works.

components/pii_redaction/src/pii_redaction.py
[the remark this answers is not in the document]
RobbeSneyders: Would be ideal if this was a library.

examples/pipelines/starcoder/README.md
RobbeSneyders, line 1: Can you add some info here?

RobbeSneyders (commented):
Thanks @NielsRogge! Left some comments

In general:
- Can we add a `code_` prefix to the names of these components? Or would they work out of the box on any text? I guess at least the `filter_comments` component wouldnt'.
- I would remove anything related to the `run_locally` script and symlinks. This will be replaced by the local runner.
- This PR has some conflicts with #191. We'll need to make sure everything is updated after merging both.

RobbeSneyders:
And can you also update the readme to include these components in the list?

RobbeSneyders:
#191 is merged, so we should make the necessary updates here.

NielsRogge:
@RobbeSneyders I've made the updates

components/pii_redaction/src/gibberish_data/big.txt
RobbeSneyders, line 1: Can we reference this file online somewhere instead? It's quite big (6.2MB) to add to the repo.

RobbeSneyders (commented):
Thanks @NielsRogge.

What do you think of my previous proposal:
> Can we add a code_ prefix to the names of these components? Or would they work out of the box on any text? I guess at least the filter_comments component wouldnt'.

And left one more comment.

components/pii_redaction/src/gibberish_data/big.txt
NielsRogge: Turns out this file is not needed, as it's only used to train a model. Have removed this

NielsRogge:
> Can we add a code_ prefix to the names of these components? Or would they work out of the box on any text? I guess at least the filter_comments component wouldnt'.

The pii redaction component would work on text as well, but the filter comments and filter metadata components are meant to be run on code.

PhilippeMoussalli (approved):

PR #176, opened 2023-05-31
Component/controlnet conditioning components

ChristiaensBert:
I made this component class that supports a bunch of conditioning strategies such as 

```
["hed", "midas", "mlsd", "openpose", "pidinet", "normalbae", "lineart", "lineart_coarse", "lineart_anime", "canny", "content_shuffle", "zoe", "mediapipe_face"]
```
from the controlnet_aux library.

The library already implemented these processing models/classes and is frequently used (was built for) for ControlNet conditioning.
I added a PR on the controlnet_aux repo to add a Processor class, that abstracts all these models and makes it easy to use by just passing a `processor_id`.

My first PR already got merged but I made a second one that adds a bit more features and improvements
https://github.com/patrickvonplaten/controlnet_aux/pull/50. The PR here is very simple, because I moved most of the logic to the Processor class within the `controlnet_aux` repo.

You can review it but we can wait to merge it until the above PR is merged on the controlnet_aux branch


components/controlnet_aux_processor/fondant_component.yaml
```diff
@@ -0,0 +1,20 @@
...
+image: ghcr.io/ml6team/controlnet_aux_processor:latest
+
+consumes:
+  images:
+    fields:
+      data:
+        type: binary
+
+produces:
+  conditioning:
```
PhilippeMoussalli, line 12: are there potential use cases where we might need to have more than one conditioning in the dataset (e.g. segmentation + canny, ...)? Or is that unlikely?

components/controlnet_aux_processor/requirements.txt
```diff
@@ -0,0 +1,7 @@
+git+https://github.com/ml6team/fondant.git@main
+gcsfs==2023.4.0
+transformers==4.29.2
+Pillow==9.4.0
+torch==2.0.1
+controlnet_aux==0.0.5
+mediapipe
```
PhilippeMoussalli, line 7: can you fix this version?

components/controlnet_aux_processor/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM --platform=linux/amd64 pytorch/pytorch:2.0.1-cuda11.7-cudnn8-devel
```
PhilippeMoussalli, line 1: I think you can change the base image to the `runtime` one instead of the `devel` one since it's more lightweight and should do the job -> `pytorch/pytorch:2.0.1-cuda11.7-cudnn8-runtime`

I don't think you need all the additional stuff that come with the `devel` version. 

https://stackoverflow.com/questions/56405159/what-is-the-difference-between-devel-and-runtime-tag-for-a-docker-container

docs/art/components/controlnet_aux/output_content_shuffle.jpg
PhilippeMoussalli, line 1: what happened here :P

PhilippeMoussalli (approved):
Nice one! :) Thanks for adding. 

Left a few comments but should be good to go

components/controlnet_aux_processor/fondant_component.yaml
ChristiaensBert: Could happen, but I suppose the user just has to rewrite the component so it does multiple at once and writes them to different columns, or that it uses 2 separate components based on this one.

docs/art/components/controlnet_aux/output_content_shuffle.jpg
ChristiaensBert: Hahah, it's a content shuffle. Bit messed up though

components/controlnet_aux_processor/Dockerfile
ChristiaensBert: > Nice one! :) Thanks for adding.
> 
> Left a few comments but should be good to go

I will still wait for the PR on the controlnet_aux repo to be merged, it's still in review :D Then I can test/merge it!

ChristiaensBert:
I'll reopen it when the controlnet_aux library is fixed

PR #177, opened 2023-06-01
hardcode images links

PhilippeMoussalli:
Harcode image links to render properly on pypi


RobbeSneyders (commented):
Thanks @PhilippeMoussalli!

I think we can now also remove the `docs/docs/art` symlink, which I added to make the links in the README work both on Github and on our documentation.

RobbeSneyders (approved):

Issue #178, opened 2023-06-01
Allow handling of nested data types


PhilippeMoussalli:
### Problem Statement

Fondant currently supports handling pyarrow data types as specified in the [schema](https://github.com/ml6team/fondant/blob/main/fondant/schema.py). However, sometimes there is a need to specify nested data types as a list/array. We have a [quick implementation](https://github.com/ml6team/fondant/blob/43567274fc00c87c96011ec0dc3c18931d54e10a/fondant/schema.py#LL54C1-L55C1) on how this is done but it's not a permanent solution since it is not nicely defined, requires a lot of replication and is only limited to nested function of 1 layer. The solution should ideally be able to handle any of the pyarrow types defined in the schema while also allow for creating nested data types that can span multiple layers deep. 

The [swagger array data type](https://swagger.io/docs/specification/data-models/data-types/#:~:text=the%20null%20value.-,Arrays,-Arrays%20are%20defined) can serve as an inspiration for this solutions where nested data types are defined as follows: 

One layer
```yaml
type: array
items:
  type: string
```

Two layers
```yaml
type: array
items:
  type: array
  items:
    type: integer
```

Currently, fondant implements a similar schema but the `array` type is not supported (only pyarrow types):

```yaml
consumes:
  images:
    fields:
      url:
        type: string
```

We need to change the schema to accept nested types similar to how Swagger does it: 

```yaml
consumes:
  embeddings:
    fields:
      url:
        type: array
        items:
          type: integer
```

### Implementation Steps/Tasks

- [x] Change the `common.json` [file](https://github.com/ml6team/fondant/blob/main/fondant/schemas/common.json) to allow fields of handling nested data types. 
- [x] Change the `Field` class in [`schema.py`](https://github.com/ml6team/fondant/blob/main/fondant/schema.py) where `type` attribute should be able to handle different nested types. 
We also might want to adjust the [`Type`](https://github.com/ml6team/fondant/blob/43567274fc00c87c96011ec0dc3c18931d54e10a/fondant/schema.py#L13) class and create a nested data type constructor method for a `list` or `array`:
```python
from schema import Type
Type.list[Type.string] # pa.list_(pa.int8())
```
- [x] Possible required changes to the `Manifest` and `ComponentSpec`
- [x] Make necessary change to affected components (embedding, segmentation)
- [x] Remove `int8_list` and `float32_list` from `schema.py`

### Potential Impact

Users will have to be more careful when defining data types and it has to be properly documented

### Testing

[...]

PR #193, opened 2023-06-08
Enable defining nested data types

PhilippeMoussalli:
PR that addresses #178 
Inspiration: https://swagger.io/docs/specification/data-models/data-types/#:~:text=the%20null%20value.-,Arrays,-Arrays%20are%20defined (credit to @GeorgesLorre)

It includes: 
* Changing the common json schema to enable defining array types 
* Changing the `Type` class to be dynamic rather than typed to enable defining nested structures without explicitly defining them
* Adding an addition test file for `schema.py`
* Updating the current existing examples (embedding + segmentation)

### DIFF
--- a/components/embedding_based_laion_retrieval/fondant_component.yaml
+++ b/components/embedding_based_laion_retrieval/fondant_component.yaml
@@ -6,7 +6,9 @@ consumes:
   embeddings:
     fields:
       data:
-        type: float32_list
+        type: array
+        items:
+          type: float32
 
 produces:
   images:
--- a/components/segment_images/fondant_component.yaml
+++ b/components/segment_images/fondant_component.yaml
@@ -12,7 +12,9 @@ produces:
   segmentations:
     fields:
       data:
-        type: binary
+        type: array
+        items:
+          type: binary
 
 args:
   model_id:
--- a/fondant/schema.py
+++ b/fondant/schema.py
@@ -53,9 +53,26 @@ class Type(enum.Enum):
 
     float32_list = pa.list_(pa.float32())
 
+    @classmethod
+    def list(cls, data_type: "Type") -> pa.DataType:
+        """
+        Create a list type from the given data type. Allows creating nested values
+        Args:
+            data_type: The data type of the elements in the list.
+
+        Returns:
+            The list type.
+        """
+        if isinstance(data_type, Type):
+            return pa.list_(data_type.value)
+        elif isinstance(data_type, pa.DataType):
+            return pa.list_(data_type)
+        else:
+            raise ValueError("Invalid data type")
+
 
 class Field(t.NamedTuple):
     """Class representing a single field or column in a Fondant subset."""
 
     name: str
-    type: Type
+    type: t.Union[Type, pa.DataType]
--- a/fondant/schema.py
+++ b/fondant/schema.py
@@ -2,77 +2,120 @@
 and pipelines.
 """
 
-import enum
 import typing as t
 
 import pyarrow as pa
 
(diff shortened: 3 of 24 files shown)


components/segment_images/fondant_component.yaml
```diff
@@ -12,7 +12,9 @@ produces:
   segmentations:
     fields:
       data:
-        type: binary
```
NielsRogge, line 15: Same question here

components/embedding_based_laion_retrieval/fondant_component.yaml
```diff
@@ -6,7 +6,9 @@ consumes:
   embeddings:
     fields:
       data:
-        type: float32_list
+        type: array
```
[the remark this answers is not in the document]
PhilippeMoussalli: it's represented as a list/array of floats 

`pa._list(pa.float32()) == List[float] ==  {"type": "array", "items": {"type": "float32"}}`
(1st is pyarrow, 2nd is python data type, third is json schema)
https://swagger.io/docs/specification/data-models/data-types/#:~:text=the%20null%20value.-,Arrays,-Arrays%20are%20defined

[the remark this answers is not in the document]
NielsRogge: Ok makes sense!

fondant/schema.py
```diff
@@ -2,56 +2,148 @@
...
+        """
+        Validates the provided data type and returns the corresponding data type object.
+
+        Args:
+            data_type: The data type to validate.
+
+        Returns:
+            The validated `pa.DataType` object.
+        """
+        if isinstance(data_type, str):
```
RobbeSneyders, line 67:
```suggestion
        if not isinstance(data_type, Type):
```

I think this is a bit more robust. What if I pass in something that is not a `Type` nor a string? :)

fondant/schema.py
```diff
@@ -2,56 +2,148 @@
...
+        Args:
+            json_schema: The dictionary representation of the data type, can represent nested values
+
+        Returns:
+            A new `Type` instance representing the specified data type.
+        """
+        if json_schema["type"] == "array":
+            items = json_schema["items"]
+            if isinstance(items, dict):
+                return cls.list(Type.from_json(items))
```
RobbeSneyders, line 109:
```suggestion
                return cls.list(cls.from_json(items))
```

fondant/schema.py
```diff
@@ -2,56 +2,148 @@
...
+        else:
+            return Type(json_schema["type"])
```
RobbeSneyders, line 111:
```suggestion
            return cls(json_schema["type"])
```

tests/test_manifest.py
```diff
@@ -60,16 +60,16 @@ def test_subset_fields():
     subset = Subset(specification=subset_spec, base_path="/tmp")
 
     # add a field
-    subset.add_field(name="data2", type_=Type.binary)
+    subset.add_field(name="data2", type_=Type("binary"))
```
RobbeSneyders, line 63: It's a bit unfortunate that we lose the ability to reference types by attribute (and therefore autocomplete, static analysis, etc., ...). Unfortunately I don't see a straightforward way to keep them either.

RobbeSneyders (commented):
Thanks @PhilippeMoussalli, left some small comments.

tests/test_manifest.py
PhilippeMoussalli: yeah indeed, you would need to know/type all possible combinations which is not very feasible. But it's not user-facing so it should not have that big of an impact.

RobbeSneyders (approved):

PR #194, opened 2023-06-09
First implementation of DockerCompiler

GeorgesLorre:
Known TODO's:

- support for build in docker compose
- update documentation
- update examples to include dockercompiler
- update pipeline.py and move compile functionality to a KubeflowCompiler

### DIFF
--- a/fondant/compiler.py
+++ b/fondant/compiler.py
@@ -1,0 +1,103 @@
+import json
+import logging
+import typing as t
+from abc import ABC, abstractmethod
+from pathlib import Path
+
+import yaml
+
+from fondant.pipeline import ComponentOp, Pipeline
+
+logger = logging.getLogger(__name__)
+
+
+class Compiler(ABC):
+    """Abstract base class for a compiler."""
+
+    def __init__(self, pipeline: Pipeline):
+        self.pipeline = pipeline
+
+    @abstractmethod
+    def compile(self):
+        """Abstract method to invoke compilation."""
+
+
+class DockerCompiler(Compiler):
+    """Compiler that creates a docker-compose spec from a pipeline."""
+
+    def compile(self, package_path: str = "docker-compose.yml") -> None:
+        """Compile a pipeline to docker-compose spec that is saved on the package_path."""
+        self._patch_path()
+        spec = self._generate_spec()
+        with open(package_path, "w") as outfile:
+            yaml.safe_dump(spec, outfile)
+
+    @staticmethod
+    def _safe_component_name(component_name: str) -> str:
+        """Transform a component name to a docker-compose friendly one.
+        eg: `Component A` -> `component_a`.
+        """
+        return component_name.replace(" ", "_").lower()
+
+    def _patch_path(self):
+        """Helper that checks if the base_path is local or remote,
+        if local it patches the base_path and prepares a bind mount.
+        """
+        base_path = Path(self.pipeline.base_path)
+        # check if base path is an existing local folder
+        if base_path.exists():
+            self.volume = {
+                "type": "bind",
+                "source": str(base_path),
+                "target": f"/{base_path.stem}",
+            }
+            self.path = f"/{base_path.stem}"
+        else:
+            self.volume = None
+            self.path = self.pipeline.base_path
+
+    def _generate_spec(self) -> dict:
+        """Generate a docker-compose spec as a python dictionary,
+        loops over the pipeline graph to create services and their dependencies.
+        """
+        services = {}
+        for component_name, component in self.pipeline._graph.items():
+            safe_component_name = self._safe_component_name(component_name)
+            services[safe_component_name] = self.compose_service(
+                component["fondant_component_op"], component["dependencies"]
+            )
+        return {"version": "3.8", "services": services}
+
+    def compose_service(
+        self, component_op: ComponentOp, dependencies: t.List[str]
+    ) -> dict:
+        """Take in a component and create a docker-compose service based on the properties."""
+        # add metadata argument to command
+        metadata = {"run_id": self.pipeline.name, "base_path": self.path}
+        command = ["--metadata", json.dumps(metadata)]
+
+        # add in and out manifest paths to command
+        command.extend(["--output_manifest_path", f"{self.path}/manifest.txt"])
+
+        # add arguments if any to command
+        for key, value in component_op.arguments.items():
+            command.extend([f"--{key}", f"{value}"])
+
+        # resolve dependencies
+        depends_on = {}
+        if dependencies:
+            # there is only an input manifest if the component has dependencies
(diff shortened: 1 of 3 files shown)


fondant/compiler.py
```diff
@@ -0,0 +1,103 @@
...
+    @abstractmethod
+    def compile(self):
+        """Abstract method to invoke compilation."""
+
+
+class DockerCompiler(Compiler):
+    """Compiler that creates a docker-compose spec from a pipeline."""
+
+    def compile(self, package_path: str = "docker-compose.yml") -> None:
+        """Compile a pipeline to docker-compose spec that is saved on the package_path."""
```
PhilippeMoussalli, line 29:
```suggestion
        """Compile a pipeline to docker-compose spec and save it to a specified package_path."""
```

fondant/compiler.py
```diff
@@ -0,0 +1,103 @@
...
+        """
+        return component_name.replace(" ", "_").lower()
+
+    def _patch_path(self):
+        """Helper that checks if the base_path is local or remote,
+        if local it patches the base_path and prepares a bind mount.
+        """
+        base_path = Path(self.pipeline.base_path)
+        # check if base path is an existing local folder
+        if base_path.exists():
```
PhilippeMoussalli, line 50: what if the path is local but does not exist? it will assume it's remote In this case

fondant/compiler.py
```diff
@@ -0,0 +1,103 @@
...
+            self.volume = None
+            self.path = self.pipeline.base_path
+
+    def _generate_spec(self) -> dict:
+        """Generate a docker-compose spec as a python dictionary,
+        loops over the pipeline graph to create services and their dependencies.
+        """
+        services = {}
+        for component_name, component in self.pipeline._graph.items():
+            safe_component_name = self._safe_component_name(component_name)
```
PhilippeMoussalli, line 69: might be worth moving the safe name functionality into the pipeline when building the graph since we might need to do the renaming across different runners

fondant/compiler.py
```diff
@@ -0,0 +1,103 @@
...
+    def compose_service(
+        self, component_op: ComponentOp, dependencies: t.List[str]
+    ) -> dict:
+        """Take in a component and create a docker-compose service based on the properties."""
+        # add metadata argument to command
+        metadata = {"run_id": self.pipeline.name, "base_path": self.path}
+        command = ["--metadata", json.dumps(metadata)]
+
+        # add in and out manifest paths to command
+        command.extend(["--output_manifest_path", f"{self.path}/manifest.txt"])
```
PhilippeMoussalli, line 84: would maybe give a unique identifier to the manifest or write it within the component folder to avoid overriding it

fondant/compiler.py
```diff
@@ -0,0 +1,103 @@
...
+
+        # add arguments if any to command
+        for key, value in component_op.arguments.items():
+            command.extend([f"--{key}", f"{value}"])
+
+        # resolve dependencies
+        depends_on = {}
+        if dependencies:
+            # there is only an input manifest if the component has dependencies
+            command.extend(["--input_manifest_path", f"{self.path}/manifest.txt"])
```
PhilippeMoussalli, line 94: same here

PhilippeMoussalli (commented):
I Really like the implementation! looks quite clean ✨

Left a few minor comments, excited to see where this is heading :)

fondant/compiler.py
```diff
@@ -0,0 +1,107 @@
...
+        """
+        base_path = Path(self.pipeline.base_path)
+        # check if base path is an existing local folder
+        if base_path.exists():
+            self.volume = {
+                "type": "bind",
+                "source": str(base_path),
+                "target": f"/{base_path.stem}",
+            }
+            self.path = f"/{base_path.stem}"
```
RobbeSneyders, line 56: Pylint would throw an `attribute-defined-outside-init` here :)

It's a bit hard to follow currently. Can we move the `_patch_path()` call to the `_generate_spec` method and pass them to `compose_service`? This is the only place they're used.

So something like:

```Python
def _generate_spec(self) -> dict:
    ...
    path, volume = self._patch_path()
    for component_name, component in self.pipeline._graph.items():
        ...
        services[safe_component_name] = self.compose_service(
            component["fondant_component_op"], 
            dependencies=component["dependencies"],
            path=path,
            volume=volume,
        )
```

RobbeSneyders (commented):
Thanks @GeorgesLorre! Really nice to have this. Left some comments :slightly_smiling_face:

fondant/compiler.py
GeorgesLorre: Yes, will probably happen once I implement the KubeflowCompiler

fondant/compiler.py
GeorgesLorre: Yes

fondant/compiler.py
GeorgesLorre: I will look more into this once I write the `KubeflowCompiler` which will hopefully clarify this

fondant/compiler.py
PhilippeMoussalli: alright but maybe we can improve the error handling or try creating a new local path? now it assumes that if it's local and that it does not exist that it's remote (you assign it the base path)

docs/pipeline.md
```diff
@@ -37,17 +36,69 @@ def build_pipeline():
...
+In order to compile your pipeline to a `docker-compose` spec you need to import the `DockerCompiler`
+
+```python
+from fondant.compiler import DockerCompiler
+
+compiler = DockerCompiler(pipeline=pipeline)
+compiler.compile()
+```
+
+The DockerCompiler will try to see if the `base_path` of the pipeline is local or remote. If local the base_path will be mounted as a bind volume on every service/component.
```
PhilippeMoussalli, line 94:
```suggestion
The DockerCompiler will try to see if the `base_path` of the pipeline is local or remote. If local the `base_path` will be mounted as a bind volume on every service/component.
```

PhilippeMoussalli (approved):
Thanks Georges :)

tests/test_compiler.py
```diff
@@ -0,0 +1,106 @@
...
+    """Test compiling a pipeline to docker-compose."""
+    compiler = DockerCompiler(pipeline=pipeline)
+    with tmp_path_factory.mktemp("temp") as fn:
+        output_path = str(fn / "docker-compose.yml")
+        compiler.compile(package_path=output_path)
+        with open(output_path, "r") as src, open(VALID_DOCKER_PIPELINE, "r") as truth:
+            assert src.read() == truth.read()
+
+
+def test_docker_local_path(pipeline, tmp_path_factory):
```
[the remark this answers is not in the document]
GeorgesLorre: fixed

Issue #138, opened 2023-05-16
Implement a WriteComponent class


PhilippeMoussalli:
### Problem Statement

Currently we have two types of components:
* `LoadComponent` : takes as input user arguments for loading dataset (e.g. dataset_id) and loads the dataset from a remote location. This component also creates the initial manifest based on the metadata passed to the component and the component specs. 
The user is expected to change the column names to `subset_field` if the columns of the loaded dataset do not have this. (this might [change](https://github.com/orgs/ml6team/projects/1/views/1?pane=issue&itemId=29600673).

* `TransformComponent` takes as input the evolved manifest based on the component spec and loads the dataset from the artifact registry. it also presents the dataframe to the user as `subset_field`

Both components share some common functionalities based on the abstract [run](https://github.com/ml6team/fondant/blob/536343a9271a06c77ae1e80ea29a615b4947b77e/fondant/component.py#LL84C9-L84C12) method: 
* loading the manifest 
* processing the dataset based on the load or transform function implemented by the user 
* evolving the manifest based on the component spec 
* writing the new datasets
* uploading the manifest 

What is still missing is a `WriteComponent` (this is currently implemented with a transform component). This component should take as an argument a `write_path` and write the dataset with an appropriate schema based on the component spec. 
### Proposed Approach

Let's take the `write_to_hub` [component](https://github.com/ml6team/fondant/blob/main/examples/pipelines/controlnet-interior-design/components/write_to_hub_controlnet/src/main.py) as an example. What we want is the following:
* Allow users to write the dataset with custom columns names 
This can be done by providing a mapping dict argument similar to what Bert proposed for the loading component [link](https://github.com/orgs/ml6team/projects/1/views/1?pane=issue&itemId=29600673)
* The user returns the dataset with the modified names and we take care of writing it to the appropriate location 
Two options here:   

   * **(Option A):** User has to implement the writing functions themselves similar to how it's currently implemented [here](https://github.com/ml6team/fondant/blob/main/examples/pipelines/controlnet-interior-design/components/write_to_hub_controlnet/src/main.py)
  		Not ideal since they have to call the schema stuff and the `to_parquet` method which we normally handle in the backend.   
  * **(Option B):** User returns a url of where to write the data as well as the dataframe. We take care of writing the dataset to the appropriate location. However, this would require us to implement a `DaskDataSink` component similar to the `DaskDataWriter` (we have to change some names) where the biggest difference is that:
[...]

PR #196, opened 2023-06-12
Add writer component

PhilippeMoussalli:
PR that adds a writer class as discussed in #138. 
This enables us to write the final dataset without having to write the dataset and manifest since there is no modification made on the data. 

Next steps: 
- Enable default and optional arguments in components. The optional arguments are needed to make the Reader/Writer components generic (e.g. Write to hub requires special hf metadata to be attached to the image column in case there is any, user needs to pass an optional argument specifying the columns name of the image) 
- Re implement load/Write to hub component to make them more generic.

### DIFF
--- a/fondant/component.py
+++ b/fondant/component.py
@@ -112,7 +112,7 @@ class Component(ABC):
         """Abstract method that returns the dataset manifest."""
 
     @abstractmethod
-    def _process_dataset(self, manifest: Manifest) -> dd.DataFrame:
+    def _process_dataset(self, manifest: Manifest) -> t.Union[None, dd.DataFrame]:
         """Abstract method that processes the manifest and
         returns another dataframe.
         """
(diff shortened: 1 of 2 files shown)


fondant/component.py
```diff
@@ -237,3 +237,57 @@ def _process_dataset(self, manifest: Manifest) -> dd.DataFrame:
         df = self.transform(dataframe=df, **self.user_arguments)
 
         return df
+
+
+class WriteComponent(Component):
+    """Base class for a Fondant write component."""
+
+    @classmethod
+    def _add_and_parse_args(cls, spec: ComponentSpec):
```
RobbeSneyders, line 246: This is currently the same as the `TransformComponent`, but I guess the output manifest is not required here?

We still have a lot of duplication in these methods for only small differences, so would like to figure out a better way to do this. Maybe we should still create separate schemas?

RobbeSneyders (commented):
Thanks @PhilippeMoussalli!

fondant/component.py
PhilippeMoussalli: I tackled this in a the PR that follows this by specifying the optional parameters as a specific attribute per component type  

https://github.com/ml6team/fondant/pull/199/files#:~:text=if%20arg.name%20in%20cls.optional_fondant_arguments()%3A 

we could also tackle it [here](https://github.com/ml6team/fondant/blob/56c2dd9f401a2dd17f64ec5dfd572602d3a7e908/fondant/component_spec.py#L211) by having separate secs per component type. This of course would need more reworking and would also required specifying the component type within the component spec yaml (which is not currently the case). Is this what you mean by separate schemas?

Issue #179, opened 2023-06-01
Support default values for component arguments


PhilippeMoussalli:
### Problem Statement

Some components have many arguments associated with them. It is not always necessary for the end-user to define all of them and we might want to stick with default values. Currently this is not available and all the components require an argument. 

This feature should enable users to set default values for certain components while defining the components. Those values can then be optionally passed during the pipeline definition.

We first define the default arguments at the `component_spec` level: 

_compnent_spec.yaml_
```yaml
name:component
description: Component that downloads images based on URLs
image: ghcr.io/ml6team/download_images:latest

consumes:
   ...

produces:
  ...

args:
  min_image_size: #mandatory argument
    description: Minimum size of the images.
    type: int
  timeout: #optional argument
    description: Maximum time (in seconds) to wait when trying to download an image 
    default: 3
    type: int
```

Then, at the step when we add and parse the arguments in  [`component.py`](https://github.com/ml6team/fondant/blob/43567274fc00c87c96011ec0dc3c18931d54e10a/fondant/component.py#L104)
we check if a default argument is provided, if so we set the optional field when adding the argument: 

```python

    def _add_and_parse_args(self):
        parser = argparse.ArgumentParser()
        component_arguments = self._get_component_arguments()

        for arg in component_arguments.values():
            # Input manifest is not required for loading component
            if arg.name == "input_manifest_path":
                input_required = False
            else:
                input_required = True
           
            if arg.default is not None:
               default = arg.default
            else: 
               default = None
            parser.add_argument(
                f"--{arg.name}",
                type=kubeflow2python_type[arg.type],
                required=input_required,
                default=default,
                help=arg.description,
            )

        return parser.parse_args()
```

Additionally, we need to make the necessary changes to the `component_spec.py` by adding an optional `default` field to the [`Argument`](https://github.com/ml6team/fondant/blob/43567274fc00c87c96011ec0dc3c18931d54e10a/fondant/component_spec.py#L41) dataclass as well as modifying the [`KubeflowComponentSpec`](https://github.com/ml6team/fondant/blob/43567274fc00c87c96011ec0dc3c18931d54e10a/fondant/component_spec.py#L195) as specified [here](https://www.kubeflow.org/docs/components/pipelines/v1/reference/component-spec/#:~:text=the%20input%20arguments.-,Example%20of%20a%20component%20specification,-A%20component%20specification) to include default arguments. 
 
### Implementation Steps/Tasks

[...]

PR #199, opened 2023-06-12
Enable default component arguments

PhilippeMoussalli:
PR that enables adding default arguments as discussed in #179. 

Users can now define default arguments in the component specs. Those arguments do not have to be explicitly defined in the `ComponentOp` but could still be overridden.  

Note: please review this [PR](https://github.com/ml6team/fondant/pull/196) beforehand since I am branched off from it. 

Created a separate ticket to do the necessary changes #198. Best to handle it in a separate PR.

### DIFF
--- a/docs/component_spec.md
+++ b/docs/component_spec.md
@@ -122,9 +122,7 @@ Please check the [examples](#examples) below to build a better understanding.
 
 The `args` section describes which arguments the component takes. Each argument is defined by a 
 `description` and a `type`, which should be one of the builtin Python types. Additionally, you can 
-pass an optional argument in the optional `default` field of the arguments. This can be useful in case
-your function has many arguments that could be set to default and that don't need to be explicitly defined when
-initializing your component. 
+set an optional `default` value for each argument.
 
 _Note:_ default iterable arguments such as `dict` and `list` have to be passed as a string 
 (e.g. `'{"foo":1, "bar":2}`, `'["foo","bar]'`)
--- a/docs/component_spec.md
+++ b/docs/component_spec.md
@@ -121,41 +121,52 @@ Please check the [examples](#examples) below to build a better understanding.
 ### Args
 
 The `args` section describes which arguments the component takes. Each argument is defined by a 
-`description` and a `type`, which should be one of the builtin Python types.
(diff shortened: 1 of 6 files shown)


docs/component_spec.md
```diff
@@ -121,41 +121,52 @@ Please check the [examples](#examples) below to build a better understanding.
...
+    description: A default argument
+    type: str
+    default: bar
 ``` 
 
-These arguments are passed in when the component is instantiated:
-
+These arguments are passed in when the component is instantiated. Notice that we are not passing the
+default argument specified above. You could override the default value in the component spec by passing
+it as an argument to the component.
```
RobbeSneyders, line 144:
```suggestion
These arguments are passed in when the component is instantiated. 
If an argument is not explicitly provided, the default value will be used instead if available.```

File: docs/component_spec.md [hunk 2]
```diff
@@ -121,41 +121,52 @@ Please check the [examples](#examples) below to build a better understanding.
 ### Args
 
 The `args` section describes which arguments the component takes. Each argument is defined by a 
-`description` and a `type`, which should be one of the builtin Python types.
+`description` and a `type`, which should be one of the builtin Python types. Additionally, you can 
+pass an optional argument in the optional `default` field of the arguments. This can be useful in case
+your function has many arguments that could be set to default and that don't need to be explicitly defined when
+initializing your component.
```
RobbeSneyders, line 127:
```suggestion
`description` and a `type`, which should be one of the builtin Python types. Additionally, you can 
set an optional `default` value for each argument.
```

fondant/component.py
```diff
@@ -242,20 +240,9 @@ def _process_dataset(self, manifest: Manifest) -> t.Union[None, dd.DataFrame]:
...
-                f"--{arg.name}",
-                type=kubeflow2python_type[arg.type],  # type: ignore
-                required=True,
-                help=arg.description,
-            )
-
-        return parser.parse_args()
+    @staticmethod
+    def optional_fondant_arguments() -> t.List[str]:
+        return []
```
RobbeSneyders, line 245: I would expect the `output_manifest_path` here

RobbeSneyders (approved):
Thanks @PhilippeMoussalli, just some suggestions to make the documentation more clear.

PR #200, opened 2023-06-13
Add Pandas interface

### LINKED ISSUE #183: Move to Pandas dataframes as interface
### Problem Statement

Currently, the Fondant component interface is based on a Dask Dataframe:

```Python
import dask.dataframe as dd

from fondant.component import TransformComponent


class MyComponent(TransformComponent):

    def transform(dataframe: dd.Dataframe, **kwargs): -> dd.Dataframe
        # User code goes here
```

Which means that the user needs to work with a Dask Dataframe, which raises two challenges:
- Dask requires quite a big learning curve to use efficiently. 
- It's hard to let Fondant completely handle Dask optimization such as indexing and partitioning as this might interfere with user Dask code

### Proposed Approach

Instead, we can base our interface on a Pandas dataframe:

```Python
import pandas as pd

from fondant.component import TransformComponent


class MyComponent(TransformComponent):

    def transform(dataframe: pd.Dataframe, **kwargs): -> pd.Dataframe
        # User code goes here
```

Which lets the user write code handling an in-memory Pandas dataframe, and moves all Dask specific optimization to Fondant.

Since a Dask dataframe is composed of many smaller Pandas dataframes (partitions) split along the index, this implementation should align nicely with the internal Dask design. It should also scale to distributed computing since the unit of distribution is a partition.

The Fondant component class can leverage [`map_partitions`](https://docs.dask.org/en/stable/generated/dask.dataframe.DataFrame.map_partitions.html) in the `_process_dataset()` method to achieve this.

### Limitations

- The `map_partitions` method assumes that the index and divisions are unchanged, and otherwise requires resetting the divisions on the result, which can be expensive if they need to be recalculated.

  I propose to check if the returned index is identical to the original index. If it is, this assumption is valid. If it is not, we call `clear_divisions()` and defer division calculation to when it is needed.

  If we clear the divisions, we should also check that `additionalSubsets` and `additionalFields` are set correctly. Since the index is changed, `additionalSubsets` should be false, and `additionalFields` should be false for any subsets part of the input subsets.

- The user only gets access to one partition at a time, inhibiting any processing spanning across partitions (eg. global operations such as mean calculation). Enabling this kind of operation is a challenge for most data processing frameworks, not the focus of Fondant, and can be approximated by running on each partition instead.

  By implementing this as a `PandasTransformComponent` subclassing the `TransformComponent`, users can still use the `TransformComponent` to work with Dask directly.

### Implementation Steps/Tasks

- [ ] Implement `PandasTransformComponent`
- [ ] Update reusable and example components

### Testing

We can test this with unit tests.

### Documentation

We should document the `PandasTransformComponent` as the default way of building components with Fondant. Using the base `TransformComponent` directly should be explained as advanced usage.

RobbeSneyders:
Fixes #183 

There's some todo's left, we should
- [ ] Look into the redefinition of the divisions after we clear them. Now that we take this out of the hands of the user, we should define which strategy we want to follow here. (#205)
- [ ] Move to `hierarchical columns`. Pandas can work with hierarchical columns, which would be a lot nicer as a user interface. I want to check if I can make this work with Dask, and otherwise move the translation from underscored names to hierarchical columns and back at the level of the `PandasTransformComponent` (#204)
- [ ] Update the reusable components to leverage the `PandasTransformComponent` (#203)


fondant/component.py
```diff
@@ -223,17 +224,70 @@ def transform(self, *args, **kwargs) -> dd.DataFrame:
...
+        Load the data based on the manifest using a DaskDataloader and call the transform method to
+        process it.
+
+        Returns:
+            A `dd.DataFrame` instance with updated data based on the applied data transformations.
+        """
+        data_loader = DaskDataLoader(manifest=manifest, component_spec=self.spec)
+        df = data_loader.load_dataframe()
+
+        self.setup(**self.user_arguments)
```
PhilippeMoussalli, line 260: This takes in all the arguments passed to the components which some of them might be related to the setup while other can be related to the custom transform.

fondant/component.py
```diff
@@ -223,17 +224,70 @@ def transform(self, *args, **kwargs) -> dd.DataFrame:
...
+            self.transform,
+            meta=meta_df,
+        )
+
+        if self._infer_index_change():
+            df.clear_divisions()
 
         return df
+
+    def _infer_index_change(self):
```
PhilippeMoussalli, line 281:
```suggestion
    def _infer_index_change(self) -> bool:
```

fondant/component.py
```diff
@@ -223,17 +224,70 @@ def transform(self, *args, **kwargs) -> dd.DataFrame:
...
         """
         data_loader = DaskDataLoader(manifest=manifest, component_spec=self.spec)
         df = data_loader.load_dataframe()
         df = self.transform(dataframe=df, **self.user_arguments)
+        return df
+
+
+class PandasTransformComponent(TransformComponent):
+    @abstractmethod
+    def setup(self, *args, **kwargs):
```
PhilippeMoussalli, line 244: Nice :) should this be moved over to the general `Component` class

PhilippeMoussalli (commented):
Thanks Robbe :) looking nice! 

Left a few comments and questions

components/prompt_based_laion_retrieval/src/main.py
```diff
@@ -81,18 +45,40 @@ def transform(
...
+                        prompt
+                    )
+                    for prompt in dataframe["prompts_text"]
+                ]
+                for response in await asyncio.gather(*futures):
+                    results.extend(response)
+
+        loop.run_until_complete(async_query())
+
+        results_df = pd.DataFrame(results)[["id", "url"]]
```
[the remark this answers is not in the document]
RobbeSneyders: Yes, but this is a Pandas dataframe, which corresponds to a Dask partition. Dask always reads data in memory per partition.

fondant/component.py
RobbeSneyders: These arguments are at the component level, so it makes sense to pass them in during setup. If the user wants to leverage them in the `transform` function, they can register them as instance attributes in the setup method (`self.arg = arg`) and use them in the `transform` method.

fondant/component.py
RobbeSneyders: No, this is only needed in the `PandasTransformComponent` since its `transform` method is called multiple times. The `TransformComponent`'s `transform` method is only called once.

components/prompt_based_laion_retrieval/src/main.py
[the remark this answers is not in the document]
PhilippeMoussalli: ok makes sense

fondant/component.py
PhilippeMoussalli: alright so if I understand correctly this is mainly to avoid re-initializing arguments everytime you execute `transform` on a partition? A bit similar to how Beam does it where you have DoFn.setup() and DoFn.process() (equivalent to our transform)

fondant/component.py
PhilippeMoussalli: ok thanks for clarifying

components/embedding_based_laion_retrieval/src/main.py
```diff
@@ -80,19 +45,40 @@ def transform(
...
+                        executor,
+                        self.client.query,
+                        embedding
+                    )
+                    for embedding in dataframe["embeddings_data"]
+                ]
+                for response in await asyncio.gather(*futures):
+                    results.extend(response)
+
+        loop.run_until_complete(async_query())
```
NielsRogge, line 75: This looks fairly complex, are users expected to write this?

fondant/component.py
```diff
@@ -223,19 +225,105 @@ def transform(self, *args, **kwargs) -> dd.DataFrame:
...
+    def transform(self, *args, **kwargs) -> dd.DataFrame:
+        """
+        Abstract method for applying data transformations to the input dataframe.
+
+        Args:
+            args: A Dask dataframe will be passed in as a positional argument
+            kwargs: Arguments provided to the component are passed as keyword arguments
+        """
+
+    def _process_dataset(self, manifest: Manifest) -> dd.DataFrame:
```
PhilippeMoussalli, line 247: the return type seems to differ from the one defined in the abstract method `t.Union[None, dd.DataFrame]`

PhilippeMoussalli (approved):
Thanks Robbe :+1:

fondant/component.py
RobbeSneyders: I don't think that's an issue. The `None` is only for the `WriteComponent`, and the `dd.DataFrame` for the `Transform` and `LoadComponent`. I updated it everywhere to reflect this.

components/prompt_based_laion_retrieval/src/main.py
```diff
@@ -81,18 +45,40 @@ def transform(
...
+            dataframe: Dask dataframe
 
-        return dataframe
+        Returns:
+            Dask dataframe
+        """
+        results: t.List[t.Tuple[str]] = []
+        loop = asyncio.new_event_loop()
+
+        async def async_query():
```
GeorgesLorre, line 62: async is cool but hopefully not required since it is not easy to write

GeorgesLorre:
Cool stuff, do we need to change more reusable components to use pandas now ?

fondant/component.py
```diff
@@ -221,21 +223,107 @@ def transform(self, *args, **kwargs) -> dd.DataFrame:
...
+
+        # Call the component transform method for each partition
+        df = df.map_partitions(
+            self.transform,
+            meta=meta_df,
+        )
+
+        # Clear divisions if component spec indicates that the index is changed
+        if self._infer_index_change():
+            df.clear_divisions()
```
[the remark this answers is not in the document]
RobbeSneyders: Dask will automatically infer them when needed. I created a ticket to optimize this.

components/prompt_based_laion_retrieval/src/main.py
RobbeSneyders: This was already in the implementation of the component, so no change here.

RobbeSneyders:
> Cool stuff, do we need to change more reusable components to use pandas now ?

They can keep working with the `DaskTransformComponent` for now, but would indeed propose to update them. There's a ticket to follow this up.

fondant/component.py
RobbeSneyders: They shouldn't return a mismatch when returning a stricter value, only when accepting a stricter one.

PR #201, opened 2023-06-13
Enable optional component arguments

PhilippeMoussalli:
PR that enables defining optional arguments to the component spec. This is needed to enable building reusable components. 

For example, the `load_from_hub` component will need to format images from hf format to bytes and this can be achieved by passing a specific argument (`image_column_name`). This can be achieved either by: 

- Passing `None` explicitly as an argument value at the `ComponentOp` level
- Assigning `None` as the `default` argument field in the component spec. 

Previous attempt was to achieve this with the [optional field of the kubeflow spec](https://www.kubeflow.org/docs/components/pipelines/v1/reference/component-spec/#:~:text=Kubeflow%20Pipelines%20SDK.-,optional,-%3A%20Specifies%20if%20input) but this returned many errors and also requires adding in an additional `optional` field to the fondant spec. The current approach achieves the same outcome with a simpler approach. 

The `component_spec` argument is now also defined as an optional argument. This allows use to define a component using the `from_file()` method without having to explicitly define that argument. We still return a valid error if `from_args()` is used and no `component_spec` argument is provided since it defaults to `None` [link](https://github.com/ml6team/fondant/blob/e90ac782a8c253720456fcbd05a9ced086ca5547/fondant/component.py#L61)


fondant/component_spec.py
```diff
@@ -18,12 +18,12 @@
...
-    "Float": float,
-    "Boolean": ast.literal_eval,
-    "JsonObject": json.loads,
-    "JsonArray": json.loads,
+    "String": lambda value: str(value) if value != "None" else None,
+    "Integer": lambda value: int(value) if value != "None" else None,
+    "Float": lambda value: float(value) if value != "None" else None,
+    "Boolean": lambda value: ast.literal_eval(value) if value != "None" else None,
+    "JsonObject": lambda value: json.loads(value) if value != "None" else None,
+    "JsonArray": lambda value: json.loads(value) if value != "None" else None,
```
RobbeSneyders, line 26: Maybe it makes sense converting this into a function.

```suggestion
kubeflow_to_python_type_dict = {
    "String": str,
    "Integer": int,
    "Float": float,
    "Boolean": ast.literal_eval,
    "JsonObject": json.loads,
    "JsonArray": json.loads,
}

def kubeflow_to_python_type(type_: str) -> t.Any:
    map_fn = kubeflow_to_python_type_dict[type_]
    return lambda value: map_fn(value) if value != "None" else None
```

RobbeSneyders (commented):
Thanks @PhilippeMoussalli, I think this approach is indeed quite elegant! Left one comment and you'll need to rebase / merge :)

fondant/component_spec.py
GeorgesLorre: or use the `match case` syntax: https://peps.python.org/pep-0622/

PhilippeMoussalli: Thanks for the feedback :) I think it's ready to go

GeorgesLorre (approved):
thx!

PR #207, opened 2023-06-14
Add build for local components

GeorgesLorre:


fondant/compiler.py
```diff
@@ -126,10 +126,17 @@ def _generate_spec(self, pipeline: Pipeline) -> dict:
...
-                "image": component_op.component_spec.image,
                 "command": command,
                 "depends_on": depends_on,
                 "volumes": volumes,
             }
 
+            if component_op.local_component:
+                services[safe_component_name][
+                    "build"
+                ] = f"./{Path(component_op.component_spec_path).parent}"
```
ChristiaensBert, line 137: Can the `output_path` argument be any location on your filesystem? If so, does it make sense to use the "./{Path()}" format or should make it depend on the `output_path`? I assume the `docker-compose up` will be ran from the `output_path`, which could be in a different directory than the local components.

fondant/compiler.py
ChristiaensBert, line 1: I only see this now, but on `line 59`

```
logger.info(f"Compiling {pipeline.name} to docker-compose.yml")
```
we won't always have `docker-compose.yml` as the filename/path. Better to change it to {output_path}

fondant/compiler.py
```diff
@@ -126,10 +126,17 @@ def _generate_spec(self, pipeline: Pipeline) -> dict:
             volumes = [asdict(volume)] if volume else []
 
             services[safe_component_name] = {
-                "image": component_op.component_spec.image,
                 "command": command,
                 "depends_on": depends_on,
                 "volumes": volumes,
```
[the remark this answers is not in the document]
GeorgesLorre: only when the base_path is local otherwise the volume will be empty
```
    depends_on:
      second_component:
        condition: service_completed_successfully
    volumes: []
```

fondant/compiler.py
GeorgesLorre: You mean reference the component code relative from the output_path ?

RobbeSneyders (approved):

PR #209, opened 2023-06-14
Run ruff on components

RobbeSneyders:


RobbeSneyders:
Since we copied them from a different source, I would leave them as is and not make any changes. This is also clearest to define which license is appropriate.

PR #212, opened 2023-06-15
Add support for mounting custom volumes (cloud credentials)

GeorgesLorre:


ChristiaensBert:
There seem to be a lot of file changes with formatting / renaming. Is this from the Pandas PR?

docs/pipeline.md
```diff
@@ -79,9 +82,11 @@ services:
     depends_on:
       component_2:
         condition: service_completed_successfully
-    image: component_3:latest
+    build: ./component_3
 ```
 
+Note that for components that do not come from the registry (local custom components) the compiler will add a build subsection instead (see component_3 in the example above) of referring to the image specified in the `component_spec.yaml`. This allows docker-compose to build and rebuild the container used in that component allowing for quicker iteration.
```
RobbeSneyders, line 88:
```suggestion
Note that for components that do not come from the registry (local custom components) the compiler will add a build subsection (see component_3 in the example above) instead of referring to the image specified in the `component_spec.yaml`. This allows docker-compose to build and rebuild the container used in that component allowing for quicker iteration.
```

fondant/compiler.py
```diff
@@ -106,7 +117,7 @@ def _generate_spec(self, pipeline: Pipeline) -> dict:
             command = ["--metadata", json.dumps(asdict(metadata))]
 
             # add in and out manifest paths to command
-            command.extend(["--output_manifest_path", f"{path}/manifest.txt"])
+            command.extend(["--output_manifest_path", f"{path}/manifest.json"])
```
RobbeSneyders, line 120: This still overwrites the manifest for every component. We should still address this, but can be taken up in a separate PR.

GeorgesLorre: will fix in another PR

PR #213, opened 2023-06-15
add remote credentials and manifest loading

ChristiaensBert:
Add support to the data explorer for remote files. Uses fsspec to load in the file before passing it to the manifest.

Use the -c or --credentials to pass your google credentials json file

For testing, you can use this remote manifest
```
gs://soy-audio-379412_datasets/data_explorer/manifest_gcs.txt
```

### DIFF
--- a/data_explorer/requirements.txt
+++ b/data_explorer/requirements.txt
@@ -4,3 +4,5 @@ streamlit-aggrid==0.3.4
 pyarrow>=7.0
 matplotlib==3.7.1
 plotly==5.15.0
+fsspec==2023.4.0
+gcsfs==2023.4.0
--- a/fondant/cli.py
+++ b/fondant/cli.py
@@ -34,7 +34,11 @@ def run_data_explorer():
     parser.add_argument(
         "--data-directory",
         "-d",
-        help="Path to the source directory that contains the data produced by a fondant pipeline.",
+        help="Path to the source directory that contains the data produced by a fondant pipeline. \
+            Google cloud credentials: \
+            https://cloud.google.com/docs/authentication/application-default-credentials \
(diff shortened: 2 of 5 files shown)


fondant/cli.py
```diff
@@ -64,12 +71,25 @@ def run_data_explorer():
...
         "--mount",
         f"type=bind,source={shlex.quote(args.data_directory)},target=/artifacts",
-        f"{shlex.quote(args.registry)}:{shlex.quote(args.tag)}",
     ]
 
+    if args.credentials:
+        cmd.extend(
+            [
+                "-v",
+                f"{args.credentials}:/root/.config/gcloud/application_default_credentials.json:ro",
```
ChristiaensBert, line 83: @GeorgesLorre This TARGET path is still hardcoded. How can we make it simple enough without having the user to search for which TARGET path they have to use for their Cloud Credentials? It's a similar problem for your PR (in the extra_volumes)

data_explorer/requirements.txt
```diff
@@ -3,4 +3,6 @@ fondant==0.1.2
 streamlit-aggrid==0.3.4
 pyarrow>=7.0
 matplotlib==3.7.1
-plotly==5.15.0
\ No newline at end of file
+plotly==5.15.0
+fsspec==2023.4.0
+gcsfs==2023.4.0
```
PhilippeMoussalli, line 8: if you're not planning on using fondant here maybe you need to install the specific cloud dependencies, see: https://github.com/orgs/ml6team/projects/1?pane=issue&itemId=30868442

ChristiaensBert: Good point, I'll add fondant[azure, aws, gcp]

ChristiaensBert: I added s3fs and adlfs

PR #214, opened 2023-06-15
Generic read write component

PhilippeMoussalli:
Draft PR for implementing generic read/write components. 

Note: the changes from `feature/local-starcoder` were merged to this branch to make testing easier. Merge this PR after the local startcoder PR.

Things to do: 

- [x] Modify `from_registry` method to enable passing path to custom spec 
- [x] Figure out how to present spec template to user. Should we keep the image/args and just add a todo in the `produces` and `consumes` (open for suggestions)
- [x] Switch over all pipelines to use the generic load/write and remove the custom ones 
- [x] Add components to affected pipelines to add missing metadata (e.g. width/height)
- [x] Add documentation


components/load_from_hf_hub/fondant_component.yaml
```diff
@@ -3,20 +3,18 @@ description: Component that loads a dataset from the hub
...
-        type: string
+  # fill here
 
 args:
   dataset_name:
     description: Name of dataset on the hub
     type: str
+  column_name_mapping:
+    description: column to map the original column names of the input dataset to
+    type: dict
```
ChristiaensBert, line 14: are we assuming that the dataset only has these two columns? What are we doing with multiple image columns etc?

RobbeSneyders: We should probably turn the `image_column_name` into an array?

components/load_from_hf_hub/fondant_component.yaml
```diff
@@ -3,20 +3,18 @@ description: Component that loads a dataset from the hub
...
+  image_column_name:
+    description: the name of a column containing images. Used to convert the columns
+            to a proper format. The column name can be either the original column name or the subset
```
RobbeSneyders, line 17: > The column name can be either the original column name or the subset

I would leave this out for simplification. Make it clear what is expected from the user and prevent edge cases. Eg:

```
column_name_mapping = {
    "image_bytes": "image_data",
    "image_data": "image_metadata"
}
```
:wink:

components/write_to_hf_hub/fondant_component.yaml
```diff
@@ -0,0 +1,25 @@
...
+    type: str
+  dataset_name:
+    description: The name of the dataset to upload
+    type: str
+  image_column:
+    description: the name of the image column. Used to format to image to HF hub format
+    type: str
+    default: None
+  column_name_mapping:
+    description: column to map the original column names of the input dataset to
```
RobbeSneyders, line 23:
```suggestion
    description:  Mapping of the consumed fondant column names to the written hub column names.```

**Review by RobbeSneyders — COMMENTED** [review]
Thanks @PhilippeMoussalli. This should make it a lot easier for users to build a pipeline without having to build any custom component!

File: components/load_from_hf_hub/fondant_component.yaml (hunk 1 above)
Comment (PhilippeMoussalli) [reply components/load_from_hf_hub/fondant_component.yaml:14]: yes it should be a `list` since we might have more than one image, i'll adjust it

File: components/load_from_hf_hub/fondant_component.yaml (hunk 2 above)
Comment (PhilippeMoussalli) [reply components/load_from_hf_hub/fondant_component.yaml:17]: oh didn't think of that one, I will keep it as the original column name

File: components/load_from_hf_hub/src/main.py [hunk 4]
```diff
@@ -13,52 +11,33 @@
...
-            lambda x: x["bytes"], meta=("bytes", bytes)
-        )
-
-        # 4) Add width and height columns
-        dask_df["images_width"] = dask_df["images_data"].map(
-            extract_width, meta=("images_width", int)
-        )
-        dask_df["images_height"] = dask_df["images_data"].map(
-            extract_height, meta=("images_height", int)
-        )
```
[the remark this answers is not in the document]
PhilippeMoussalli: I'm not completely sure about that one, it might be that the original dataset has this metadata and we're assuming that the user requires it. 

I was thinking about implementing another component that generates image metadata to go around this, it can be based on a conditional arguments (e.g. `estimate_width`: True/False, ..).

components/write_to_hf_hub/fondant_component.yaml
```diff
@@ -0,0 +1,25 @@
...
+  hf_token:
+    description: The hugging face token used to write to the hub
+    type: str
+  username:
+    description: The username under which to upload the dataset
+    type: str
+  dataset_name:
+    description: The name of the dataset to upload
+    type: str
+  image_column:
```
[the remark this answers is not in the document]
PhilippeMoussalli: yeah indeed

components/load_from_hf_hub/src/main.py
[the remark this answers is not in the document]
RobbeSneyders: Yes, downside is that it requires loading the image data into the component. We might be able to do this by only loading the first x bytes, but not sure how this works with different image formats and if this can be done performant with parquet.

components/load_from_hf_hub/requirements.txt
```diff
@@ -1,4 +1,5 @@
 huggingface_hub==0.14.1
+datasets==2.10.1
```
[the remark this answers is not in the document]
PhilippeMoussalli: good catch, this should be in the 'write_to_hub` component and not here

components/load_from_hf_hub/src/main.py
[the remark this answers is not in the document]
PhilippeMoussalli: found this library: https://github.com/shibukawa/imagesize_py
Tested it out and it seems to work pretty well

components/image_resolution_extraction/src/main.py
```diff
@@ -0,0 +1,55 @@
+"""This component filters images of the dataset based on image size (minimum height and width)."""
+import io
+import logging
+import typing as t
+
+import dask.dataframe as dd
+import imagesize
+import numpy as np
+
+from fondant.component import DaskTransformComponent
```
RobbeSneyders, line 10:
```suggestion
from fondant.component import PandasTransformComponent
```

components/image_resolution_extraction/src/main.py
```diff
@@ -0,0 +1,55 @@
...
+        """
+        logger.info("Length of the dataframe before filtering: %s", len(dataframe))
+
+        logger.info("Filtering dataset...")
+
+        dataframe[["images_width", "images_height"]] = \
+            dataframe[["images_data"]].apply(extract_dimensions,
+                                             axis=1, result_type="expand", meta={0: int, 1: int})
+
+        return dataframe
```
RobbeSneyders, line 50:
```suggestion
class ImageResolutionExtractionComponent(PandasTransformComponent):
    """Component that extracts image dimensions."""

    def transform(self, *, dataframe: pd.DataFrame) -> pd.DataFrame:
        logger.info("Length of the dataframe before filtering: %s", len(dataframe))

        logger.info("Filtering dataset...")

        dataframe[[("images", "width"), ("images", "height")]] = \
            dataframe[[("images", "data")]].apply(extract_dimensions,
                                             axis=1, result_type="expand")

        return dataframe
```

components/image_resolution_extraction/src/main.py
```diff
@@ -0,0 +1,55 @@
...
+    Args:
+        image_df (dd.DataFrame): input dataframe with images_data column
+
+    Returns:
+        np.int16: width of the image
+        np.int16: height of the image
+    """
+    width, height = imagesize.get(io.BytesIO(image_df["images_data"]))
+
+    return np.int16(width), np.int16(height)
```
RobbeSneyders, line 29:
```suggestion
def extract_dimensions(images: pd.Series) -> t.Tuple[np.int16, np.int16]:
    """Extract the width and height of an image.

    Args:
        images: input series with image data

    Returns:
        np.int16: width of the image
        np.int16: height of the image
    """
    width, height = imagesize.get(io.BytesIO(images))

    return np.int16(width), np.int16(height)
```

components/image_resolution_extraction/src/main.py
```diff
@@ -0,0 +1,55 @@
+"""This component filters images of the dataset based on image size (minimum height and width)."""
+import io
+import logging
+import typing as t
+
+import dask.dataframe as dd
```
RobbeSneyders, line 6:
```suggestion
import pandas as pd
```

components/image_resolution_extraction/src/main.py
RobbeSneyders: You might want to test this :sweat_smile:

components/load_from_hf_hub/fondant_component.yaml
```diff
@@ -2,21 +2,25 @@ name: Load from hub
...
     type: str
+  column_name_mapping:
+    description: Mapping of the consumed hub dataset to fondant column names
+    type: dict
+  image_column_names:
+    description: A list containing the original hub image column names. Used to format the image
+      from HF hub format to a byte string
+    type: list
+    default: None
+  nb_rows_to_load:
```
RobbeSneyders, line 23:
```suggestion
  n_rows_to_load:
```

Don't think "nb" is used as abbreviation of number. Maybe "no", but that would be confusing here.

components/load_from_hf_hub/src/main.py
```diff
@@ -13,52 +11,43 @@
...
-    height = pil_image.size[1]
-
-    return np.int16(height)
-
-
 class LoadFromHubComponent(LoadComponent):
-    def load(self, *, dataset_name: str) -> dd.DataFrame:
+    def load(self,
+             *,
+             dataset_name: str,
```
RobbeSneyders, line 17:
```suggestion
             dataset_name: str,
             *,
```

The `dataset_name` is the main argument and the subject to `load`, so it makes sense to include this as a positional argument.

This makes sense:
```python
component.load("my_dataset")
```

This doesn't:
```python
component.load("my_dataset", {"original_name": "new_name"}, ["original_name"], 5)
```

components/load_from_hf_hub/src/main.py
```diff
@@ -13,52 +11,43 @@
...
-        # 4) Add width and height columns
-        dask_df["images_width"] = dask_df["images_data"].map(
-            extract_width, meta=("images_width", int)
-        )
-        dask_df["images_height"] = dask_df["images_data"].map(
-            extract_height, meta=("images_height", int)
-        )
+        if nb_rows_to_load:
+            dask_df = dask_df.head(nb_rows_to_load)
+            dask_df = dd.from_pandas(dask_df, npartitions=1)
```
RobbeSneyders, line 50: Calling `.head()` and setting `npartitions=1` only makes sense for small `nb_rows_to_load`, which might not be the case. What if I want to run on 1M rows for testing instead of 5B? :slightly_smiling_face: 

On the other hand, I'm having trouble finding a good way to do this using Dask without knowing the index labels. What is possible, is sampling, or selecting a number of partitions, but this might be less usable for the user.

examples/pipelines/controlnet-interior-design/components/generate_prompts/fondant_component.yaml
```diff
@@ -6,4 +6,10 @@ produces:
   prompts:
     fields:
       text:
-        type: string
\ No newline at end of file
+        type: string
+
+args:
+  nb_rows_to_load:
```
RobbeSneyders, line 12:
```suggestion
  n_rows_to_load:
```

Same as above

examples/pipelines/controlnet-interior-design/components/generate_prompts/src/main.py
```diff
@@ -97,10 +98,13 @@ def make_interior_prompt(room: str, prefix: str, style: str) -> str:
 
 
 class GeneratePromptsComponent(LoadComponent):
-    def load(self) -> dd.DataFrame:
+    def load(self, nb_rows_to_load: t.Optional[int]) -> dd.DataFrame:
```
RobbeSneyders, line 101:
```suggestion
    def load(self, n_rows_to_load: t.Optional[int]) -> dd.DataFrame:
```

examples/pipelines/controlnet-interior-design/components/generate_prompts/src/main.py
```diff
@@ -109,6 +113,9 @@ def load(self) -> dd.DataFrame:
 
         pandas_df = pd.DataFrame(prompts, columns=["prompts_text"])
 
+        if nb_rows_to_load:
+            pandas_df = pandas_df.head(nb_rows_to_load)
```
RobbeSneyders, line 117:
```suggestion
        if n_rows_to_load:
            pandas_df = pandas_df.head(n_rows_to_load)
```

examples/pipelines/controlnet-interior-design/pipeline.py
```diff
@@ -82,4 +87,18 @@
...
+        extra_volumes = [
+            "$HOME/.config/gcloud/application_default_credentials.json:/root/.config/gcloud/"
+            "application_default_credentials.json:ro"
+        ]
+        compiler.compile(pipeline=pipeline, extra_volumes=extra_volumes)
+        logger.info("Run `docker-compose up` to run the pipeline.")
+
+    else:
+        client = Client(host=PipelineConfigs.HOST)
+        client.compile_and_run(pipeline=pipeline)
```
RobbeSneyders, line 104: Let's not include this. This should be handled by the `fondant` command in the end.

examples/pipelines/finetune_stable_diffusion/components/load_from_hf_hub/fondant_component.yaml
```diff
@@ -0,0 +1,31 @@
...
+    type: str
+  column_name_mapping:
+    description: Mapping of the consumed hub dataset to fondant column names
+    type: dict
+  image_column_names:
+    description: A list containing the original hub image column names. Used to format the image
+      from HF hub format to a byte string
+    type: list
+    default: None
+  nb_rows_to_load:
```
RobbeSneyders, line 28:
```suggestion
  n_rows_to_load:
```

examples/pipelines/finetune_stable_diffusion/pipeline.py
```diff
@@ -60,13 +78,41 @@
...
+            "$HOME/.config/gcloud/application_default_credentials.json:/root/.config/gcloud/"
+            "application_default_credentials.json:ro"
+        ]
+        compiler.compile(pipeline=pipeline, extra_volumes=extra_volumes)
+        logger.info("Run `docker-compose up` to run the pipeline.")
 
-client.compile_and_run(pipeline=pipeline)
+    else:
+        client = Client(host=PipelineConfigs.HOST)
+        client.compile_and_run(pipeline=pipeline)
```
RobbeSneyders, line 118: Let's not include this

examples/pipelines/starcoder/components/load_from_hub/fondant_component.yaml
```diff
@@ -0,0 +1,40 @@
...
+    type: str
+  column_name_mapping:
+    description: Mapping of the consumed hub dataset to fondant column names
+    type: dict
+  image_column_names:
+    description: A list containing the original hub image column names. Used to format the image
+      from HF hub format to a byte string
+    type: list
+    default: None
+  nb_rows_to_load:
```
RobbeSneyders, line 37:
```suggestion
  n_rows_to_load:
```

examples/pipelines/starcoder/pipeline.py
```diff
@@ -23,9 +38,14 @@
...
-load_from_hub_op = ComponentOp(
-    component_spec_path="components/load_from_hub_stack/fondant_component.yaml",
-    arguments={"dataset_name": "ml6team/the-stack-smol-python"},
+load_from_hub_op = ComponentOp.from_registry(
+    name="load_from_hub",
+    component_spec_path="components/load_from_hub/fondant_component.yaml",
+    arguments={
+        "dataset_name": "ml6team/the-stack-smol-python",
+        "column_name_mapping": load_component_column_mapping,
+        "nb_rows_to_load": None,
```
RobbeSneyders, line 47:
```suggestion
        "n_rows_to_load": None,
```

fondant/compiler.py
RobbeSneyders, line 1: This should probably be in Georges' PR.

RobbeSneyders (commented):
Thanks @PhilippeMoussalli.

Let's use the `PandasTransformComponent` for all new transform components from now on. And some more smaller comments :slightly_smiling_face:

components/load_from_hf_hub/src/main.py
PhilippeMoussalli: hmm good point, I was considering small scale for end-to-end testing but this might not always be the case. Maybe best to tackle this separately. Created a [ticket](https://github.com/orgs/ml6team/projects/1/views/1?pane=issue&itemId=31253849) for it

fondant/compiler.py
PhilippeMoussalli: Added as a suggestion

components/image_resolution_extraction/src/main.py
```diff
@@ -0,0 +1,52 @@
+"""This component filters images of the dataset based on image size (minimum height and width)."""
```
NielsRogge, line 1: Docstring to be updated

components/image_resolution_extraction/fondant_component.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+produces:
+  images:
+    fields:
+      width:
+        type: int16
+      height:
+        type: int16
+      data:
+        type: binary
```
NielsRogge, line 19: For components that only add columns, is there a need to specify existing columns in the produces section?

cc @RobbeSneyders

components/load_from_hf_hub/fondant_component.yaml
```diff
@@ -2,21 +2,25 @@ name: Load from hub
...
 args:
   dataset_name:
     description: Name of dataset on the hub
     type: str
+  column_name_mapping:
+    description: Mapping of the consumed hub dataset to fondant column names
+    type: dict
+  image_column_names:
+    description: A list containing the original hub image column names. Used to format the image
+      from HF hub format to a byte string
```
NielsRogge, line 20:
```suggestion
    description: A list containing the original image column names in case the dataset on the hub contains them. Used to format the image from HF hub format to a byte string.
```

components/load_from_hf_hub/fondant_component.yaml
```diff
@@ -2,21 +2,25 @@ name: Load from hub
...
+    type: list
+    default: None
+  n_rows_to_load:
```
NielsRogge, line 23:
```suggestion
  num_rows_to_load:
```

components/load_from_hf_hub/src/main.py
```diff
@@ -13,52 +11,43 @@
...
-        )
+        # 2) Make sure images are bytes instead of dicts
+        if image_column_names:
+            for image_column_name in image_column_names:
+                dask_df[image_column_name] = dask_df[image_column_name].map(
+                    lambda x: x["bytes"], meta=("bytes", bytes)
+                )
+
+        # 3) Rename columns
+        dask_df = dask_df.rename(columns=column_name_mapping)
```
NielsRogge, line 44: This doesn't create hierarchical columns right? 

Is this necessary given that we now use them?

components/write_to_hf_hub/src/main.py
```diff
@@ -0,0 +1,101 @@
...
+                else:
+                    schema_dict[column_name] = datasets.Value(str(field.type.value))
+
+        schema = datasets.Features(schema_dict).arrow_schema
+        dataframe = dataframe[write_columns]
+
+        # Map image column to hf data format
+        feature_encoder = datasets.Image(decode=True)
+
+        if image_column_names:
```
NielsRogge, line 84:
```suggestion
        if image_column_names is not None:
```

components/write_to_hf_hub/src/main.py
```diff
@@ -0,0 +1,101 @@
...
+            for image_column_name in image_column_names:
+                dataframe[image_column_name] = dataframe[image_column_name].map(
+                    lambda x: convert_bytes_to_image(x, feature_encoder),
+                    meta=(image_column_name, "object")
+                )
+
+        # Map column names to hf data format
+        if column_name_mapping:
```
NielsRogge, line 92:
```suggestion
        if column_name_mapping is not None:
```

components/image_resolution_extraction/fondant_component.yaml
RobbeSneyders: In theory not, but I'm not sure if it works in practice already if you leave them out.

components/load_from_hf_hub/fondant_component.yaml
PhilippeMoussalli: I think they're both clear ;P I'll leave it as is for now

components/write_to_hf_hub/src/main.py
RobbeSneyders: If the mapping is empty, we don't need to do this either.

RobbeSneyders (approved):
Great work @PhilippeMoussalli!

PR #215, opened 2023-06-16
Update starcoder example to use the docker compiler

GeorgesLorre:


fondant/compiler.py
```diff
@@ -119,6 +119,11 @@ def _generate_spec(self, pipeline: Pipeline, extra_volumes: list) -> dict:
             # add in and out manifest paths to command
             command.extend(["--output_manifest_path", f"{path}/manifest.json"])
 
+            # add component spec to command if does not exist yet
+            if not component_op.arguments.get("component_spec"):
+                component_spec = component_op.component_spec.specification
+                command.extend(["--component_spec", json.dumps(component_spec)])
+
```
RobbeSneyders, line 126: I would move this to the `ComponentOp` class as this is not compiler specific.

examples/pipelines/starcoder/pipeline.py
```diff
@@ -46,7 +48,20 @@
...
+            "$HOME/.config/gcloud/application_default_credentials.json:/root/.config/gcloud/application_default_credentials.json:ro"
+        ]
+        compiler.compile(pipeline=pipeline, extra_volumes=extra_volumes)
+        logger.info("Run `docker-compose up` to run the pipeline.")
 
-# compile
-client.compile_and_run(pipeline=pipeline)
+    else:
+        client = Client(host=PipelineConfigs.HOST)
+        client.compile_and_run(pipeline=pipeline)
```
RobbeSneyders, line 67: I would prefer to include this into the `Pipeline` class or a separate command before updating the example. Ideally, we don't need any code changes in this file to make it work with a different runner.

RobbeSneyders (commented):
Thanks @GeorgesLorre!

examples/pipelines/starcoder/pipeline.py
```diff
@@ -46,7 +48,20 @@
...
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--local", action="store_true")
+    if parser.parse_args().local:
+        compiler = DockerCompiler()
+        extra_volumes = [
+            "$HOME/.config/gcloud/application_default_credentials.json:/root/.config/gcloud/application_default_credentials.json:ro"
+        ]
+        compiler.compile(pipeline=pipeline, extra_volumes=extra_volumes)
+        logger.info("Run `docker-compose up` to run the pipeline.")
```
PhilippeMoussalli, line 63:
```suggestion
        logger.info("Run `docker compose up` to run the pipeline.")
```

examples/pipelines/starcoder/pipeline.py
GeorgesLorre: Agreed, maybe as a different PR as I don't want to bloat this one

examples/pipelines/starcoder/pipeline.py
PhilippeMoussalli: oh it seems like it works but I had to install it. never mind you can keep it as is

RobbeSneyders: It doesn't for me. `docker compose` is the recommended way ([link](https://stackoverflow.com/a/66516826/4098821))

fondant/compiler.py
```diff
@@ -117,7 +122,17 @@ def _generate_spec(self, pipeline: Pipeline, extra_volumes: list) -> dict:
...
+                ]
+            )
+
+            # add component spec to command if does not exist yet
+            if not component_op.arguments.get("component_spec"):
+                component_spec = component_op.component_spec.specification
+                command.extend(["--component_spec", json.dumps(component_spec)])
 
             # add arguments if any to command
             for key, value in component_op.arguments.items():
```
PhilippeMoussalli, line 138: could you replace 

```python
for key, value in component_op.arguments.items():
    command.extend([f"--{key}", f"{value}"])
```

with 

```python
for key, value in component_op.arguments.items():
  command.extend([f"--{key}", f"{value}"])
  if isinstance(value, (dict, list)):
      command.extend([f"--{key}", json.dumps(value)])
  else:
      command.extend([f"--{key}", f"{value}"])
```

to fix parsing dictionary, list arguments?

GeorgesLorre: done

PhilippeMoussalli: thanks

PhilippeMoussalli (approved):
Thanks Georges!

PR #217, opened 2023-06-19
Include component spec in non-generic reusable components

RobbeSneyders:
Currently running a pipeline with reusable components fails as the component spec is no longer included in the image, but also not provided by the pipeline. We should assume that we have no access to the individual files of these components, as the goal is to move them out of the repository in the future. Since these components are not generic, we can keep including the component spec in the image.


RobbeSneyders:
Closing as this should be handled by the pipeline as discussed.

PR #218, opened 2023-06-19
Add cli entrypoint for compile

GeorgesLorre:
WIP!

First implementation of the fondant CLI (tried and tested):

`fondant compile -p /playground/mycoolpipeline.py:pipeline`

(note you will need to (re)run `poetry install --all-extras`)

I'm mainly looking for feedback on this approach before I spend more time expanding it


fondant/cli.py
```diff
@@ -0,0 +1,96 @@
+import argparse
+import importlib.util
+import re
+
+from fondant.compiler import DockerCompiler
+from fondant.pipeline import Pipeline
+
+
+class PipelinePythonReference(argparse.Action):
```
GeorgesLorre, line 9: I did not find anything standard on importstrings (`file.py:my_object`) so I made my own parser.

RobbeSneyders: Any reason to go for a custom `Action` instead of a custom `type`? I think a custom `type` could be a bit simpler.

FYI, [reference implementation](https://github.com/encode/uvicorn/blob/master/uvicorn/importer.py#L9) from `uvicorn`.

fondant/cli.py
```diff
@@ -0,0 +1,96 @@
...
+            return match.groups()
+        else:
+            raise ValueError(f"Could not parse input {input}")
+
+    @staticmethod
+    def load_pipeline(module: str, object: str) -> Pipeline:
+        """Attempt to load a pipeline from a module file path and object name."""
+        spec = importlib.util.spec_from_file_location(object, module)
+        pipeline = importlib.util.module_from_spec(spec)  # type: ignore
+        spec.loader.exec_module(pipeline)  # type: ignore
```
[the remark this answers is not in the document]
RobbeSneyders: See my comment above for an implementation based on `import_module` and `getattr`.

examples/pipelines/starcoder/pipeline.py
```diff
@@ -1,5 +1,6 @@
 """Pipeline used to create the dataset to train the StarCoder model."""
 
+import argparse
```
RobbeSneyders, line 3: Why was this added? I think we should be able to remove the `if __name__ = "__main__":` block at the bottom of the file, no?

fondant/compiler.py
```diff
@@ -129,6 +129,11 @@ def _generate_spec(self, pipeline: Pipeline, extra_volumes: list) -> dict:
                 ]
             )
 
+            # add component spec to command if does not exist yet
+            if not component_op.arguments.get("component_spec"):
+                component_spec = component_op.component_spec.specification
+                command.extend(["--component_spec", json.dumps(component_spec)])
+
```
RobbeSneyders, line 136: You added this to the `ComponentOp` class in #215, right?

RobbeSneyders (approved):
Thanks @GeorgesLorre! I know it must hurt to see the coverage decline, but looking at the uncovered lines, I don't think there's a lot to test without just duplicating the same logic.

examples/pipelines/starcoder/docker-compose.yml
RobbeSneyders, line 1: This should be left out of git.

RobbeSneyders (changes requested):
One more thing :wink:

RobbeSneyders (approved):
Thanks!

Issue #203, opened 2023-06-14
Update components to implement the PandasTransformComponent


RobbeSneyders:
As follow up of #200, we should update our reusable transform components to leverage the `PandasTransformComponent`

RobbeSneyders:
#219 did this for all relevant components in the Controlnet and Stable Diffusion pipelines. We still need to do this for the others.

PR #219, opened 2023-06-19
 Migrate stable diffusion & controlnet components to PandasTransformComponent

RobbeSneyders:
Contributes to #203 

This PR updates the reusable components of the stable diffusion / controlnet pipeline to use the new Pandas interface. I submit these separately since I've already tested them. 

Some simplifications are clear:
- No more `map_partitions` needed for batching
- No more `meta` needed for `apply` or similar methods

Since we now set the `meta` of the `map_partitions` in the `PandasTransformComponent` based on the component spec, the returned dataframes are validated quite strictly. It does not allow any additional columns.

The Pandas implementation can probably be improved further as I'm not the biggest Pandas pro.

### DIFF
--- a/components/caption_images/src/main.py
+++ b/components/caption_images/src/main.py
@@ -60,7 +60,7 @@ class CaptionImagesComponent(PandasTransformComponent):
         logger.info(f"Device: {self.device}")
 
         self.processor = BlipProcessor.from_pretrained(model_id)
-        self.model = BlipForConditionalGeneration.from_pretrained(model_id)
+        self.model = BlipForConditionalGeneration.from_pretrained(model_id).to(self.device)
 
         self.batch_size = batch_size
         self.max_new_tokens = max_new_tokens
--- a/components/segment_images/src/main.py
+++ b/components/segment_images/src/main.py
@@ -88,7 +88,7 @@ class SegmentImagesComponent(PandasTransformComponent):
         logger.info(f"Device: {self.device}")
 
         self.processor = SegformerImageProcessor.from_pretrained(model_id)
-        self.model = AutoModelForSemanticSegmentation.from_pretrained(model_id)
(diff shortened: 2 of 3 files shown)


components/caption_images/src/main.py
```diff
@@ -53,69 +52,39 @@ def caption_image_batch(
...
-            batch_size=batch_size,
-            max_new_tokens=max_new_tokens,
-            device=device,
-            meta={0: str}
+    def setup(self, *, model_id: str, batch_size: int, max_new_tokens: int) -> None:
+        self.device = "cuda" if torch.cuda.is_available() else "cpu"
+        logger.info(f"Device: {self.device}")
+
+        self.processor = BlipProcessor.from_pretrained(model_id)
+        self.model = BlipForConditionalGeneration.from_pretrained(model_id)
```
NielsRogge, line 63: It doesn't seem like the model is put on the `device`

components/segment_images/src/main.py
```diff
@@ -76,64 +75,42 @@ def segment_image_batch(image_batch: pd.DataFrame, *, model: AutoModelForSemanti
...
-            model=model,
-            processor=processor,
-            batch_size=batch_size,
-            device=device,
-            meta={0: object}
+        self.device = "cuda" if torch.cuda.is_available() else "cpu"
+        logger.info(f"Device: {self.device}")
+
+        self.processor = SegformerImageProcessor.from_pretrained(model_id)
+        self.model = AutoModelForSemanticSegmentation.from_pretrained(model_id)
```
NielsRogge, line 91: Same here, model needs to be put on the `device`

NielsRogge (approved):
Looks very nice!

components/caption_images/src/main.py
RobbeSneyders: Good catch. Seems like this wasn't done before either.

PR #221, opened 2023-06-20
[WIP] Add component type to spec

PhilippeMoussalli:
PR that implement a component type per spec as discussed [here](https://github.com/ml6team/fondant/pull/196/files#r1227898938) to explicitly define the kubeflow/fondant input and output arguments per component type. This is still work in progress, current changes include adding the necessary changes to the schema and component spec script. 

What still needs to be done: 

- [ ] Add `type` field to all the affected components 
- [ ] Add relevant tests

### DIFF
--- a/fondant/component_spec.py
+++ b/fondant/component_spec.py
@@ -6,6 +6,7 @@ import pkgutil
 import types
 import typing as t
 from dataclasses import dataclass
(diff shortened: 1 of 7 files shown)


RobbeSneyders:
Will close this for now. Feel free to reopen in the future.

PR #223, opened 2023-06-21
[DataComp pipeline] Add first 2 components

NielsRogge:
This PR adds the first two components of the DataComp pipeline, where it reuses the `load_from_hf_hub` component but with a different `fondant_component.yaml` file.


components/image_resolution_filtering/src/main.py
RobbeSneyders, line 1: I don't think this will work due to a couple of reasons. You are still using the `DaskTransformComponent`, but using the dataframe as if it is a Pandas dataframe. And the `PandasTransformComponent` doesn't receive its arguments in the `transform` method.

I already ported this component, but didn't push it yet, since I didn't test it yet:
```python
"""This component filters images of the dataset based on image size (minimum height and width)."""
import logging

import pandas as pd

from fondant.component import PandasTransformComponent
from fondant.logger import configure_logging

configure_logging()


class ImageFilterComponent(PandasTransformComponent):
    """Component that filters images based on height and width."""

    def setup(self, *, min_width: int, min_height: int) -> None:
        """
        Args:
            min_width: min width to filter on
            min_height: min height to filter on.
        """
        self.min_width = min_width
        self.min_height = min_height

        filtered_df = dataframe[
            (dataframe["images"]["width"] > self.min_width)
            & (dataframe["images"]["height"] > self.min_height)
        ]
        return filtered_df


if __name__ == "__main__":
    component = ImageFilterComponent.from_file()
    component.run()
```

examples/pipelines/datacomp/pipeline.py
```diff
@@ -0,0 +1,59 @@
...
+    "text": "text_data",
+    "clip_b32_similarity_score": "image_text_clip_b32_similarity_score",
+    "clip_l14_similarity_score": "image_text_clip_b32_similarity_score",
+}
+
+load_from_hub_op = ComponentOp.from_registry(
+    name="load_from_hub",
+    component_spec_path="components/load_from_hf_hub/fondant_component.yaml",
+    arguments={
+        "dataset_name": "ml6team/the-stack-smol-python",
```
RobbeSneyders, line 41: I guess this should be another dataset? :)

RobbeSneyders (commented):
Thanks @NielsRogge!

The pipeline is failing on pre-commit. If you haven't yet, you can run `pre-commit install` so it automatically runs on every commit.

examples/pipelines/datacomp/components/filter_text_complexity/src/main.py
```diff
@@ -0,0 +1,77 @@
...
+        min_complexity: int,
+        min_num_actions: int
+    ) -> None:
+        self.nlp = spacy.load(spacy_pipeline, exclude=["ner"])
+        self.batch_size = batch_size
+        self.min_complexity = min_complexity
+        self.min_num_actions = min_num_actions
+
+    def transform(self, dataframe: pd.DataFrame) -> pd.DataFrame:
+        texts = dataframe["text_data"]
```
RobbeSneyders, line 60: I think this needs to be:
```suggestion
        texts = dataframe["text"]["data"]
```

examples/pipelines/datacomp/docker-compose.yml
RobbeSneyders, line 1: This file should not be tracked in git.

RobbeSneyders (commented):
Some small comments. Can you address them and then merge the PR so you can open new PRs for additional components?

examples/pipelines/datacomp/components/filter_text_complexity/src/main.py
NielsRogge: Yes that was what I expected to receive but I get:

```
KeyError: 'text'
```
it turns out the columns are called `text_data`.

Are the hierarchical columns part of the latest release?

I also tried adding a main install of Fondant in the requirements, but no luck

NielsRogge: Fixed when using `docker compose up --build`

NielsRogge:
Local pipeline is running fine until writing of the data, where it gives:

```
ValueError: The columns in the computed data do not match the columns in the provided metadata
datacomp-filter_text_complexity-1  |   Extra:   ['text_data']
datacomp-filter_text_complexity-1  |   Missing: []
``` 
=> the `filter_text_complexity` component is just a filtering component, so it doesn't add any new columns, hence its produces section in the spec is empty. This is correct right?

NielsRogge:
Ok, I guess we should wait to merge this PR before that is resolved?

RobbeSneyders (approved):
Let's merge now and I'll update it in my PR.

PR #225, opened 2023-06-22
Leverage manifest instead of component spec in component data validation

RobbeSneyders:
This PR does 2 things
- It leverages the manifest instead of the component spec for data validation in the `PandasTransformComponent` class. This is better since the output manifest is evolved based on the input manifest and component spec, taking into account things such as `additionalSubsets` and `additionalFields`
- It filters out data not in the manifest from the pandas dataframe returned by the user. Previously, returning additional columns would raise an error. (see https://github.com/ml6team/fondant/pull/223#issuecomment-1602233930)

### DIFF
--- a/components/download_images/src/main.py
+++ b/components/download_images/src/main.py
@@ -127,8 +127,7 @@ class DownloadImagesComponent(PandasTransformComponent):
             axis=1,
             result_type="expand",
         )
-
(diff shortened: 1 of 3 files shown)


RobbeSneyders:
We might not want to use the manifest here after all, since we are validating user code, and they only need to return the code which is explicitly produced by the component. Will revert the manifest change and resubmit the fix for the strict validation.

PR #226, opened 2023-06-22
Don't validate returned Pandas dataframe strictly

RobbeSneyders:
This PR filters out data from the pandas dataframe returned by the user that is not defined in the component spec. Previously, returning additional columns would raise an error. (see https://github.com/ml6team/fondant/pull/223#issuecomment-1602233930)


NielsRogge (approved):
Thanks for improving!

PR #228, opened 2023-06-22
[DataComp pipeline] Add text spotting component

NielsRogge:
This PR adds a "text spotting" component as proposed in [Radenovic et al., 2023](https://arxiv.org/abs/2301.02280).

This component doesn't run on my Macbook though. It returns:
```
[2023-06-22 15:06:30,541 | root | INFO] Writing data...
datacomp-filter_text_spotting-1    | [W NNPACK.cpp:64] Could not initialize NNPACK! Reason: Unsupported hardware.
```
One thing to note is that sometimes URLs don't work, so we need a strategy to counter that. One option is to first run the "image downloader" component before this one, to make sure we only run it on images that could be downloaded.

### DIFF
--- a/examples/pipelines/datacomp/components/filter_text_spotting/Dockerfile
+++ b/examples/pipelines/datacomp/components/filter_text_spotting/Dockerfile
@@ -1,0 +1,21 @@
+FROM --platform=linux/amd64 python:3.8-slim
+
+## System dependencies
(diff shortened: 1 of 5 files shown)


NielsRogge:
Closing this one in favor of a new component that will leverage ONNX models for faster inference.

PR #229, opened 2023-06-22
Centralize logging configuration

RobbeSneyders:
I was getting double logs with the local runner and was never a fan of the `configure_logging` utility function which had to be called by the user. This sets the logging format and level centrally in a way that doesn't overwrite already configured handlers and can easily be overwritten for each separate logger by the user.


GeorgesLorre (approved):
makes sense, thx

NielsRogge (approved):

PR #230, opened 2023-06-23
[DataComp] Add image resolution filtering component

NielsRogge:
This PR adds an image resolution filtering component to the DataComp pipeline.

I had to use a symlink in order to build a Docker image and use the local runner.


RobbeSneyders:
FYI, instead of using a symlink, you can also just add a path to the top components directory directly.

NielsRogge:
@RobbeSneyders working now, had to update the requirements.txt to the latest commit as otherwise the Docker image wasn't updated. Feel free to approve

components/filter_image_resolution/src/main.py
```diff
@@ -0,0 +1,49 @@
...
+        width = dataframe["image"]["width"]
+        height = dataframe["image"]["height"]
+        min_image_dim = np.minimum(width, height)
+        max_image_dim = np.maximum(width, height)
+        aspect_ratio = max_image_dim / min_image_dim
+        mask = (min_image_dim >= self.min_image_dim) & (aspect_ratio <= self.max_aspect_ratio)
+
+        dataframe = dataframe[mask]
+
+        dataframe = dataframe.drop([("image", "width"), ("image", "height")], axis=1)
```
RobbeSneyders, line 42: Should no longer be needed after #233

examples/pipelines/datacomp/components/filter_text_complexity/requirements.txt
```diff
@@ -1,4 +1,4 @@
-git+https://github.com/ml6team/fondant.git@main
+git+https://github.com/ml6team/fondant.git@6ef449cabcd56a59b2d4405e82060b745f992560
```
RobbeSneyders, line 1: This was only needed for you locally to invalidate the cache. We don't need to add this to git. I'll open a PR soon that tries to tackle these issues.

RobbeSneyders (commented):
Thanks @NielsRogge!
Can you rebase on main? That will be needed to address one of my comments and will update the CI to run the latest version (I think I see some Ruff issues which are now checked on main).

RobbeSneyders (approved):
LGTM!

PR #232, opened 2023-06-23
[LLM pipeline] Language filter component

mrchtr:
This PR adds the first component for the LLM dataset creation pipeline. The component is a language filter which filters out rows in a provided dataframe that are not matching the provided language. 
FastText is used for the language detection. 

Changes
- add component
- add unit test to test the filter logic inside of the component

Note: Did not create a pipeline that uses this component yet.


components/language_filter/fondant_component.yaml
```diff
@@ -0,0 +1,14 @@
+name: Language filter
+description: A component that filter a provided dataframe based on the language.
+image: ghcr.io/ml6team/language_filter:latest
+
+consumes:
+  passages:
+    fields:
+      text:
+        type: string
```
NielsRogge, line 9:
```suggestion
  text:
    fields:
      data:
        type: string
```
In order to make the component reusable across different pipelines, we could use these names

components/language_filter/src/main.py
```diff
@@ -0,0 +1,70 @@
...
+            str: The predicted language label.
+        """
+        predictions = self.model.predict(text, k=1)
+        return predictions[0][0]
+
+    def is_language(self, row, language):
+        return language in self.predict_lang(row["text"])
+
+
+class LanguageFilterComponent(DaskTransformComponent):
```
NielsRogge, line 42:
```suggestion
class LanguageFilterComponent(PandasTransformComponent):
```
Feel free to upgrade to the Pandas interface (see caption_images as an example)

components/language_filter/src/main.py
```diff
@@ -0,0 +1,70 @@
...
+        """
+        Args:
+            dataframe: Dask dataframe.
+            language: Only keep text passages which are in the provided language
+
+        Returns:
+            Dask dataframe
+        """
+
+        lang_detector = LanguageIdentification()
```
NielsRogge, line 60: this can then go into a `setup()` method when using `PandasTransformComponent`

components/language_filter/src/main.py
mrchtr: Where is the difference between them? Isn't `DaskTransformComponent` the default way for transformation?

components/language_filter/fondant_component.yaml
```diff
@@ -0,0 +1,14 @@
+name: Language filter
+description: A component that filter a provided dataframe based on the language.
+image: ghcr.io/ml6team/language_filter:latest
```
NielsRogge, line 3: Nit: let's use consistent naming for all components, I'm currently using verb + noun, so this one can be named "Filter language"
```suggestion
name: Filter languages
description: A component that filters text based on the language.
image: ghcr.io/ml6team/filter_language:latest
```

components/language_filter/src/main.py
```diff
@@ -0,0 +1,73 @@
+"""A component that filter a provided dataframe based on the language."""
```
NielsRogge, line 1:
```suggestion
"""A component that filters text based on the language."""
```

components/language_filter/src/main.py
```diff
@@ -0,0 +1,73 @@
...
+    """Component that filter columns based on provided language."""
+
+    def setup(self, *args, **kwargs):
+        """Setup language filter component."""
+        self.lang_detector = LanguageIdentification()
+
+    def transform(
+            self,
+            dataframe: pd.DataFrame
+    ) -> pd.DataFrame:
```
NielsRogge, line 55: Probably fits on one line

components/language_filter/src/main.py
```diff
@@ -0,0 +1,73 @@
...
+        """Predict if text of a row is written in the defined language."""
+        return language in self.predict_lang(row["text"])
+
+
+class LanguageFilterComponent(PandasTransformComponent):
+    """Component that filter columns based on provided language."""
+
+    def setup(self, *args, **kwargs):
+        """Setup language filter component."""
+        self.lang_detector = LanguageIdentification()
```
NielsRogge, line 50: This method should take the language specified by the user as an argument, and pass it to the `lang_detector`

components/language_filter/src/main.py
```diff
@@ -0,0 +1,73 @@
...
+    ) -> pd.DataFrame:
+        """
+        Args:
+            dataframe: Pandas dataframe.
+            language: Only keep text passages which are in the provided language.
+
+        Returns:
+            Pandas dataframe
+        """
+        language = self.user_arguments["language"]
```
NielsRogge, line 64: See my comment above, can be done in the setup method

components/language_filter/src/main.py
```diff
@@ -0,0 +1,72 @@
...
+import pandas as pd
+
+from fondant.component import PandasTransformComponent
+from fondant.logger import configure_logging
+
+configure_logging()
+logger = logging.getLogger(__name__)
+
+
+class LanguageIdentification:
```
NielsRogge, line 14: rather than including the ftz file, can we load from the hub since FastText is now [hosted there](https://huggingface.co/facebook/fasttext-language-identification)?

just:
```
import fasttext
from huggingface_hub import hf_hub_download

model_path = hf_hub_download(repo_id="facebook/fasttext-language-identification", filename="model.bin")
model = fasttext.load_model(model_path)
```

components/language_filter/README.md
```diff
@@ -0,0 +1,7 @@
+# Language filter
+
+## Description
+This component is based on the `TransformComponent` and is used to filter a dataframe based on language. 
+It allows you to remove rows that do not match the provided language, thus providing a way to focus 
+on specific languages within your data.
```
NielsRogge, line 6: there's no need for the empy __init__.py

components/language_filter/src/main.py
mrchtr: What speaks against to include the ftz file in the repository? Alternative we could download the file during the image build process. Just want to avoid the situation, if some external dependencies can not be reached that the execution of the component will fail.

components/language_filter/src/main.py
```diff
@@ -0,0 +1,72 @@
+"""A component that filters text based on the language."""
+import logging
+
+import fasttext
+import pandas as pd
+
+from fondant.component import PandasTransformComponent
+from fondant.logger import configure_logging
+
+configure_logging()
```
RobbeSneyders, line 10: This is outdated. You can rebase on / merge with main and you'll get this for free.

components/language_filter/src/main.py
```diff
@@ -0,0 +1,72 @@
...
+        """
+        Args:
+            dataframe: Pandas dataframe.
+            language: Only keep text passages which are in the provided language.
+
+        Returns:
+            Pandas dataframe
+        """
+        mask = dataframe.apply(
+            lambda row: self.lang_detector.is_language(row), axis=1)
```
RobbeSneyders, line 65:
```suggestion
        mask = dataframe.apply(self.lang_detector.is_language, axis=1)
```
I think you can drop the lambda.

components/language_filter/src/main.py
```diff
@@ -0,0 +1,72 @@
...
+    def setup(self, *, language):
+        """Setup language filter component."""
+        self.lang_detector = LanguageIdentification(language)
+
+
+    def transform(self, dataframe: pd.DataFrame) -> pd.DataFrame:
+        """
+        Args:
+            dataframe: Pandas dataframe.
+            language: Only keep text passages which are in the provided language.
```
RobbeSneyders, line 59: This argument should move to the `setup` method.

components/language_filter/tests/language_filter_component_test.py
RobbeSneyders, line 1: Interesting to see these tests.

This could probably also be easier if we split the general component behavior from the user implementation into separate classes as discussed in chat. Since then we could test the user implementation without having to provide dummy variables for all the general component behavior.

components/language_filter/requirements.txt
```diff
@@ -0,0 +1,4 @@
+fondant
```
RobbeSneyders, line 1:
```suggestion
git+https://github.com/ml6team/fondant@main
```
Our build / release pipelines expect fondant to be installed from git.

RobbeSneyders (approved):

PR #234, opened 2023-06-23
Refactor local runner and add `run` command

GeorgesLorre:


fondant/cli.py
```diff
@@ -170,40 +68,198 @@ def pipeline_from_string(import_string: str) -> Pipeline:
...
+
+    def _help_or_parse(
+        self,
+        input_args: t.List[str],
+        parser: argparse.ArgumentParser,
+    ) -> t.Tuple[argparse.Namespace, t.List[str]]:
+        """Check if the next command is help and print help if it is.
+        Otherwise parse the known args.
+        """
+        if input_args[0] in ["help", "-h", "--help"]:
```
RobbeSneyders, line 92: Why don't we rely on the built-in behavior of `argparse` here?

fondant/cli.py
```diff
@@ -170,40 +68,198 @@ def pipeline_from_string(import_string: str) -> Pipeline:
...
+                "Docker runtime not found. Please install Docker and try again.",
+            )
+
+        run_explorer_app(**vars(args))
+
+    def compile(self, input_args):
+        parser = argparse.ArgumentParser(add_help=False)
+
+        parser.add_argument(
+            "mode",
```
RobbeSneyders, line 172: I think it's a more logical user interface if this is a flag instead of a subcommand.

So:
```
fondant compile <pipeline> --local
```
instead of
```
fondant compile local <pipeline>
```

I think this can be achieved in the current implementation. We might just have to lift the `pipeline` parameter from `compile_{mode}` to this function, which I think makes sense.

fondant/cli.py
```diff
@@ -170,40 +68,198 @@ def pipeline_from_string(import_string: str) -> Pipeline:
...
+
+    def compile_kubeflow(self, input_args):
+        msg = "Kubeflow compiler not implemented"
+        raise NotImplementedError(msg)
+
+    def run(self, input_args):
+        parser = argparse.ArgumentParser(add_help=False)
+
+        parser.add_argument(
+            "mode",
```
RobbeSneyders, line 216: Same comment here as above.

RobbeSneyders (commented):
Thanks @GeorgesLorre!

Can you explain a bit why you switched to this approach? One downside I see is that we're no longer building the parser hierarchy up front, which might not allow for cli autocompletion (to be checked).

fondant/cli.py
GeorgesLorre: The `parse_known_args()` will always consume the help. So if you do `fondant compile local --help` you will get the help page for just fondant.

GeorgesLorre:
> Thanks @GeorgesLorre!
> 
> Can you explain a bit why you switched to this approach? One downside I see is that we're no longer building the parser hierarchy up front, which might not allow for cli autocompletion (to be checked).

Because of multiple subcommand levels `fondant(1) compile(2) local(3)` (see our chat). But we could change it back and build the hierarchy up front.

fondant/cli.py
```diff
@@ -130,17 +117,123 @@ def explore(args):
...
+        help="Extra volumes to mount in containers",
+        nargs="+",
+    )
+    parser.set_defaults(func=run)
+
+
+def run(args):
+    if args.local:
+        try:
+            pipeline = pipeline_from_string(args.ref)
```
RobbeSneyders, line 209: We only need the try except around this line I think.

fondant/cli.py
```diff
@@ -28,91 +27,77 @@
...
+    parser = argparse.ArgumentParser(description="Fondant CLI")
+    subparsers = parser.add_subparsers()
+    register_explore(subparsers)
+    register_compile(subparsers)
+    register_run(subparsers)
+
     sys.path.append(".")
-    args = cli.parse_args()
+
+    args = parser.parse_args(sys.argv[1:])
```
RobbeSneyders, line 45:
```suggestion
    args = parser.parse_args()
```
I don't think explicitly providing the arguments is necessary.

RobbeSneyders (commented):
I like the `--local` and `--kubeflow` flags instead of subcommands as the user interface. But can we now still define separate flags for each of them?

README.md
```diff
@@ -117,9 +118,15 @@ pip install git+https://github.com/ml6team/fondant.git
 
 ### 🧱 Deploying Fondant
 
-Fondant is flexible and works by leveraging [kubeflow pipelines](https://www.kubeflow.org/docs/components/pipelines/v1/introduction/) on any kubernetes cluster. All fondant needs is an url pointing to the kubeflow pipeline host and a Object Storage provider (S3, GCS, etc) to store data produced in the pipeline between steps.
 
-We have compiled some references and created some scripts to [get you started](https://fondant.readthedocs.io/en/latest/infrastructure) with setting up the required infrastructure.
+There are 2 ways of using fondant:
+
+- Leveraging [kubeflow pipelines](https://www.kubeflow.org/docs/components/pipelines/v1/introduction/) on any kubernetes cluster. All fondant needs is an url pointing to the kubeflow pipeline host and a Object Storage provider (S3, GCS, etc) to store data produced in the pipeline between steps.
```
ChristiaensBert, line 124:
```suggestion
- Leveraging [kubeflow pipelines](https://www.kubeflow.org/docs/components/pipelines/v1/introduction/) on any kubernetes cluster. All that fondant needs is an url pointing to the kubeflow pipeline host and an Object Storage provider (S3, GCS, etc) to store data produced in the pipeline between steps.
```

README.md
```diff
@@ -117,9 +118,15 @@ pip install git+https://github.com/ml6team/fondant.git
...
+  We have compiled some references and created some scripts to [get you started](https://fondant.readthedocs.io/en/latest/infrastructure) with setting up the required infrastructure.
+- Or locally by using [docker compose](https://docs.docker.com/compose/). This way is mainly aimed at helping you develop fondant pipelines and components faster by making it easier to run things on a smaller scale.
+
+
+The same pipeline can be used in both variant allowing you to quickly develop and iterate using the local docker compose implementation and then using the power of kubeflow pipelines to run a large scale pipeline.
```
ChristiaensBert, line 129:
```suggestion
The same pipeline can be used in both variants, allowing you to quickly develop and iterate using the local docker compose implementation and then using the power of kubeflow pipelines to run a large scale pipeline.
```

README.md
```diff
@@ -212,6 +219,29 @@ class ExampleComponent(TransformComponent):
         """
 ```
 
+
+###  Running your pipeline
+
+Once you have a pipeline you can easily run (and compile it) using the builtin CLI:
```
ChristiaensBert, line 225:
```suggestion
Once you have a pipeline you can easily run (and compile) it by using the built-in CLI:
```

README.md
```diff
@@ -212,6 +219,29 @@ class ExampleComponent(TransformComponent):
...
+
+```bash
+fondant run foo/bar.py:pipeline --local
```
ChristiaensBert, line 228: This is for the --local, should we add both?

RobbeSneyders, line 228:
```suggestion
fondant run foo.bar:pipeline --local
```
You should pass the file as a module, right?

fondant/cli.py
RobbeSneyders: Bump

RobbeSneyders (approved):
Thanks @GeorgesLorre! 2 small comments.

README.md
NielsRogge, line 124:
```suggestion
- Leveraging [Kubeflow pipelines](https://www.kubeflow.org/docs/components/pipelines/v1/introduction/) on any Kubernetes cluster. All Fondant needs is a url pointing to the Kubeflow pipeline host and an Object Storage provider (S3, GCS, etc) to store data produced in the pipeline between steps.
```

README.md
```diff
@@ -117,9 +118,15 @@ pip install git+https://github.com/ml6team/fondant.git
...
 
-We have compiled some references and created some scripts to [get you started](https://fondant.readthedocs.io/en/latest/infrastructure) with setting up the required infrastructure.
+There are 2 ways of using fondant:
+
+- Leveraging [kubeflow pipelines](https://www.kubeflow.org/docs/components/pipelines/v1/introduction/) on any kubernetes cluster. All fondant needs is an url pointing to the kubeflow pipeline host and a Object Storage provider (S3, GCS, etc) to store data produced in the pipeline between steps.
+  We have compiled some references and created some scripts to [get you started](https://fondant.readthedocs.io/en/latest/infrastructure) with setting up the required infrastructure.
+- Or locally by using [docker compose](https://docs.docker.com/compose/). This way is mainly aimed at helping you develop fondant pipelines and components faster by making it easier to run things on a smaller scale.
```
NielsRogge, line 126:
```suggestion
- Or locally by using [Docker Compose](https://docs.docker.com/compose/). This way is mainly aimed at helping you develop Fondant pipelines and components faster by making it easier to run things on a smaller scale.
```

README.md
NielsRogge, line 129:
```suggestion
The same pipeline can be used in both variant allowing you to quickly develop and iterate using the local Docker Compose implementation and then using the power of Kubeflow pipelines to run a large scale pipeline.
```

fondant/cli.py
GeorgesLorre: I could move the logic around but I believe this is the clearest. Try to see if it is an importstring if it is then compile it otherwise assume it is a reference to an already compiled pipeline.

README.md
GeorgesLorre: the kubeflow command is there but raises NotImplementedError's

fondant/cli.py
RobbeSneyders: You can use 
```
try:
    ...
except:
    ...
else:
    ...
finally:
    ...
```

README.md
```diff
@@ -212,6 +219,29 @@ class ExampleComponent(TransformComponent):
         """
 ```
 
+
+###  Running your pipeline
+
+Once you have a pipeline you can easily run (and compile) it by using the built-in CLI:
+
+```bash
+fondant run foo.bar.py:pipeline --local
```
RobbeSneyders, line 228:
```suggestion
fondant run foo.bar:pipeline --local
```
The `.py` should also be removed.

fondant/runner.py
```diff
@@ -0,0 +1,18 @@
...
+    def run(self, *args, **kwargs):
+        """Abstract method to invoke running."""
+
+
+class DockerRunner(Runner):
+    def run(cls, input_spec: str, *args, **kwargs):
+        """Run a docker-compose spec."""
+        cmd = ["docker", "compose", "-f", input_spec, "up", "--build"]
+
+        subprocess.call(cmd, stdout=subprocess.PIPE)  # nosec
```
RobbeSneyders, line 18:
```suggestion
        subprocess.call(cmd)  # nosec
```
As discussed, this intercepts the build logs and docker compose log colors.

PR #235, opened 2023-06-27
Enforce binding of absolute path

mrchtr:
If someone initializes the pipeline with a local base path and uses a relative path, `docker-compose up` will fail because the volume bindings require an absolute path.

To avoid this, I propose resolving the absolute path during the generation of the `docker-compose.yaml`.


GeorgesLorre (approved):
Nice

PR #236, opened 2023-06-28
Build dev images on main

### LINKED ISSUE #220: Build components on main
It's currently hard to test framework changes in the components since they are only built on release. We can instead build components on every merge to master and tag them with the related commit.

The build pipeline currently already takes 23 minutes, which is too long to run on every merge. We should leverage docker caching where possible.

There's [multiple options](https://docs.docker.com/build/ci/github-actions/cache/) to leverage docker caching in github actions. The Github Actions cache is [limited to 10GB](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy) which is too low for us since we have a lot of images, some of which are quite large. Instead, we can leverage the docker registry with in-line caching.

RobbeSneyders:
Fixes #220 

This PR adds a github actions workflow that builds our reusable images on every merge to main. This should mitigate some of the issues we're having with outdated fondant versions in the components. 

Each image is tagged with the associated commit and we move the `dev` tag to always point to the latest images build from main. This is now the default version included in the `fondant_component.yaml` of the reusable components. Just note that your runner might not recognize that the image associated with the `dev` tag has changed, and you might have to force it to pull the latest version.

The workflow uses [registry caching](https://docs.docker.com/build/cache/backends/registry/) to cache the docker images, which brings down the build time from [~40m](https://github.com/ml6team/fondant/actions/runs/5398796173) to [~20m](https://github.com/ml6team/fondant/actions/runs/5399309438). I opted for registry caching as local github action caching is limited to 10GB, which would be too limited for the amount and size of images we're building.

We can look into some additional improvements to this:
- [ ] **Direct the runners to always pull images**
This will introduce a small delay, but might be worth it to prevent hard to detect issues because of outdated images.
- [ ] **Build fondant from local path**
The workflow introduced in this PR still requires changes to be merged to main before they are built into the images. Especially while developing on fondant itself, this might still be lacking in usability, as we'd like to be able to test our local changes. This is possible by passing an additional [`--build-context`](https://docs.docker.com/engine/reference/commandline/buildx_build/#build-context), but we'd have to integrate this into the local runner to really be useful, without impacting the user experience.
- [ ] **Cache pip**
We can also leverage pip caching on top of / instead of docker caching. This wouldn't add a lot of benefit on top of the current approach, as the time spent on downloading packages and building wheels is negligible compared to the time spent communicating with the registry cache. It could be useful in allowing us to include the fondant dependency in the `requirements.txt` again, which would be a bit more transparent. Then the "Install dependencies" layer would no longer be cached by docker, since we change the fondant version on each build, but the other dependencies would be cached using pip. I believe this would still increase the build time though as the cached packages still need to be installed. To leverage the pip cache across containers, we can use a [cache mount](https://docs.docker.com/build/guide/mounts/#add-a-cache-mount).


.github/workflows/build.yaml
```diff
@@ -0,0 +1,29 @@
+name: Build dev images
+
+on:
+  push:
+    branches:
+      - "feature/main-build"
```
RobbeSneyders, line 6: This needs to be changed to `main` before merging.

GeorgesLorre:
the `dev` tag being the default, will that not cause potential issues for our users ?

We want to be able to develop en test images while still having a stable older version available. I would still make the `latest` tag the default in all reusable components (maybe `latest` is then a bit of a misnomer). That way developers can switch to `dev` in order to develop but users are guaranteed to have stable images.

RobbeSneyders:
The `dev` tag is the default when developing from source. If you install fondant from PyPi, the version of fondant in the images is the same version you installed.

`latest` points to the latest released stable / released version.

scripts/build_components.sh
```diff
@@ -37,20 +39,40 @@ for dir in "$component_dir"/*/; do
...
+    full_image_names+=("$full_image_name")
+  done
+  # Prevent this from mistakenly being used below
+  unset full_image_name
 
-  # Update the fondant requirement to the version being built
-  sed -i "s|fondant@main|fondant@$tag|" requirements.txt
+  echo "Updating the image version in the fondant_component.yaml with:"
+  echo "${full_image_names[0]}"
+  sed -i "s|^image: .*|image: ${full_image_names[0]}|" fondant_component.yaml
```
GeorgesLorre, line 54: So here you update the images, how does this make it into main ?

RobbeSneyders: It doesn't. This actually doesn't have an effect when triggered on main, only when releasing, as the updated component specifications are included in the `fondant` package. You could argue that it shouldn't be in this script, but then we need to create the `full_image_name`s in multiple places.

Come to think of it, I do think we have an issue in the release pipeline now, as this script only runs when triggered by a tag, not by a release. So the component specs are only updated in the package pushed to `test.PyPi`, not to the actual `PyPi`.

RobbeSneyders: Opened #255 for this.

GeorgesLorre: Ok makes sense, then this PR is a nice upgrade

GeorgesLorre (approved):
Don't forget to update the branch name

RobbeSneyders:
> **Direct the runners to always pull images**
> 
> I think this will always be the case for KFP since it does not cache any docker images. We might need to change that for the local runner. Does the `--build` flag for the docker compose force the rebuilding of all the images?

It only builds custom components, while this PR is focused on reusable components. I opened #259 for this, but note that this is only necessary if you're working from source. If you install `fondant` from PyPI, the component version is fixed to the released version, so this won't be an issue. So this would mostly be useful for our own development, less for users.

> **Build fondant from local path**
> 
> We can expand this to also build components from current changes so that we don't have to wait until they are merged to be able to test them. I think this is currently implement for the local runner for custom components but is not the case for reusable ones. For the remote runner (KFP) its currently hard to test things out properly since you always have to rely on merging your components to pull them from the remote registry. Remote testing is useful for testing things that are hard to test locally (e.g. GPU components).

Building from local path indeed means building the local changes to fondant into the image. And of course also the local changes to the component itself. The local runner currently builds custom components, but it will not include local fondant changes. For this we would need to change the Dockerfile to leverage additional build context and then pass it with the runner.

> One solution for this is to maybe modify some of the script in the `scripts` folder to enable only building certain components and giving them the `dev` tag as well. wdyt?

You should already be able to run the `build_components.sh` script locally, but you **should not** use the `dev` tag when running this manually, as it should always point to `main`. You can run it with any other tag though and it should both build the images and update the component specifications to use the new version.

Issue #180, opened 2023-06-02
Map subset names between components


PhilippeMoussalli:
### Problem Statement

To make the pre-made components more useable we need a good way to specify your own columns without having to change the component specs, component code and rebuild the Docker image.

This applies for all component types:
* `LoadComponent`: user might want to change the default column names of the source dataset 
* `WriteComponent`: user might want to rename final written dataset differently then `subset_field`
* `TransformComponent`: user may want to introduce their own map component that produces columns of certain field names but then want to use a resuable component with different field/subset names. In that case, they will need to remap the column names. 

 
### Proposed Approach

[Bert]: I would suggest having support for passing a `renaming dictionary` which contains a mapping of the columns in the spec and the columns that you want to use instead.

Suppose that the `fondant_spec` contains these columns
```yaml
consumes:
  images:
    fields:
      data:
        type: binary

produces:
  images:
    fields:
      data:
        type: binary
      width:
        type: int16
      height:
        type: int16
```

but you want `pictures_content` and `pictures_w`, `pictures_h` because your previous component produces this.

```python

column_mapping = {
                                  'images_data': 'pictures_content',
                                  'images_width': 'pictures_w',
                                  'images_height': 'pictures_h'
                                  }

caption_images_op = ComponentOp.from_registry(
    name="<DOCKER_IMAGE>",
    arguments={},
    number_of_gpus=1,
    node_pool_name="model-inference-pool",
    column_mapping=column_mapping
)
```

While reading/writing the parquet file, Fondant can use this mapping to retrieve/write the correct subsets.

[Philippe]: this might work but can introduce a lot of complexity -> we might want to introduce a taxonomy.

There are advantages or disadvantages to both. I created a schema to make this a bit more clear -> [to Taxonomy or not to Taxonomy?](https://excalidraw.com/#room=8b0695dbe8fb8594a53b,EXnYA0oCzOABBOYyZY2oNQ)

**Taxonomy free approach:** 

* Assumes the users know the one-to-one mapping between fields/subsets produced/consumed by custom components and their equivalent in the reusable components (and vice-versa) 

* Fields under subsets may switch between components -> undesired behavior may occur 

 * User needs to track how fields/subset names are changed throughout the pipeline and  take extra consideration when defining the renaming dictionary (additional work at the user's end)

**Taxonomy approach:** 

* Less complexity since users have to conform to a taxonomy -> no need to track how subsets/fields are changing between components -> less unpredictable behavior  

[...]

PR #239, opened 2023-06-28
[DataComp] Add cluster component

NielsRogge:
This PR adds a cluster component to the DataComp pipeline.

It leverages sklearn's [mini batch k-means](https://scikit-learn.org/stable/modules/generated/sklearn.cluster.MiniBatchKMeans.html). However I'm not sure this scales to billions of images, as for that you need a GPU.

So I guess we can also have a separate component that leverages [faiss-gpu](https://github.com/facebookresearch/faiss/wiki/Faiss-on-the-GPU) for those who have the resources.

Edit; switched to subsampled k-means.

To do:

- [ ] I see that when adding a column to a subset, all existing fields currently need to be included in the `fondant_component.yaml` file as they otherwise aren't included anymore. 
- [ ] Currently I have an "image" subset with an "embedding" column. However other components consider embeddings to be a separate subset.


RobbeSneyders:
> I see that when adding a column to a subset, all existing fields currently need to be included in the fondant_component.yaml file as they otherwise aren't included anymore.

This is indeed not working as specified. Something to fix in Fondan.

examples/pipelines/datacomp/components/cluster_image_embeddings/src/main.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def transform(self, dataframe: pd.DataFrame) -> pd.DataFrame:
+        """
+        Args:
+            dataframe: Pandas dataframe
+
+        Returns:
+            Pandas dataframe
+        """
+        embeddings = dataframe["image"]["embedding"]
```
mrchtr, line 30: Regarding your question in the daily: Would there be any option to load the column names in generic way? If the component always consumes a dataframe with one subset/column the naming shouldn't matter?

components/filter_image_resolution/requirements.txt
```diff
@@ -1,3 +1,3 @@
-git+https://github.com/ml6team/fondant@main
+git+https://github.com/ml6team/fondant.git@main
```
RobbeSneyders, line 1: I don't think this is needed? It works without this for me.

components/load_from_hf_hub/fondant_component.yaml
RobbeSneyders, line 1: To use this, you should just add the `fondant_component.yaml` as a custom component:

```
|- examples/pipelines/datacomp
   |- components
      |- load_datacomp_from_hf_hub
         |- fondant_component.yaml
```
and just refer to the reusable docker image in the component spec.

components/load_from_hf_hub/requirements.txt
RobbeSneyders, line 1: You should not have to edit this generic reusable component.

examples/pipelines/datacomp/components/cluster_image_embeddings/src/main.py
RobbeSneyders: Good question. I see 2 mechanisms we could leverage:
- Generic components. These read their data based on the component specification, which is the only thing the user needs to provide. This is already used for the `load_from_hf_hub` and `write_to_hf_hub` components.
- Column mapping between components. We don't have this yet, but are planning to implement this in #180 

Not sure how easy it would be to implement this component in a generic way, but it can be a good test.

examples/pipelines/datacomp/components/filter_image_resolution
RobbeSneyders, line 1: Should not be committed. See my suggestion below on how you can achieve the same behavior without symlinks.

examples/pipelines/datacomp/components/load_from_hf_hub
RobbeSneyders, line 1: Should not be committed, see other comments.

examples/pipelines/datacomp/components/load_from_hf_hub/fondant_component.yaml
RobbeSneyders, line 1: This was the correct way to do it. Was there something not working with this approach?

RobbeSneyders (commented):
> This is indeed not working as specified. Something to fix in Fondant.

Created #244 for this.

RobbeSneyders:
I'm not sure that the current implementation is the best approach. It basically comes down to this:

```python
kmeans = init_kmeans()
labels = []
for partition in dataset:
    kmeans.partial_fit(partition)
    labels += kmeans.predict(partition)
```

Since we are updating the clusters on each partition, I don't think the labels assigned across partitions are consistent.

The algorithm I had in mind is:

```python
kmeans = init_kmeans()
sample = dataset.sample()
kmeans = kmeans.fit(sample)
labels = []
for partition in dataset:
    labels += kmeans.predict(partition)
```

For which we would need to leverage the `DaskTransformComponent` instead.

components/load_from_hf_hub/fondant_component.yaml
NielsRogge: Yes, although that doesn't work when I wanted to rebuild the image as I was updating this component to also produce embeddings.

NielsRogge:
> Since we are updating the clusters on each partition, I don't think the labels assigned across partitions are consistent.

Yes that's right, actually we could have a `post_transform` method (similar to the `setup` method - we could rename setup to `pre_transform` for consistency) which does things after transforming the data. `kmeans.predict` should only happen after all mini batches are fitted. This way we would have:

```
kmeans = init_kmeans()
labels = []
# first fit on all partitions
for partition in dataset:
    kmeans.partial_fit(partition)

# next assign labels to all partitions
for partition in dataset:
    kmeans.predict(partition)
```

The algorithm that you are proposing also seems a valid option, but that's not mini batch kmeans right? That's more like subsampled k-means :D

components/load_from_hf_hub/fondant_component.yaml
RobbeSneyders: The component should be implemented generically, so you should be able to control the data it produces using arguments only.

RobbeSneyders:
> Yes that's right, actually we could have a `post_transform` method (similar to the `setup` method - we could rename setup to `pre_transform` for consistency) which does things after transforming the data. `kmeans.predict` should only happen after all mini batches are fitted. This way we would have:
> 
> ```
> kmeans = init_kmeans()
> labels = []
> # first fit on all partitions
> for partition in dataset:
>     kmeans.partial_fit(partition)
> 
> # next assign labels to all partitions
> for partition in dataset:
>     kmeans.predict(partition)
> ```

The difference with `.setup()` is that you would need to iterate through the dataset again for this implementation of `.post_transform()`, which I don't think is a logical general mechanism for a component class. We can take this approach of iterating twice over the dataset using the `DaskTransformComponent`, but it's more expensive than what I proposed.

> The algorithm that you are proposing also seems a valid option, but that's not mini batch kmeans right? That's more like subsampled k-means :D

Yes indeed, but I just shared `miniBatchKMeans` as an interesting resource. We don't need to use it, but figure out the best solution.

mrchtr:
> I'm not sure that the current implementation is the best approach. It basically comes down to this:
> 
> ```python
> kmeans = init_kmeans()
> labels = []
> for partition in dataset:
>     kmeans.partial_fit(partition)
>     labels += kmeans.predict(partition)
> ```
> 
> Since we are updating the clusters on each partition, I don't think the labels assigned across partitions are consistent.
> 
> The algorithm I had in mind is:
> 
> ```python
> kmeans = init_kmeans()
> sample = dataset.sample()
> kmeans = kmeans.fit(sample)
> labels = []
> for partition in dataset:
>     labels += kmeans.predict(partition)
> ```
> 
> For which we would need to leverage the `DaskTransformComponent` instead.

I think this approach is a better fit. I even though about a step further. 
Agree with fitting clusters on a subset and predict for each entry in each partition a specific cluster. 
Afterwards, I would re-partition the DaskDataframe (with the naive hope that this can be done easily) in a way that each cluster is related to one partition. If that is possible, we can use a PandasTransformComponent to remove duplicates in each partition.

RobbeSneyders:
> I think this approach is a better fit. I even though about a step further.
> Agree with fitting clusters on a subset and predict for each entry in each partition a specific cluster.
> Afterwards, I would re-partition the DaskDataframe (with the naive hope that this can be done easily) in a way that each cluster is related to one partition. If that is possible, we can use a PandasTransformComponent to remove duplicates in each partition.

I think that's indeed a good approach, but I don't think we need to split it across multiple components. If you split it up, you would end up with a `PandasTransformComponent` which deduplicates per partition, which might not be very reusable. Instead, I would propose a single `DaskTransformComponent` which does (pseudocode):

```python
kmeans = init_kmeans()
sample = dataframe.sample()  # Dask has sample functionality
kmeans = kmeans.fit(sample)

dataframe = dataframe.apply/map_partitions(kmeans.predict)  # Call predict per row or partition

---

dataframe = dataframe.set_index(label, divisions=kmeans.labels)  # Set labels as index, pass list of labels as divisions for optimization

dataframe = dataframe.map_partitions(deduplicate)  # Deduplicate per partition (=cluster)
```

You could split it up at the `---`, but then you wouldn't be able to pass in the kmeans labels as optimization to `set_index`.

examples/pipelines/datacomp/components/load_from_hf_hub
RobbeSneyders, line 1: Should not be included.

RobbeSneyders (commented):
Thanks @NielsRogge! Have you been able to test this?

NielsRogge:
Yes @RobbeSneyders I confirm that the component works.

NielsRogge:
@RobbeSneyders when running this pipeline, the final dataframe of the `image` subset looks as follows:
```
>>> df.head()
    cluster_label
id               
14              0
25              1
38              2
41              0
43              2
```

however I'd like to just add `cluster_label` as additional column, keeping the existing ones. Should I change the `fondant_component.yaml`? The manifest.json of the cluster_embeddings component says that all these fields are still present:

```
{"metadata":
{"base_path": "/fondant_artifacts_datacomp", 
"run_id": "Datacomp filtering pipeline-20230705140722", 
"component_id": "cluster_embeddings"}, 

"index": {"location": "/index/Datacomp filtering pipeline-20230705140722/cluster_embeddings"},

"subsets":
{"image": 
 {"location": "/image/Datacomp filtering pipeline-20230705140722/cluster_embeddings", 
 "fields": {"url": {"type": "string"}, "width": {"type": "int16"}, "height": {"type": "int16"},
 "face_bboxes": {"type": "array", "items": {"type": "array", "items": {"type": "float32"}}},
 "sha256": {"type": "string"}, "embedding": {"type": "array", "items": {"type": "float32"}},
 "cluster_label": {"type": "int16"}}},
 
 "text": {"location": "/text/Datacomp filtering pipeline-20230705140722/load_from_hub",
 "fields": {"data": {"type": "string"}}},
 
 "image_text": {"location": "/image_text/Datacomp filtering pipeline-20230705140722/load_from_hub",
 "fields": {"clip_b32_similarity_score": {"type": "float32"}, "clip_l14_similarity_score":
 {"type": "float32"}}}}}
```

RobbeSneyders (commented):
Thanks @NielsRogge!

LGTM minus 2 minor comments.

RobbeSneyders (approved):

PR #246, opened 2023-06-29
[LLM pipeline] Add normalize text component

mrchtr:
Component which applies different text normalization (nfc, lowercasing and regex pattern replacements)

This component is needed for the LLM dataset creation pipeline.


components/text_normalization/fondant_component.yaml
```diff
@@ -0,0 +1,20 @@
+name: Normalize text.
+description: A component that normalizes text.
+image: ghcr.io/ml6team/text_normalization:latest
```
NielsRogge, line 3:
```suggestion
image: ghcr.io/ml6team/normalize_text:latest
```

components/text_normalization/requirements.txt
```diff
@@ -0,0 +1,3 @@
+git+https://github.com/ml6team/fondant.git@main
```
NielsRogge, line 1:
```suggestion
fondant
```
As there was a new Fondant release, this should work

components/text_normalization/fondant_component.yaml
RobbeSneyders: As mentioned in chat, agree with consistent naming, but would actually prefer `text_normalization` here as it will group all text components alphabetically.

components/text_normalization/requirements.txt
RobbeSneyders: We should keep the install form git, since our build and release pipelines expect it.

See https://github.com/ml6team/fondant/pull/238.

components/text_normalization/src/main.py
```diff
@@ -0,0 +1,67 @@
...
+import re
+import unicodedata
+from typing import List
+
+import pandas as pd
+
+from fondant.component import PandasTransformComponent
+from fondant.logger import configure_logging
+
+configure_logging()
```
RobbeSneyders, line 12: This is outdated, please rebase on / merge with main.

RobbeSneyders (commented):
Thanks @mrchtr! One small change is needed to the logging, otherwise LGTM.

components/text_normalization/fondant_component.yaml
NielsRogge: Hmm not necessarily agree on that one, we can group them ourselves in the docs?

Would prefer to have verb + noun for all our components

components/text_normalization/requirements.txt
NielsRogge: Ok, will close this comment

RobbeSneyders (approved):
Thanks @mrchtr!

PR #247, opened 2023-06-29
[LLM pipeline] Add filter out short texts component

mrchtr:
This component filter out short text passages. Text length can be passed as argument. 

This component is needed for the LLM dataset creation pipeline.


components/text_length_filter/src/main.py
```diff
@@ -0,0 +1,39 @@
+"""A component that filters out text passages which are to short."""
+import logging
+
+import pandas as pd
+
+from fondant.component import PandasTransformComponent
+from fondant.logger import configure_logging
+
+configure_logging()
```
RobbeSneyders, line 9: This is outdated. Please rebase on / merge with main.

components/text_length_filter/src/main.py
```diff
@@ -0,0 +1,39 @@
...
+        Filter out text passages which are to short.
+
+        Args:
+            dataframe: Pandas dataframe.
+
+        Returns:
+            Pandas dataframe.
+        """
+        mask = (
+                dataframe["text"]["data"].apply(lambda x: len(x.split()))
```
RobbeSneyders, line 30: This filters on the amount of words, right? This isn't really clear from the documentation of the components or arguments. I would expect `min_length` to define the number of characters based on the current description.

components/text_length_filter/fondant_component.yaml
```diff
@@ -0,0 +1,14 @@
+name: Filter text length
+description: A component that filter out text passages which are to short
```
NielsRogge, line 2:
```suggestion
description: A component that filters out text based on their length
```

components/text_length_filter/src/main.py
NielsRogge: Yes as an FYI the DataComp competition filters text based on 2 things: minimum amount of characters and minimum amount of words: https://github.com/mlfoundations/datacomp/blob/3ac587eefa16e57e4494f18024894fa642cd1a9f/baselines/apply_filter.py#L67-L75

Maybe we can include the same functionality here.

components/text_length_filter/Dockerfile
RobbeSneyders, line 1: The dockerfiles in our reusable components have changed. Can you have a look and copy the same one here?

components/text_length_filter/requirements.txt
RobbeSneyders, line 1: Once the Dockerfile is updated, both the `fondant` and `gcsfs` dependencies can be removed here.

RobbeSneyders (commented):
Thanks @mrchtr!

Looks like you'll need to update this branch, as there's currently some conflicts on code that I don't think should be included in this PR (the `language_filter` component).

Apart from that and some comments to update the component structure to the latest version, it looks good to me.

PR #250, opened 2023-06-29
Add getting started documentation

GeorgesLorre:


docs/getting_started.md
```diff
@@ -0,0 +1,169 @@
...
+        'column_name_mapping': {
+            'image': 'images_raw',
+        },
+        "image_column_names": ["image"],
+        
+    }
+)
+
+my_pipeline.add_op(load_from_hf_hub, dependencies=[])
+```
```
GeorgesLorre, line 70: Not sure if it is the best example since it seems quite complex

docs/getting_started.md
```diff
@@ -0,0 +1,169 @@
+
+
+## Setting up the environment
+
+### Installing fondant
+
+We suggest that you use a [virtual environment](https://docs.python.org/3/library/venv.html) for you project. Fondant supports Python >=3.8.
```
shayorshay, line 7:
```suggestion
We suggest that you use a [virtual environment](https://docs.python.org/3/library/venv.html) for your project. Fondant supports Python >=3.8.
```

docs/getting_started.md
```diff
@@ -0,0 +1,169 @@
...
+    pipeline_name='my_pipeline',
+    base_path='/home/username/my_pipeline',
+    pipeline_description='This is my pipeline',
+)
+```
+
+This is all you need to initialize a fondant pipeline:
+
+- **pipeline_name**: A name to reference your pipeline.
+- **base_path**: The base path of your pipeline, this will be used to store artifacts: the data between steps and the pipeline manifests. This base_path can be a local path or a cloud storage path (e.g. s3://my_bucket/artifacts, or gs://my_bucket/artifacts).
```
RobbeSneyders, line 41:
```suggestion
- **base_path**: The base path that fondant should use to store artifacts and data. This base_path can be a local path or a cloud storage path (e.g. s3://my_bucket/artifacts, or gs://my_bucket/artifacts).
```
Some simplification

docs/getting_started.md
```diff
@@ -0,0 +1,169 @@
...
+- **pipeline_description**: A description of your pipeline.
+
+## Adding components
+
+Now that we have a pipeline, we can add components to it. Components are the building blocks of your pipeline. They are the individual steps that will be executed in your pipeline. There are 2 main types of components:
+
+- reusable components: These are components that are already created by the community and can be easily used in your pipeline. You can find a list of reusable components [here](https://github.com/ml6team/fondant/tree/main/components). They often have arguments that you can set to configure them for your use case.
+
+- custom components: These are the components you create to solve your use case. A custom component can be easily created by adding a `fondant_component.yaml`, `dockerfile` and `main.py` file to your component subdirectory. This file contains the metadata of your component. You can find more information about the `fondant_component.yaml` file [here](https://github.com/ml6team/fondant/blob/main/docs/component_spec.md)
```
RobbeSneyders, line 50:
```suggestion
- custom components: These are the components you create to solve your use case. A custom component can be easily created by adding a `fondant_component.yaml`, `dockerfile` and `main.py` file to your component subdirectory. The `fondant_component.yaml` file contains the specification of your component. You can find more information about it [here](https://github.com/ml6team/fondant/blob/main/docs/component_spec.md)
```

docs/getting_started.md
```diff
@@ -0,0 +1,169 @@
...
+
+Let's add a reusable component to our pipeline. We will use the `load_from_hf_hub` component to read data from huggingface. Add the following code to your `pipeline.py` file:
+
+```Python
+load_from_hf_hub = ComponentOp.from_registry(
+    name='load_from_hf_hub',
+    component_spec_path='components/load_from_hf_hub/component_spec.yml',
```
RobbeSneyders, line 57: Would stick to `fondant_component.yaml` as name.

docs/getting_started.md
```diff
@@ -0,0 +1,169 @@
...
+- **dataset_name**: The name of the dataset on huggingface hub, here we load a [dataset with pokemon images](https://huggingface.co/datasets/huggan/pokemon)
+- **n_rows_to_load**: The number of rows to load from the dataset. This is useful for testing your pipeline on a small scale.
+- **column_name_mapping**: A mapping of the columns in the dataset to the columns in fondant. Here we map the `image` column in the dataset to the `images_raw` subset_column in fondant.
+- **image_column_names**: A list of the original image column names in the dataset. This is used to format the image from the huggingface hub format to a byte string.
+
+
+2. We add our created componentOp to the pipeline using the `add_op` method. This component has no dependencies since it is the first component in our pipeline.
+
+
+Next create a file `load_from_hf_hub/component_spec.yml` with the following content:
```
RobbeSneyders, line 84: Would stick to `fondant_component.yaml` as name.

docs/getting_started.md
RobbeSneyders: We might just have to explain it a bit better, as it is the most useful component to easily load some data to get started. Not sure if we want to describe this as a third kind of component, which is a "generic" component.

docs/getting_started.md
```diff
@@ -0,0 +1,169 @@
...
+│   └── load_from_hf_hub
+│       └── component_spec.yml
+└── pipeline.py
+```
+
+We now have a fully functional Fondant pipeline, it does not have much functionality yet, but it is a good starting point to build upon. We can already try running this limited example in order to validate our setup.
+
+## Running your pipeline
+
+A Fondant pipeline needs to be compiled before it can be ran. This means translating the user friendly Fondant pipeline definition into something that can be executed by a runner.
```
RobbeSneyders, line 131: I would leave out the compilation step here and only explain the combined functionality which you get when using `fondant run`.

pyproject.toml
```diff
@@ -46,13 +46,13 @@ dask = {extras = ["dataframe"], version = ">= 2023.4.1"}
...
 pyarrow = ">= 11.0.0"
+kfp = { version = ">= 1.8.19"}
 
-kfp = { version = ">= 1.8.19", optional = true }
 kubernetes = { version = ">= 18.20.0", optional = true }
 pandas = { version = ">= 1.3.5", optional = true }
 
 [tool.poetry.extras]
-pipelines = ["kfp", "kubernetes"]
+pipelines = ["kubernetes"]
```
RobbeSneyders, line 55: What's the reason for these changes? `kfp` is quite a big dependency which you don't need when installing `fondant` in a component, which is why we added it as an optional dependency to the `pipeline` extra.

RobbeSneyders (commented):
Thanks @GeorgesLorre! don't forget to add the file to the documentation index.

docs/getting_started.md
```diff
@@ -0,0 +1,169 @@
...
+```bash
+fondant compile pipeline:my_pipeline --local
+```
+
+We call the fondant CLI to compile our pipeline, we pass a reference to our pipeline using the import_string syntax `<module>:<instance>`. We also pass the `--local` flag to indicate we want to compile our pipeline for the local runner.
+Running this command will create a `docker-compose.yml` file with the compiled pipeline definition. Feel free to inspect this file but changing it is not needed.
+
+Note that if you use a local `base_path` in your pipeline declaration that this path will be mounted in the docker containers. This means that the data will be stored locally on your machine. If you use a cloud storage path, the data will be stored in the cloud.
+
+Now that we have compiled our pipeline, we can run it:
```
mrchtr, line 148: Maybe we should mentioned that the docker images are stored in the Github container registry. I didn't pulled images from there before which leads me into an additional docker login step. 
Basically following the steps here: https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#authenticating-to-the-container-registry

pyproject.toml
GeorgesLorre: kfp is used by the pipeline.py so it is needed for `fondant compile` and `fondant run`

RobbeSneyders: Yes, so then the user needs to install `fondant[pipelines]` (`fondant[kfp]` might be better). But we don't want to install this in every component.

README.md
```diff
@@ -85,6 +85,11 @@ Eg. generating logos:
 
 <p align="right">(<a href="#chocolate_bar-fondant">back to top</a>)</p>
 
+## 💨 Getting Started
```
RobbeSneyders, line 88: I would move this before the example pipelines

components/image_resolution_extraction/src/main.py
```diff
@@ -39,8 +39,9 @@ def transform(self, dataframe: pd.DataFrame) -> pd.DataFrame:
         """
         logger.info("Filtering dataset...")
 
-        dataframe[[("images", "width"), ("images", "height")]] = \
-            dataframe[[("images", "data")]].map(extract_dimensions)
+        dataframe[[("images", "width"), ("images", "height")]] = dataframe[
+            [("images", "data")]
+        ].apply(lambda x: extract_dimensions(x.iloc[0]), axis=1)
```
RobbeSneyders, line 44: Can we use the column name here instead of the index? That would make it easier to understand.

docs/getting_started.md
```diff
@@ -0,0 +1,310 @@
...
+## Your first pipeline
+
+Create a `pipeline.py` file in the root of your project and add the following code:
+
+```Python
+from fondant.pipeline import Pipeline, ComponentOp
+
+my_pipeline = Pipeline(
+    pipeline_name='my_pipeline',
+    base_path='/home/username/my_pipeline', <--- Make sure to update this
```
RobbeSneyders, line 29:
```suggestion
    base_path='/home/username/my_pipeline',  # TODO: update this
```

docs/getting_started.md
```diff
@@ -0,0 +1,310 @@
...
+
+- **pipeline_name**: A name to reference your pipeline.
+- **base_path**: The base path that fondant should use to store artifacts and data. This base_path can be a local path or a cloud storage path (e.g. s3://my_bucket/artifacts, or gs://my_bucket/artifacts).
+- **pipeline_description**: A description of your pipeline.
+
+## Adding components
+
+Now that we have a pipeline, we can add components to it. Components are the building blocks of your pipeline. They are the individual steps that will be executed in your pipeline. There are 2 main types of components:
+
+- reusable components: These are components that are already created by the community and can be easily used in your pipeline. You can find a list of reusable components [here](https://github.com/ml6team/fondant/tree/main/components). They often have arguments that you can set to configure them for your use case.
```
RobbeSneyders, line 44:
```suggestion
- **reusable components**: These are components that are already created by the community and can be easily used in your pipeline. You can find a list of reusable components [here](https://github.com/ml6team/fondant/tree/main/components). They often have arguments that you can set to configure them for your use case.
```

docs/getting_started.md
```diff
@@ -0,0 +1,310 @@
...
+        """
+        Args:
+            dataframe: Dask dataframe
+        Returns:
+            dataset.
+        """
+        logger.info("Filtering dataset...")
+
+        dataframe[[("images", "width"), ("images", "height")]] = \
+            dataframe[[("images", "data")]].apply(lambda x:extract_dimensions(x.iloc[0]), axis=1)
```
RobbeSneyders, line 231: Same comment as above.

RobbeSneyders (commented):
Thanks @GeorgesLorre, some minor comments, but otherwise looks good!

RobbeSneyders (approved):
LGTM!

PR #260, opened 2023-07-03
Large scale controlnet

PhilippeMoussalli:
PR for running the controlnet pipeline end-to-end on KFP. 

Some observations when doing the pipeline testing: 

- Tested with @ChristiaensBert VM and it runs really nice and much faster than the public clip service. 
- I could not test everything end to end locally since the GPU component are difficult to run locally -> switched to KFP to leverage the GPU VMs
- I had to rebuild images using the build and tag images in the `scripts` folder. I think we still need to modify the script to enable only building specified components since it currently default to all components in the `components` directory which might take some time to build
- The local runner does not seem to do the subset checking yet and we still need to expand the CLI to be able to use the kfp runner (currently not supported). Although the CLI is really nice overall :) 
- Pipeline runs fine and writes the dataset to the hub but fails at the end since it expects an output manifest. This can be resolved with this [ticket](https://github.com/ml6team/fondant/pull/221). We should prioritize this. 

Notes:
- Changed the segmentation to output a segmentation image instead of a segmentation array since that's the output expected for controlnet training 

Things to do: 
- Estimate how much the job would cost


GeorgesLorre:
> The local runner does not seem to do the subset checking yet and we still need to expand the CLI to be able to use the kfp runner (currently not supported). Although the CLI is really nice overall :)

components/prompt_based_laion_retrieval/fondant_component.yaml
```diff
@@ -25,3 +25,7 @@ args:
   aesthetic_weight:
     description: Weight of the aesthetic embedding when added to the query, between 0 and 1
     type: float
+  url:
+    description: The url of the backend clip retrieval service, defaults to the public service
+    type: str
+    default: https://knn.laion.ai/knn-service
```
NielsRogge, line 31: It doesn't seem like the main.py script has a default

NielsRogge (approved):

components/prompt_based_laion_retrieval/fondant_component.yaml
PhilippeMoussalli: The defaults defined here translate internally to defaults defined in the argparser since kfp always requires a given provided argument if specified and cannot be empty. 
```
parser.add_argument("--url", default="https://knn.laion.ai/knn-service")
```

The values defined in the argument parser generally take precedence over the default values defined in the main.py file so adding them there can be a bit misleading (e.g. if the user attempts to change them, the default values won't be used).

components/download_images/fondant_component.yaml
```diff
@@ -1,6 +1,6 @@
 name: Download images
 description: Component that downloads images based on URLs
-image: ghcr.io/ml6team/download_images:dev
+image: ghcr.io/ml6team/download_images:latest
```
RobbeSneyders, line 3:
```suggestion
image: ghcr.io/ml6team/download_images:dev
```

The images on main should be fixed to `dev`, which corresponds to the latest main version. `latest` corresponds to the latest release.

RobbeSneyders (changes requested):

components/segment_images/src/main.py
```diff
@@ -32,9 +32,13 @@ def convert_to_rgb(seg: np.array):
     for label, color in enumerate(palette):
         color_seg[seg == label, :] = color
 
-    color_seg = color_seg.astype(np.uint8).tobytes()
+    color_seg = color_seg.astype(np.uint8)
+    image = Image.fromarray(color_seg).convert('RGB')
 
-    return color_seg
+    crop_bytes = io.BytesIO()
+    image.save(crop_bytes, format="JPEG")
```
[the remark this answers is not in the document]
PhilippeMoussalli: No, this just saves it to `crop_bytes` which is a BytesIO object (in-memory buffer to store the image in binary format)

components/download_images/fondant_component.yaml
PhilippeMoussalli: right forget to revert this back!

components/segment_images/src/main.py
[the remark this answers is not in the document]
RobbeSneyders: Ok makes sense, thanks!

PR #262, opened 2023-07-04
Redefine empty images array each loop

RobbeSneyders:
The array of images was not reset in between loops, leading to the same image being set in each component specification.


PhilippeMoussalli (approved):
Thanks for the quick fix :)

PR #265, opened 2023-07-04
Enable building specified components

PhilippeMoussalli:
PR that enables specifying components during the build and tagging process, this makes it easier to build and push specific components during remote development.


scripts/build_components.sh
```diff
@@ -6,12 +6,16 @@ function usage {
   echo "Options:"
   echo "  -t, --tag=<value>        Tag to add to image, repeatable
                                    The first tag is set in the component specifications"
+  echo "  -c, --cache <value>      Use registry caching when building the components (default:true)"
```
RobbeSneyders, line 9: The default is false :)

RobbeSneyders (commented):
Thanks @PhilippeMoussalli 

I would like to be able to build the `build_components.sh` script for example components as well. So I would propose to add the following arguments:

- `-d, --components-dir`  directory with components to build, defaults to `components`
- `-co, --component` specific component to build in the components dir, defaults to all

Note that currently we are passing in the first `tag` as the fondant version as well, which is used as `pip install git+https://github.com/ml6team/fondant@$tag`. So this will not work with arbitrary tags. They actually need to be refs available on the github repository. This script was made for usage in CI/CD and I'm not sure how well it will work outside it.

PhilippeMoussalli:
> Thanks @PhilippeMoussalli
> 
> I would like to be able to build the `build_components.sh` script for example components as well. So I would propose to add the following arguments:
> 
> * `-d, --components-dir`  directory with components to build, defaults to `components`
> * `-co, --component` specific component to build in the components dir, defaults to all
> 
> Note that currently we are passing in the first `tag` as the fondant version as well, which is used as `pip install git+https://github.com/ml6team/fondant@$tag`. So this will not work with arbitrary tags. They actually need to be refs available on the github repository. This script was made for usage in CI/CD and I'm not sure how well it will work outside it.

I can add the component directory argument in this PR. Regarding the other issue, I am still tagging them with the latest release tag and then using the `tag_component` script to give them a desired tag and test them out (`dev` or `latest`). It's fine for now for testing but perhaps we should later on look into building directly from the current source code.

RobbeSneyders:
> I can add the component directory argument in this PR. Regarding the other issue, I am still tagging them with the latest release tag and then using the tag_component script to give them a desired tag and test them out (dev or latest). It's fine for now for testing but perhaps we should later on look into building directly from the current source code.

Not sure if I understand what you mean with this, but outside of the CI/CD pipeline, we should never tag images with:
- release versions
- dev
- latest

As these have special meaning in our release flow.

PhilippeMoussalli:
> > I can add the component directory argument in this PR. Regarding the other issue, I am still tagging them with the latest release tag and then using the tag_component script to give them a desired tag and test them out (dev or latest). It's fine for now for testing but perhaps we should later on look into building directly from the current source code.
> 
> Not sure if I understand what you mean with this, but outside of the CI/CD pipeline, we should never tag images with:
> 
> * release versions
> * dev
> * latest
> 
> As these have special meaning in our release flow.

Updated to add the directory as an argument to enable building custom images. 

I see, so if we want to avoid tagging image with the `release_version` we have to uncouple the fact that the tag is used to specify the fondant version to build and also specifies the image tag.

RobbeSneyders:
I don't think arbitrary tags will work, but (remote) commit hashes will.

Eg. if you run `./scripts/build_components.sh -t d23e23`, it will:
- Tag images with `d23e23`
- Pin the fondant dependency to `git+https://github.com/ml6team/fondant@d23e23`
- Update `fondant_component.yaml` to use the `d23e23` image

PhilippeMoussalli:
> I don't think arbitrary tags will work, but (remote) commit hashes will.
> 
> Eg. if you run `./scripts/build_components.sh -t d23e23`, it will:
> 
> * Tag images with `d23e23`
> * Pin the fondant dependency to `git+https://github.com/ml6team/fondant@d23e23`
> * Update `fondant_component.yaml` to use the `d23e23` image

yes, the tagging I mentioned was mainly related to running the `tag_component.sh` script but I think this one is mainly reserved for releases to tag with the `latest` tag. In that case, we can still use the `build_components.sh` script for:
- developing before merging to main (using the commit as the tag) -> requires running the script manually
- start developing on the current main branch before release (using the dev as the tag) -> handled by CI/CD upon merge to main
- using the latest images after the release (using the `latest` or `release` tag) -> handled by CI/CD upon release

Do you agree with this workflow? Anything else that needs to be done in this PR?

scripts/build_components.sh
```diff
@@ -24,56 +35,59 @@ if [ -z "${tags}" ]; then
...
+components_dir=$root_dir/${components_dir}
 
 # Loop through all subdirectories
-for dir in "$component_dir"/*/; do
+for dir in "$components_dir"/*/; do
   pushd "$dir"
 
   BASENAME=${dir%/}
   BASENAME=${BASENAME##*/}
+  if [[ "$BASENAME" == "${component}" ]] || [[ "${component}" == "all" ]]; then
```
RobbeSneyders, line 56: Can we do this a bit different? Python pseudo code:

```
components = [dir in components_dir] if component == "all" else [component]
print(f"building components {components}")
for component in components:
    ...
```

I think this would be easier to understand and debug.

scripts/build_components.sh
```diff
@@ -75,5 +99,6 @@ for dir in "$component_dir"/*/; do
    --label org.opencontainers.image.source=https://github.com/${namespace}/${repo} \
    .
 
-  popd
+   popd
```
RobbeSneyders, line 102: Wrong indentation

scripts/build_components.sh
```diff
@@ -61,10 +84,11 @@ for dir in "$component_dir"/*/; do
     args+=(-t "$tag")
   done
 
-  # Add cache arguments if caching is enabled
+#  # Add cache arguments if caching is enabled
   if [ "$caching" = true ] ; then
+
     cache_name=ghcr.io/${namespace}/${BASENAME}:build-cache
-    echo "Caching from/to ${cache_name}"
+     echo "Caching from/to ${cache_name}"
```
RobbeSneyders, line 91: Wrong indentation

scripts/build_components.sh
```diff
@@ -61,10 +84,11 @@ for dir in "$component_dir"/*/; do
     args+=(-t "$tag")
   done
 
-  # Add cache arguments if caching is enabled
+#  # Add cache arguments if caching is enabled
```
RobbeSneyders, line 87: Double comment

scripts/build_components.sh
```diff
@@ -4,39 +4,61 @@ set -e
...
+  -t |--tag) tags+=("$2"); shift;;
+  -co|--component) components+=("$2"); shift;;
+  -c |--cache) caching=true;;
+  -h |--help) usage; exit;;
   *) echo "Unknown parameter passed: $1"; exit 1;;
 esac; shift; done
 
 # Check for required argument
-if [ -z "${tags}" ]; then
+if [ ${#tags[@]} -eq 0 ]; then
```
RobbeSneyders, line 33: Why was this needed?

RobbeSneyders (commented):
Thanks @PhilippeMoussalli! Looks good to me apart from some formatting issues.

scripts/build_components.sh
PhilippeMoussalli: both achieve the same purpose but was getting suggestion to change it in my IDE so I opted for checking if the array length is zero instead. I can revert it back if you prefer. 
image

RobbeSneyders: My question was mainly meant to learn, not to ask you to revert :sweat_smile: 
But if they're equivalent, both are fine for me.

RobbeSneyders (approved):

RobbeSneyders:
Seems like you have a conflict @PhilippeMoussalli. Feel free to merge after you resolve it.

PR #269, opened 2023-07-05
[Commoncrawl pipeline] Add load from commoncrawl component

shayorshay:
This component is the first part of the commoncrawl pipeline. Given an index, this component loads the corresponding index file from the AWS Public Data Sets (S3 bucket) and returns a list of its WARC segment file paths as a dataframe.

### DIFF
--- a/examples/pipelines/commoncrawl/components/load_from_commoncrawl/requirements.txt
+++ b/examples/pipelines/commoncrawl/components/load_from_commoncrawl/requirements.txt
@@ -1,3 +1,2 @@
 boto3==1.26.158
 fondant
-pyarrow>=7.0
--- a/examples/pipelines/commoncrawl/components/load_from_commoncrawl/requirements.txt
+++ b/examples/pipelines/commoncrawl/components/load_from_commoncrawl/requirements.txt
@@ -1,2 +1,2 @@
 boto3==1.26.158
-fondant
+git+https://github.com/ml6team/fondant@main
--- a/examples/pipelines/commoncrawl/components/load_from_commoncrawl/requirements.txt
+++ b/examples/pipelines/commoncrawl/components/load_from_commoncrawl/requirements.txt
@@ -1,0 +1,3 @@
+boto3==1.26.158
+fondant
+pyarrow>=7.0
--- a/examples/pipelines/commoncrawl/components/load_from_commoncrawl/src/main.py
+++ b/examples/pipelines/commoncrawl/components/load_from_commoncrawl/src/main.py
@@ -1,0 +1,71 @@
+"""This component loads a dataset from CommonCrawl based on a given index."""
+import logging
+import typing as t
+
+import io
+import boto3
+import gzip
+
+import dask.dataframe as dd
+import pandas as pd
+
+from fondant.component import LoadComponent
+
+logger = logging.getLogger(__name__)
+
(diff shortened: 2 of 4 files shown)


examples/pipelines/commoncrawl/components/load_from_commoncrawl/requirements.txt
```diff
@@ -0,0 +1,3 @@
+boto3==1.26.158
+fondant
+pyarrow>=7.0
```
PhilippeMoussalli, line 3: I think pyarrow is installed with Fondant so no need to include it here

examples/pipelines/commoncrawl/components/load_from_commoncrawl/src/main.py
```diff
@@ -0,0 +1,71 @@
...
+import gzip
+
+import dask.dataframe as dd
+import pandas as pd
+
+from fondant.component import LoadComponent
+
+logger = logging.getLogger(__name__)
+
+S3_BASE_URL = "s3://commoncrawl/crawl-data"
```
PhilippeMoussalli, line 16: This is not being used anywhere

examples/pipelines/commoncrawl/components/load_from_commoncrawl/src/main.py
```diff
@@ -0,0 +1,71 @@
...
+S3_COMMONCRAWL_BUCKET = "commoncrawl"
+
+
+def fetch_warc_file_from_s3(s3_bucket: str, s3_key) -> dd.DataFrame:
```
PhilippeMoussalli, line 20: Can you add a docstring for the argumens. 

Would also prefer to change `s3_key` to `bucket_path`. Feels more intuitive

PhilippeMoussalli (approved):
Thanks @shayorshay ! Looks like a very clean component ;) 
Left a few minor comments but should be good to go

examples/pipelines/commoncrawl/components/load_from_commoncrawl/src/main.py
shayorshay: `s3_key` is the key of the object in the bucket not the bucket path. Would `object_key` be a better name?

PhilippeMoussalli: my bad, just read on it and it seems to be an AWS specific notation (still used to GCP). It's fine to leave it as is

shayorshay: ok great!

examples/pipelines/commoncrawl/components/load_from_commoncrawl/requirements.txt
```diff
@@ -0,0 +1,2 @@
+boto3==1.26.158
+fondant
```
RobbeSneyders, line 2:
```suggestion
git+https://github.com/ml6team/fondant@main
```

RobbeSneyders (approved):
Thanks @shayorshay, one minor comment since our latest version of the build script expects fondant dependencies to be defined from github.

examples/pipelines/commoncrawl/components/load_from_commoncrawl/requirements.txt
shayorshay: done :)

PR #273, opened 2023-07-05
[Commoncrawl pipeline] Add component download_commoncrawl_segments

shayorshay:
This PR adds the second component of the Commoncrawl pipeline. The component downloads the WARC segment files and extracts the webpage urls and html code to be returned as a dask dataframe.


examples/pipelines/commoncrawl/components/download_commoncrawl_segments/src/main.py
```diff
@@ -0,0 +1,77 @@
...
+from warcio.archiveiterator import ArchiveIterator
+
+from fondant.component import DaskTransformComponent
+
+logger = logging.getLogger(__name__)
+
+BASE_URL = "https://data.commoncrawl.org/"
+
+
+def get_records_from_warc_file(warc_file: str, n_records_to_download: int) -> List:
```
PhilippeMoussalli, line 23: The `n_records_to_download` should be an optional argument and below you should only break if it not `None`

examples/pipelines/commoncrawl/components/download_commoncrawl_segments/src/main.py
```diff
@@ -0,0 +1,77 @@
...
+    """Downloads a WARC file and extracts the webpages.
+    Args:
+        warc_file: The path to the WARC file.
+        n_records_to_download: The number of webpages to download from the WARC file.
+    Returns:
+        A list of webpages.
+    """
+    logger.info(f"Processing WARC file from segment path: {warc_file}...")
+    records = []
```
PhilippeMoussalli, line 32: do we risk running into out-of-memory issues if the number of extracted records is too large? right now it seems like we're collecting them all in a list and materializing them in memory before transforming them into a Dask Dataframe.

PhilippeMoussalli (commented):
Thanks Sharon :) 
Looks good overall but just have a small doubt regarding the current scalability of this approach.

examples/pipelines/commoncrawl/components/download_commoncrawl_segments/fondant_component.yaml
```diff
@@ -0,0 +1,23 @@
...
+      path:
+        type: string
+
+produces:
+  webpage:
+    fields:
+      url:
+        type: string
+      html:
+        type: binary
```
PhilippeMoussalli, line 17: why is this not a string as well?

examples/pipelines/commoncrawl/components/download_commoncrawl_segments/src/main.py
RobbeSneyders: Materializing a list and then converting it to a dataframe is the recommended approach, since appending a dataframe is a lot more expensive. We might want to do this on a partition level though.

examples/pipelines/commoncrawl/components/download_commoncrawl_segments/src/main.py
```diff
@@ -0,0 +1,81 @@
...
+        self, df: dd.DataFrame, n_records_to_download: t.Optional[int] = None
+    ) -> dd.DataFrame:
+        """Downloads Commoncrawl segments based on a list of WARC paths.
+        Args:
+            df: A Dask DataFrame containing a column of WARC paths.
+            n_webpages_to_download: The number of webpages to download from each segment.
+        Returns:
+            A Dask DataFrame containing the downloaded webpages.
+        """
+        segment_paths = df["segment_path"].to_bag()
```
RobbeSneyders, line 67: What is the reason to use the `bag` interface here?

examples/pipelines/commoncrawl/components/download_commoncrawl_segments/src/main.py
PhilippeMoussalli: does that mean that we need one partition per WARC file to keep the dataframes as small as possible?

examples/pipelines/commoncrawl/components/download_commoncrawl_segments/src/main.py
shayorshay: So we can use dask `map` and `flatten` operations in the next steps. Using a list caused errors when converting the results to a dask dataframe later.

RobbeSneyders: Yes, but I mean compared to the dataframe API, where we have the `apply` function which behaves similarly.

examples/pipelines/commoncrawl/components/download_commoncrawl_segments/src/main.py
```diff
@@ -0,0 +1,81 @@
...
+    logger.info(f"Processing WARC file from segment path: {warc_file}...")
+    records = []
+    counter = 0
+    response = requests.get(BASE_URL + warc_file, stream=True)
+    response.raise_for_status()
+
+    for record in ArchiveIterator(response.raw, arc2warc=True):
+        if record.rec_type == "response":
+            url = record.rec_headers.get_header("WARC-Target-URI")
+            content = (
```
mrchtr, line 42: Do you think it would be possible to add an optional possibility to return plain text instead of the html? 
We could add an additional parameter if html or plain text should be returned. I think you don't need it for your use case, but for the large language model use case it would be super useful! :)

For the plain text transformation you could use BeautifulSoup. Some time ago I did something similar by using this code function: 
```
    def _convert_to_plain_text(html):
        """Convert html body into plain text. Making sure table rows are on seperate lines."""
        soup = BeautifulSoup(html)
        body_content = soup.find("body")
        text = ""

        for e in body_content.descendants:
            if isinstance(e, str):
                text += e.strip()
            elif e.name in ['br', 'h1', 'h2', 'h3', 'h4', 'tr', 'th']:
                text += '\n'
            elif e.name in ['td']:
                text += ' '
            elif e.name == 'p' and not any(parent.name == 'table' for parent in e.parents):
                text += '\n'
            elif e.name in ['td']:
                text += '\t'
            elif e.name == 'li':
                text += '\n- '

        return text
```

RobbeSneyders: Is `soup.get_text()` not sufficient?

mrchtr: Probably. The custom code handles table a bit differently. Basically, try to keep the table structure line by line.

shayorshay: I see the custom code keeps the CSS content in the <style> tag. I'm assuming you don't want this. Will `soup.get_text()` be ok for you @mrchtr ? otherwise, we'll have a long list of tags to parse

mrchtr: `soup.get_text()` works! Even better as my custom code.

examples/pipelines/commoncrawl/components/download_commoncrawl_segments/fondant_component.yaml
```diff
@@ -0,0 +1,35 @@
...
+      url:
+        type: string
+      html:
+        type: string
+
+args:
+  use_s3:
+    description: Whether to use S3 to download the commoncrawl segment file. Set to True if you are running this component on an AWS cluster.
+    type: bool
+    default: 'False'
```
RobbeSneyders, line 23:
```suggestion
    default: false
```

Aren't you passing a string here instead of a boolean? Same for the argument below.

examples/pipelines/commoncrawl/components/download_commoncrawl_segments/src/utils/text_utils.py
```diff
@@ -0,0 +1,21 @@
+import logging
+from bs4 import BeautifulSoup
+
+logger = logging.getLogger(__name__)
+
+BASE_URL = "https://data.commoncrawl.org/"
```
RobbeSneyders, line 6: I don't think this is used.

examples/pipelines/commoncrawl/components/download_commoncrawl_segments/src/utils/download_utils.py
```diff
@@ -0,0 +1,42 @@
...
+    Returns:
+        The WARC file as a bytes object.
+    """
+    s3 = boto3.client("s3")
+    response = s3.get_object(Bucket=S3_COMMONCRAWL_BUCKET, Key=s3_key)
+    return response["Body"]
+
+
+def get_warc_file_using_requests(warc_file: str) -> requests.Response:
+    retry = 0
```
RobbeSneyders, line 28: FYI, requests provides functionality for retry:
https://requests.readthedocs.io/en/latest/user/advanced/?highlight=retry#example-automatic-retries

examples/pipelines/commoncrawl/components/download_commoncrawl_segments/src/main.py
```diff
@@ -0,0 +1,122 @@
...
+        if partition_size:
+            df = df.repartition(partition_size=f"{partition_size}MB")
+
+        df = df.reset_index(drop=True)
+
+        return df
+
+
+if __name__ == "__main__":
+    component = DownloadCommoncrawlSegments.from_args()
```
RobbeSneyders, line 121: This is outdated since #302. Can you rebase on main and update this?

examples/pipelines/commoncrawl/components/download_commoncrawl_segments/src/main.py
```diff
@@ -0,0 +1,122 @@
...
+    else:
+        response = get_warc_file_using_requests(warc_file)
+        return get_records(response.raw, get_plain_text, n_records_to_download)
+
+
+class DownloadCommoncrawlSegments(DaskTransformComponent):
+    def transform(
+        self,
+        df: dd.DataFrame,
+        use_s3: Optional[bool] = False,
```
RobbeSneyders, line 75: Since #302, arguments are passed to `__init__` instead of `transform`. Can you rebase on main and update this?

examples/pipelines/commoncrawl/components/download_commoncrawl_segments/src/main.py
```diff
@@ -0,0 +1,122 @@
...
+
+        df.columns = [
+            "webpage_url",
+            "webpage_html",
+        ]
+
+        if partition_size:
+            df = df.repartition(partition_size=f"{partition_size}MB")
+
+        df = df.reset_index(drop=True)
```
RobbeSneyders, line 115: Why are you resetting the index here?

examples/pipelines/commoncrawl/components/download_commoncrawl_segments/src/main.py
```diff
@@ -0,0 +1,122 @@
...
+            .explode()
+            .apply(pd.Series, meta={0: "object", 1: "object"})
+        )
+
+        df.columns = [
+            "webpage_url",
+            "webpage_html",
+        ]
+
+        if partition_size:
```
RobbeSneyders, line 112: I don't think we need an argument for this. This can be `250MB` at all times. I think it's mainly the partition size before the `apply` that might be useful to provide an argument for. Then the user can define a smaller size so the data will still fit into memory after the `apply`.

RobbeSneyders (commented):
Thanks @shayorshay, left some more comments.

@PhilippeMoussalli can you follow up this PR?

examples/pipelines/commoncrawl/components/download_commoncrawl_segments/src/main.py
mrchtr: I have tested this component and run into the expected out-of-memory. 
I have tried to download 1 segment with 30k records. Seems to be Philippes assumption is correct, that the `records` list didn't fit into memory in my case. 

What do you think about using `dask.delayed`? Basically, create a `delayed_object` with a fixed size (e.g. 1000 records), append them to a dask dataframe, and afterwards initialise the dataframe from the delayed objects. 

```
def load_record(record_counter, batch_size) -> list:
    # returning a list of delayed objects with a fixed size 
    offset = record_counter * batch_size # use this to determine the starting point to read from
    counter = 0
    records = []
    ...
    for record in ArchiveIterator(response.raw, arc2warc=True):
        if counter >= offset:
             # read content and append to list
        if len(records) >= batch_size:
             return records


delayed_data = []
for segment in segments:
    total_number_of_records = xxx
    batch_size = 1000
    delayed_data.extend([delayed(load_record)(record_counter, batch) for record_counter in range(0, 
    total_number_of_records, batch_size)])

# initialise the final dataframe, download will be executed
dataframe = dd.from_delayed(delayed_data)

```

satishjasthi: @mrchtr That's good idea, in fact Im also using chunking to create delayed objects for load_from_files component

examples/pipelines/commoncrawl/components/download_commoncrawl_segments/src/utils/text_utils.py
```diff
@@ -0,0 +1,21 @@
...
+def convert_to_plain_text(html: str) -> str:
+    """Converts an HTML string to plain text.
+    Args:
+        html: The HTML string.
+    Returns:
+        The plain text string.
+    """
+    try:
+        soup = BeautifulSoup(html, "html.parser")
+        return soup.get_text()
```
mrchtr, line 18: Currently there seems to be a performance bottleneck by converting the page into plain text.
I have found a different library (specially designed for this use case) that seems to be faster and more robust.
Can you check out trafilatura? https://trafilatura.readthedocs.io/en/latest/

shayorshay: done. I'll update the PR to reflect the change.

examples/pipelines/commoncrawl/components/download_commoncrawl_segments/fondant_component.yaml
shayorshay: [component_spec.json](https://github.com/ml6team/fondant/blob/0abc5973f82c13b3deca544cd80f9b697e55a14a/src/fondant/schemas/component_spec.json#L84) currently allows only strings and integers for default values

RobbeSneyders: Let's update the schema then :)

examples/pipelines/commoncrawl/components/download_commoncrawl_segments/requirements.txt
RobbeSneyders, line 1: Can you pin all the versions here?

RobbeSneyders (commented):
Thanks @shayorshay. Two final comments and we can merge.

> - reverted to using a dask bag for downloading webpages. compared to dask.dataframe.apply(), this is much faster when I I tested them. I think this has to do with the flatten step at the end where we apply pd.Series.

Can you explain this a bit more? I wouldn't really expect a bag to be faster here. Maybe it's due to the different schedulers they use.

examples/pipelines/commoncrawl/components/download_commoncrawl_segments/requirements.txt
shayorshay: I retested both approaches since there've been updates wrt repartitioning and you're right, the apply approach is faster than the bag. Here's the time comparisons:
1. without repartition on 10k webpages. bag - 01:12, apply - 00:56
2. with repartition on 10k webpage. bag - 01:19, apply - 01:11

shayorshay: Testing the component on 40k webpages with repartition, the bag approach beats apply:
1. bag - 02:48
2. apply - 04:30

I'm very curious why this is the case. FYI @RobbeSneyders

examples/pipelines/commoncrawl/components/download_commoncrawl_segments/src/main.py
```diff
@@ -0,0 +1,139 @@
...
+        )
+
+        dataframe.columns = [
+            "webpage_url",
+            "webpage_html",
+        ]
+
+        dataframe = dataframe.reset_index(drop=True)
+
+        logger.info(f"Downloaded {len(dataframe)} webpages from Commoncrawl.")
```
RobbeSneyders, line 133: `len()` triggers a compute here. We might be able to use `.size` here instead, which is evaluated lazily, but not sure how we can combine it with logging. Also fine for me to just remove this.

shayorshay: Ok, I've removed it. I'll look into `.size` for local testing

mrchtr:
Thanks for the update @shayorshay! 

I have tested the component in depth and tried to make some changes to improve the performance slightly.

- We should use `trafilatura`. The library offers fast and powerful html extraction including language filtering based on meta data, blocking urls, local deduplication of repetitions within a single document

- The performance is already good. Nevertheless, machine cores are not used equally. I observed some performance peaks (repeating behaviour: CPU usage super high for some seconds, and goes almost down to zero, stays there for several ,seconds). 

Maybe we could merge this PR already. I think this would make it easier to incorporate my changes and improve the performance step wise.

RobbeSneyders (approved):

PR #276, opened 2023-07-05
Order output columns in PandasTransformComponent

### LINKED ISSUE #275: Ordering of columns should not be validated
The `PandasTransformComponent` uses `map_partitions` to call the `transform` method for each partition and takes a `meta` argument to define which output it expects. `Dask` builds the graph based on this expectation and validates the returned dataframe at runtime. Apparently this validation includes the ordering of the columns, which is not desired.

We can fix this by reordering the columns in the `wrap_transform` method.

RobbeSneyders:
Fixes #275

### DIFF
--- a/fondant/component.py
+++ b/fondant/component.py
@@ -264,27 +264,32 @@ class PandasTransformComponent(TransformComponent):
             dataframe: A Pandas dataframe containing a partition of the data
         """
 
-    def wrapped_transform(self, dataframe: pd.DataFrame) -> pd.DataFrame:
-        """Method wrapping the transform method to switch between hierarchical and flattened
-        columns.
-        """
-        dataframe.columns = pd.MultiIndex.from_tuples(
-            tuple(column.split("_")) for column in dataframe.columns
-        )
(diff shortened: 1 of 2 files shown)


fondant/component.py
```diff
@@ -264,27 +264,32 @@ def transform(self, dataframe: pd.DataFrame) -> pd.DataFrame:
...
-                or field not in self.spec.produces[subset].fields
-            ],
-        )
-        dataframe.columns = [
-            "_".join(column) for column in dataframe.columns.to_flat_index()
-        ]
-        return dataframe
+    @staticmethod
+    def wrap_transform(transform: t.Callable, *, spec: ComponentSpec) -> t.Callable:
+        def wrapped_transform(dataframe: pd.DataFrame) -> pd.DataFrame:
```
PhilippeMoussalli, line 269: Can you re-add the deleted docstring 

  """Method wrapping the transform method to switch between hierarchical and flattened
  columns.
  """

PhilippeMoussalli (approved):
Thanks Robbe :)

PR #278, opened 2023-07-06
[DataComp] Add Dedup component

NielsRogge:
This PR adds a deduplication component as described in the [SemDedup](https://arxiv.org/abs/2303.09540) paper.

### DIFF
--- a/examples/pipelines/datacomp/components/dedup_image_embeddings/src/main.py
+++ b/examples/pipelines/datacomp/components/dedup_image_embeddings/src/main.py
@@ -26,6 +26,45 @@ def sort_by_centroid_distance(embeddings, centroid, descending=True):
     return embeddings[sorted_indices.flatten()]
 
 
+def deduplicate(df, epsilon):
+    embeddings = df.image_embedding
+    cluster_embeddings = np.vstack(embeddings)
+
+    # get cluster centroid
+    cluster_center = np.mean(cluster_embeddings, axis=0)
+
+    # sort the cluster embeddings by the distance to the cluster centroid
+    cluster_embeddings = sort_by_centroid_distance(
+        cluster_embeddings, cluster_center, descending=True
+    )
+
+    # compute the pairwise cosine similarity between embeddings
+    pairwise_sim_matrix = cosine_similarity(cluster_embeddings)
+
+    # get upper triangular part of the matrix (excluding the diagonal)
+    triu_sim_matrix = np.triu(pairwise_sim_matrix, k=1)
+
+    # find max value in each column
+    M = np.max(triu_sim_matrix, axis=0)
+
+    # Check if the maximum similarity <= the threshold.
+    points_to_keep_from_cluster_i = M <= 1 - epsilon
+
+    print(points_to_keep_from_cluster_i)
+
+    # Convert to boolean Pandas series
+    points_to_keep_from_cluster_i = pd.Series(
+        points_to_keep_from_cluster_i, index=embeddings.index
+    )
+
+    # add the points to keep to the list
+    index_to_keep = points_to_keep_from_cluster_i[
+        points_to_keep_from_cluster_i == True
+    ].index
+
+    return list(index_to_keep)
+
+
 class DedupImageEmbeddingsComponent(DaskTransformComponent):
     """Component that deduplicates images based on embeddings."""
 
@@ -40,46 +79,13 @@ class DedupImageEmbeddingsComponent(DaskTransformComponent):
         Returns:
             Dask dataframe
         """
-        num_clusters = dataframe["image_cluster_label"].unique()
 
-        indices_to_keep = []
-        for i in num_clusters:
-            mask = dataframe["image_cluster_label"] == i
-            cluster_i_embeddings = dataframe[mask]["image_embedding"]
-            cluster_embeddings = np.vstack(cluster_i_embeddings)
+        # filter dataframe
+        indices_to_keep = dataframe.groupby("image_cluster_label").apply(deduplicate,
+                                                                         epsilon=self.epsilon,
+                                                                         meta=pd.Series(dtype="int8"))
 
-            print(f"Shape of embeddings cluster {i}:", list(cluster_embeddings.shape))
-
-            # get cluster centroid
-            cluster_center = np.mean(cluster_embeddings, axis=0)
-
-            # sort the cluster embeddings by the distance to the cluster centroid
-            cluster_embeddings = sort_by_centroid_distance(
-                cluster_embeddings, cluster_center, descending=True
-            )
-
-            # compute the pairwise cosine similarity between embeddings
-            pairwise_sim_matrix = cosine_similarity(cluster_embeddings)
-
-            # get upper triangular part of the matrix (excluding the diagonal)
-            triu_sim_matrix = np.triu(pairwise_sim_matrix, k=1)
-
-            # find max value in each column
-            M = np.max(triu_sim_matrix, axis=0)
-
-            # Check if the maximum similarity <= the threshold.
-            points_to_keep_from_cluster_i = M <= 1 - self.epsilon
-
-            # Convert to boolean Pandas series
-            points_to_keep_from_cluster_i = pd.Series(
-                points_to_keep_from_cluster_i, index=cluster_i_embeddings.index
-            )
-
-            # add the points to keep to the list
-            index_to_keep = points_to_keep_from_cluster_i[
-                points_to_keep_from_cluster_i == True
-            ].index
-            indices_to_keep.extend(index_to_keep)
+        indices_to_keep = list(indices_to_keep.explode())
 
         # filter dataframe
         dataframe = dataframe.loc[dataframe.index.isin(indices_to_keep)]
--- a/examples/pipelines/datacomp/components/dedup_image_embeddings/src/main.py
+++ b/examples/pipelines/datacomp/components/dedup_image_embeddings/src/main.py
@@ -5,6 +5,7 @@ As proposed in [Abbas et al., 2022](https://arxiv.org/abs/2303.09540).
 Implementation is based on https://github.com/conceptofmind/SemDeDup/tree/main.
 """
 import logging
+from typing import Union
 
 import dask.dataframe as dd
 import numpy as np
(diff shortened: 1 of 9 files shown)


RobbeSneyders:
Thanks @NielsRogge!

How is the performance of this component? I would have expected us to first partition the dataframe by cluster label and then do the actual deduplication per partition.

CC: @mrchtr

mrchtr:
Thanks @NielsRogge. Looks good so far, but I agree with @RobbeSneyders. 

If we first apply partitioning based on clusters, we should be more accurate since we expect to compare already similar entries to each other. This reduces the chance of missing a duplicate.

examples/pipelines/datacomp/components/dedup_image_embeddings/src/main.py
```diff
@@ -0,0 +1,89 @@
...
+            cluster_i_embeddings = dataframe[mask]["image_embedding"]
+            cluster_embeddings = np.vstack(cluster_i_embeddings)
+
+            print(f"Shape of embeddings cluster {i}:", list(cluster_embeddings.shape))
+
+            # get cluster centroid
+            cluster_center = np.mean(cluster_embeddings, axis=0)
+
+            # sort the cluster embeddings by the distance to the cluster centroid
+            cluster_embeddings = sort_by_centroid_distance(
```
mrchtr, line 54: Is the sorting inside of a single partition really needed?

examples/pipelines/datacomp/components/dedup_image_embeddings/src/main.py
```diff
@@ -0,0 +1,89 @@
...
+
+Implementation is based on https://github.com/conceptofmind/SemDeDup/tree/main.
+"""
+import logging
+
+import dask.dataframe as dd
+import numpy as np
+import pandas as pd
+from scipy.spatial.distance import cdist
+from sklearn.metrics.pairwise import cosine_similarity
```
mrchtr, line 13: Maybe we can pass the similarity function as argument to the component? I think for some use cases it could be useful to use different similarity methods as well.

NielsRogge:
> I would have expected us to first partition the dataframe by cluster label and then do the actual deduplication per partition.

I'm not sure that's possible, because you need to keep the embeddings of each cluster in memory to do the similarity calculation.  What if a given cluster gets split into multiple partitions?

examples/pipelines/datacomp/components/dedup_image_embeddings/src/main.py
NielsRogge: Which other similarities are you thinking of? Cause they might require specific sklearn functions

examples/pipelines/datacomp/components/dedup_image_embeddings/src/main.py
NielsRogge: It's based on the pseudocode provided in the semdedup paper: https://arxiv.org/abs/2303.09540.

I also wonder whether it influences the results

examples/pipelines/datacomp/components/dedup_image_embeddings/src/main.py
```diff
@@ -0,0 +1,96 @@
...
+        self.epsilon = epsilon
+
+    def transform(self, dataframe: dd.DataFrame) -> dd.DataFrame:
+        """
+        Args:
+            dataframe: Dask dataframe
+
+        Returns:
+            Dask dataframe
+        """
```
RobbeSneyders, line 81: Doesn't provide a lot of information. Would either add more info or remove.

examples/pipelines/datacomp/components/dedup_image_embeddings/src/main.py
```diff
@@ -0,0 +1,96 @@
...
+        points_to_keep_from_cluster_i == True
+    ].index
+
+    return list(index_to_keep)
+
+
+class DedupImageEmbeddingsComponent(DaskTransformComponent):
+    """Component that deduplicates images based on embeddings."""
+
+    def __init__(self, *_, epsilon: float) -> None:
```
RobbeSneyders, line 71: Can you document the argument?

examples/pipelines/datacomp/components/dedup_image_embeddings/src/main.py
```diff
@@ -0,0 +1,96 @@
...
+        """
+        Args:
+            dataframe: Dask dataframe
+
+        Returns:
+            Dask dataframe
+        """
+        indices_to_keep = dataframe.groupby("image_cluster_label").apply(deduplicate,
+                                                                         epsilon=self.epsilon,
+                                                                         meta=pd.Series(dtype="int8"))
```
RobbeSneyders, line 84: You can get the `dtype` for `meta` from the index of the original dataframe. Otherwise this component only works for integer indices.

examples/pipelines/datacomp/components/dedup_image_embeddings/src/main.py
```diff
@@ -0,0 +1,96 @@
...
+    embeddings = df.image_embedding
+    cluster_embeddings = np.vstack(embeddings)
+
+    # get cluster centroid
+    cluster_center = np.mean(cluster_embeddings, axis=0)
+
+    # sort the cluster embeddings by the distance to the cluster centroid
+    cluster_embeddings = sort_by_centroid_distance(
+        cluster_embeddings, cluster_center, descending=True
+    )
```
RobbeSneyders, line 39: Do we sort here to keep the entries closest to the center since we expect them to be most representative? Because it doesn't really seem necessary otherwise.

If my understanding is correct, can you add this to the comment? It took me a while to figure out that this might be the reason.

examples/pipelines/datacomp/components/dedup_image_embeddings/src/main.py
```diff
@@ -0,0 +1,96 @@
...
+    # get upper triangular part of the matrix (excluding the diagonal)
+    triu_sim_matrix = np.triu(pairwise_sim_matrix, k=1)
+
+    # find max value in each column
+    M = np.max(triu_sim_matrix, axis=0)
+
+    # Check if the maximum similarity <= the threshold.
+    points_to_keep_from_cluster_i = M <= 1 - epsilon
+
+    print(points_to_keep_from_cluster_i)
```
RobbeSneyders, line 53: Print statements can be removed or switched to logging.

examples/pipelines/datacomp/components/dedup_image_embeddings/src/main.py
```diff
@@ -0,0 +1,96 @@
...
+    )
+
+    # compute the pairwise cosine similarity between embeddings
+    pairwise_sim_matrix = cosine_similarity(cluster_embeddings)
+
+    # get upper triangular part of the matrix (excluding the diagonal)
+    triu_sim_matrix = np.triu(pairwise_sim_matrix, k=1)
+
+    # find max value in each column
+    M = np.max(triu_sim_matrix, axis=0)
```
RobbeSneyders, line 48: Can you pick a lowercase multi-letter variable name her?

examples/pipelines/datacomp/components/dedup_image_embeddings/src/main.py
```diff
@@ -0,0 +1,96 @@
...
+    points_to_keep_from_cluster_i = pd.Series(
+        points_to_keep_from_cluster_i, index=embeddings.index
+    )
+
+    # add the points to keep to the list
+    index_to_keep = points_to_keep_from_cluster_i[
+        points_to_keep_from_cluster_i == True
+    ].index
+
+    return list(index_to_keep)
```
RobbeSneyders, line 65: Wouldn't the following one-liner work here as well:

```python
return df[points_to_keep_from_cluster_i].index.tolist()
```

examples/pipelines/datacomp/pipeline.py
```diff
@@ -58,24 +57,21 @@
...
-# compile
-if __name__ == "__main__":
-    compiler = DockerCompiler()
-    # mount the gcloud credentials to the container
-    extra_volumes = [
-        "$HOME/.config/gcloud/application_default_credentials.json:/root/.config/gcloud/application_default_credentials.json:ro"
-    ]
-    compiler.compile(pipeline=pipeline, extra_volumes=extra_volumes)
-    logger.info("Run `docker compose up` to run the pipeline.")
+# pipeline.add_op(filter_image_resolution_op, dependencies=load_from_hub_op)
```
RobbeSneyders, line 73: Should be reverted before merging.

examples/pipelines/datacomp/components/dedup_image_embeddings/src/main.py
```diff
@@ -0,0 +1,96 @@
...
+
+def sort_by_centroid_distance(embeddings, centroid, descending=True):
+    distances = cdist(embeddings, centroid.reshape(1, -1), "euclidean")
+    sorted_indices = np.argsort(distances, axis=0)
+    if descending:
+        sorted_indices = sorted_indices[::-1]
+    return embeddings[sorted_indices.flatten()]
+
+
+def deduplicate(df, epsilon):
```
RobbeSneyders, line 29: Can you add types here? I assumed at first that this was a Dask dataframe, but if I understand correctly it's a pandas dataframe.

RobbeSneyders (commented):
Thanks @NielsRogge!

Left some more comments, but a lot of them are stuff to revert before merging. I think this component could really benefit from some unit tests as well, so would be nice if you could add them.

scripts/build_components.sh
```diff
@@ -75,7 +75,7 @@ for dir in "${components_to_build[@]}"; do
 
   echo "Updating the image version in the fondant_component.yaml with:"
   echo "${full_image_names[0]}"
-  sed -i "s|^image: .*|image: ${full_image_names[0]}|" fondant_component.yaml
+  sed -i '' "s|^image: .*|image: ${full_image_names[0]}|" fondant_component.yaml
```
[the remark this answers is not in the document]
RobbeSneyders: Just saw in your other PR that this was meant to be pushed, but this does not work on linux.

examples/pipelines/datacomp/components/dedup_image_embeddings/src/main.py
mrchtr: For my use case I need the jaccard similarity. I modified the component slightly. Maybe we can try to combine it later on. For now, I will use my custom one.

scripts/build_components.sh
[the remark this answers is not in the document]
NielsRogge: Yes this fix was needed to resolve this issue: https://stackoverflow.com/questions/29081799/sed-1-invalid-command-code-f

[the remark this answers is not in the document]
NielsRogge: cc @GeorgesLorre

PR #279, opened 2023-07-06
Always pull images in local runner

### LINKED ISSUE #259: Direct the runners to always pull images
The version of the reusable components is fixed to `dev` on the main branch. The `dev` tag is always moved to the last image built on the `main` branch. However if there's already a local image available with the `dev` tag, docker will not try to pull it again. This can lead to a pipeline using an older image and hard to debug errors.

For released fondant versions, this is not an issue, as the reusable component versions are fixed when building fondant, and a fixed version will always refer to the same docker image.

Since this is mostly an issue for fondant developers, we might want to add an optional flag to the runner. Especially if always pulling would have a negative impact on performance.

RobbeSneyders:
Fixes #259 

This PR:
- Instructs the local runner to always pull images. This makes sure that moving tags like `dev` or `latest` are up to date before running the pipeline. As far as I can tell from my testing, this doesn't cause a noticeable delay when the image was not updated.
- Adds the pipeline name to the docker compose yaml and removes orphan containers. Containers are namespaced by the pipeline name, which defaults to the cwd if not specified. So I set the name explicitly so we don't accidentally remove containers from another pipeline.

### DIFF
--- a/fondant/compiler.py
+++ b/fondant/compiler.py
@@ -178,4 +178,8 @@ class DockerCompiler(Compiler):
                 services[safe_component_name][
                     "image"
                 ] = component_op.component_spec.image
-        return {"version": "3.8", "services": services}
+        return {
+            "name": pipeline.name,
+            "version": "3.8",
+            "services": services,
+        }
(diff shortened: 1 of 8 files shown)


PhilippeMoussalli (approved):
Thanks Robbe! Seems like there are still some issues related to some tests that have failed

RobbeSneyders:
Yeah, I forgot to remove a change I made to make sure the tests fail in the right scenario :stuck_out_tongue:

PR #281, opened 2023-07-06
Make components cloud agnostic

### LINKED ISSUE #210: Make components cloud agnostic
## Problem Statement

Connecting to different remote data sources with Dask (uses fsspec behind the scenes) requires the installation of different packages for each remote environment as mentioned [here](https://docs.dask.org/en/stable/how-to/connect-to-remote-data.html):
* s3fs for AWS
* gcsfs for GCP 
* adlfs for Azure

Currently, we're only supporting GCP by hardcoding the installation of gcsfs inside the component as a separate package. As a results, the components can only interface with the GCP storage, we need to:
* Make them generic to interface with different cloud storages 
* Make sure that it is clear for the user that those extra dependencies need to be installed 

## Proposed Approach

Two options:
**1) Make all of them available within the fondant package.** 
We define extras for every cloud within the fondant package and include the relevant dependencies for each cloud. `fondant[gcp], fondant[aws], fondant[azure]`
For the reusable component we make sure we install all of them in the docker image: `fondant[azure, aws, gcp]` to make sure they are cloud generic. The packages themselves are relatively lightweight [source-aws](https://github.com/fsspec/s3fs/blob/main/requirements.txt)[source-azure](https://github.com/fsspec/adlfs/blob/main/setup.py)[source-gcp](https://github.com/fsspec/gcsfs/blob/main/requirements.txt) and share some common packages dependencies. We'll just need to check that they don't conflict with each other. For custom components, the users can define the dependency themselves if they are only working with one cloud provider (e.g. only install `fondant[gcp]` if they are working with GCP) 

**2) Install them directly within the docker images (current way)**
Less of a favorable option since the user needs to know about the name of the specific packages + more error prone (if user forgets to include them) and does not offer any significant advantage over the previous approach 

# Implementation Steps/Tasks

Specified above+ update docker images 

## Potential Impact

No potential impact, only caveat for not passing custom storage arguments is that we are assuming the SA of KFP is setup correctly with proper permissions. 

## Testing

Test environment should not have the specific cloud packages installed, check that a proper error is returned when attempting to interface with remote location.

Another nice to have: linting reusable component requirements to check that `fondant[gcp, aws, azure]` is included (maybe another ticket)

## Documentation

Update [custom_component.md requirement section](https://github.com/ml6team/fondant/blob/main/docs/custom_component.md#:~:text=py%22%5D-,Requirements.txt,-A%20requirements.txt) with instructions on the required extras. 

## Dependent features

None 

## Additional Notes

PhilippeMoussalli:
fixes #210 

Pyarrow was also removed from many of the components' requirements since it is a Fondant dependency


components/caption_images/Dockerfile
```diff
@@ -12,7 +12,7 @@ RUN pip3 install --no-cache-dir -r requirements.txt
 # Install Fondant
 # This is split from other requirements to leverage caching
 ARG FONDANT_VERSION=main
-RUN pip3 install git+https://github.com/ml6team/fondant@${FONDANT_VERSION}
+RUN pip3 install git+https://github.com/ml6team/fondant@${FONDANT_VERSION}#egg=fondant[aws, azure, gcp]
```
RobbeSneyders, line 15: Thanks @PhilippeMoussalli 

In the pip documentation ([example 7])(https://pip.pypa.io/en/stable/cli/pip_install/#examples), they use the following format:
```
pip install fondant[aws,azure,gcp] @ git+https://github.com/ml6team/fondant@${FONDANT_VERSION}
```

Which I think is a bit more readable and more future-proof, as eggs have been deprecated as far as I'm aware.

PhilippeMoussalli: oh good to know. I was not aware. I'll change them

RobbeSneyders (approved):

PR #282, opened 2023-07-10
[Commoncrawl pipeline] Add component extract free-to-use images

shayorshay:
This 3rd component extracts the image url, alt text and license metadata from the webpage url and html code.


examples/pipelines/commoncrawl/components/extract_free_to_use_images/src/main.py
```diff
@@ -0,0 +1,94 @@
...
+                logger.info(f"Found {len(unique_images)} unique images.")
+
+                return unique_images
+
+    return None
+
+
+class ExtractFreeToUseImages(PandasTransformComponent):
+    def setup(self, *args, **kwargs):
+        pass
```
RobbeSneyders, line 51: I don't think you need to add this method if you don't implement it.

examples/pipelines/commoncrawl/components/extract_free_to_use_images/src/main.py
```diff
@@ -0,0 +1,94 @@
...
+        df = pd.DataFrame(
+            flattened_results,
+            columns=[
+                ("image", "image_url"),
+                ("image", "alt_text"),
+                ("image", "webpage_url"),
+                ("image", "license_type"),
+                ("image", "license_location"),
+            ],
+        )
```
RobbeSneyders, line 87: Can we rewrite this using the `.apply()` method?

examples/pipelines/commoncrawl/components/extract_free_to_use_images/src/main.py
```diff
@@ -0,0 +1,94 @@
...
+def get_image_info_from_webpage(webpage_url: str, webpage_html: str) -> List[str]:
+    """Extracts image urls and license metadata from the parsed html code.
+    Args:
+        webpage_url: The url of the webpage.
+        webpage_html: The html content of the webpage.
+    Returns:
+        A list of image urls and license metadata.
+    """
+
+    soup = BeautifulSoup(webpage_html, "html.parser")
```
PhilippeMoussalli, line 25: perhaps this should go in the `setup()` method? Seems like you're now initializing this on every function call

examples/pipelines/commoncrawl/components/extract_free_to_use_images/fondant_component.yaml
```diff
@@ -0,0 +1,25 @@
+name: Extract image url and license from commoncrawl
+description: Component that extracts image url and license metadata from a dataframe of webpage urls and html codes
+image: ghcr.io/ml6team/extract_image_licenses:latest
```
PhilippeMoussalli, line 3: can you rename the component folder to also be `extract_image_licenses` for consistency?

examples/pipelines/commoncrawl/components/extract_free_to_use_images/src/main.py
shayorshay: doesn't look like `BeautifulSoup` has a a parser function without the constructor so I'm leaving it as it is

examples/pipelines/commoncrawl/components/extract_free_to_use_images/src/main.py
RobbeSneyders: Have you been able to have a look at this @shayorshay?

shayorshay: yes, its done. could you take a look?

examples/pipelines/commoncrawl/components/extract_image_licenses/src/main.py
```diff
@@ -0,0 +1,88 @@
...
+        df = (
+            df.apply(
+                lambda row: get_image_info_from_webpage(
+                    row[("webpage", "url")], row[("webpage", "html")]
+                ),
+                axis=1,
+                result_type="expand",
+            )
+            .explode(0)
+            .apply(pd.Series)
```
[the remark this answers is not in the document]
shayorshay: get_image_info_from_webpage` returns a nested list of image urls per webpage. this flattens both and transforms the elements into rows. not sure if there's a cleaner way to do this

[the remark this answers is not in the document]
RobbeSneyders: Can you check the typing of that function and the one it depends on? It says it returns a `List[str]`, which I don't think is correct then.

examples/pipelines/commoncrawl/components/extract_image_licenses/src/main.py
```diff
@@ -0,0 +1,88 @@
...
+        )
+
+        df = df.dropna().reset_index(drop=True)
```
RobbeSneyders, line 73: Why are you resetting the index?

shayorshay: The final dataframe keeps the id of the wepbpage but on a second thought, I think this is not needed since the `image_url` is the id we want anyway. I'll remove it

RobbeSneyders (approved):
Thanks @shayorshay! Would be great if you could add a readme for the different components in your pipeline as well.

PR #283, opened 2023-07-10
Run tests against fondant package with tox

RobbeSneyders:
This PR contains a couple of changes to make sure we test against the built fondant package instead of against the raw source code. This is best practice, but I also need this for the next PR I'll open as there the tests will fail if they are not run against the built package.

- Introduce tox, which makes it easy to build the fondant package before running our tests
- Move the fondant source code into a `src` folder, so we don't accidentally test against the source code. See these articles for more info (https://hynek.me/articles/testing-packaging/, https://blog.ganssle.io/articles/2019/08/test-as-installed.html)
- Ruff now recognizes `fondant` as a third party package in our components and tests, and has reorganized the imports accordingly

Each of these changes is contained in a separate PR.


tox.ini
```diff
@@ -0,0 +1,39 @@
...
+    3.10: pre-commit,check-licenses, py310
+    3.11: py311
+
+[testenv:pre-commit]
+deps=pre-commit
+commands=pre-commit run --all-files --show-diff-on-failure
+
+[testenv:check-licenses]
+deps=
+    liccheck==0.7.3
```
PhilippeMoussalli, line 21: can we have soft dependency versions here like we had in `pyproject.toml`?  (same for pytest, and pytest-cov)

tox.ini
```diff
@@ -0,0 +1,42 @@
+[tox]
+isolated_build = True
+envlist =
+    pre-commit
```
PhilippeMoussalli, line 4: do these dependencies only get built when running tox then? does it automatically invoke `pip install` on them since we no longer have them in the `pyproject.toml`

PhilippeMoussalli (approved):
Thanks Robbe! 

Left a few questions

tox.ini
RobbeSneyders: Yes, installs in this file are only installed by tox when executing the related environment. If a dependency is only needed for a single environment type, I moved them from `pyproject.toml` to here, and I added `tox` to `pyproject.toml`.

tox.ini
RobbeSneyders: Good point. Updated.

tox.ini
PhilippeMoussalli: alright thanks for the clarification

PR #285, opened 2023-07-10
Don't use from_registry for generic components

### LINKED ISSUE #251: Don't use from_registry for generic components
Summary of discussion in #224

Currently the operation for a generic component (a reusable component that dynamically takes a component specification), needs to be created as follows:

```python
component_op = ComponentOp.from_registry(
    name="generic_component",
    component_spec_path="components/custom_generic_component/fondant_component.yaml",
)
```

But the name isn't actually used. Instead the component specification contains a reference to the generic image that it should use. So ideally, we could create it as follows instead:

```python
component_op = ComponentOp(
    component_spec_path="components/custom_generic_component/fondant_component.yaml",
)
```

One issue here is that we currently use the way a `ComponentOp` is initialized to decide if the local runner should add a build step for it. Since this looks like a custom component, it will try to build it, while there is no Dockerfile or code available at the specified location.

We should therefore find a different mechanism to decide if a component needs to be built. If we agree that a component should have a fixed directory structure, we can instead detect if a Dockerfile is present. We currently already use the same directory structure for each component, which is:

```
|- my_component
   |- src
      |- main.py
   |- Dockerfile
   |- fondant_component.yaml
   |- README.md
   |- requirements.txt
```

If we fix this structure, we can also deduct the location of the component specification and further simplify the definition of a component to:

```python
component_op = ComponentOp(
    path="components/custom_generic_component",
)
```

Such a fixed directory structure also provides opportunities to implement a component generator / cookie-cutter in the future.
```[tasklist]
### Tasks
- [ ] https://github.com/ml6team/fondant/issues/252
- [ ] https://github.com/ml6team/fondant/issues/253
```

RobbeSneyders:
Fixes #251, #252, #253

Before this PR, the operation for a generic component (a reusable component that dynamically takes a component specification), needed to be created as follows:

```python
component_op = ComponentOp.from_registry(
    name="generic_component",
    component_spec_path="components/custom_generic_component/fondant_component.yaml",
)
```

But the provided name wasn't actually used, since the component specification already contains a reference to the reusable image that it should use. Now we can define both custom and generic components as follows:

```python
component_op = ComponentOp(
    component_dir="components/custom_generic_component",
)
```

There is still a difference in how we want to handle custom and generic components though. Custom components should be built by the local runner, while generic components should not, since they use a reusable image. To make this differentiation, we now simply check if a `Dockerfile` is present in the provided `Component_dir`. This will be the case for a custom component, but not for a generic component.

---

This has a nice implicit result for reusable components as well, which can still be defined as:

```python
component_op = ComponentOp.from_registry(
    name="reusable_component",
)
```

Which `fondant` now resolves to:
```python
component_op = ComponentOp(
    component_dir="{fondant_install_dir}/components/custom_generic_component",
)
```

I added a change to this PR which no longer packages the reusable component code with the `fondant` package, but only the component specifications, as those are all that is needed since they contain a reference to the reusable image on the registry. This means that the `component_dir` above doesn't contain a `Dockerfile` when `fondant` is installed from PyPI, but does when you locally install `fondant` using `poetry install`. So the local runner doesn't build reusable components when users install `fondant` from PyPI, but it does when you're working on the `fondant` repo, which is useful for us `fondant` developers.

---

The only thing we still need is an option on the runner to provide `build_args`, so we can pass in the `fondant` version to build into the image. But I'll open a separate PR for that.


tests/test_pipeline.py
```diff
@@ -217,26 +211,19 @@ def test_defining_reusable_component_op_with_custom_spec():
...
-    load_from_hub_op_custom_spec = ComponentSpec(yaml_file_to_dict(custom_spec_path))
-
-    assert load_from_hub_op_custom_op.component_spec == load_from_hub_op_custom_spec
-    assert load_from_hub_op_default_op.component_spec == load_from_hub_op_default_spec
-    assert load_from_hub_op_default_op.component_spec != load_from_hub_op_custom_spec
-    assert load_from_hub_op_custom_op.component_spec != load_from_hub_op_default_spec
+    assert (
+        load_from_hub_custom_op.component_spec
+        == load_from_hub_default_op.component_spec
+    )
```
RobbeSneyders, line 226: Some of these checks were trivial but harder to replicate, so only kept the one that is non-trivial.

src/fondant/pipeline.py
```diff
@@ -49,44 +49,41 @@ class ComponentOp:
...
-        return ComponentSpec.from_file(self.component_spec_path)
-
-    def extend_arguments(self):
-        """Add the component specification to the arguments if not already present."""
-        if not self.arguments.get("component_spec"):
-            self.arguments["component_spec"] = json.dumps(
-                self.component_spec.specification,
-            )
+    def dockerfile_path(self) -> t.Optional[Path]:
+        path = self.component_dir / "Dockerfile"
```
GeorgesLorre, line 79: `Dockerfile` is the default and should be used but I have seen people use `dockerfile` before (and it also works) so maybe we should also check for that ?

PhilippeMoussalli (commented):
Thanks Robbe! 

I just have a few doubts about whether this can be clear for the end user. Right now we have: 

1) Custom components 
```python
ComponentOp(component_dir="")
```

2) Resuable components 

```python
ComponentOp.from_registry(name="")
```

3) Generic components 

Can be considered to be a hybrid of custom component (require a customized component spec) and a reusable component (images are already built and available in the registry). 

However, right now it seems like we're defining them as if they're fully custom
```python
ComponentOp(component_dir="")
```
At first glance, it seems like the directory should contain all the component files (src_code, Dockerfile, ...) similar to the custom components where they are used to build the images for the local runner. But in fact, we're just looking for the spec file within the specified directory. 

Wondering whether it wouldn't better to have generic components defined more explicitly by just passing in the component spec path since the users will anyway have to create a custom component spec for that component: 
```python
ComponentOp.from_spec(component_spec_path="")
```

I think it would be nice to have separate constructors for each component type, Wdyt?

RobbeSneyders:
> Did you check the docs for needed changes ? I think at least the getting-started might need some updates aswel.

Not yet, I'll have a look.

---

> I think it would be nice to have separate constructors for each component type, Wdyt?

I partially agree. I agree that we need to make sure it is clear to the user, but on the other hand, there's not really that big of a difference between a custom and a generic component. Both only need a component specification referring to an existing image to work.

You could for instance define a custom component as follows and it would work as well:
```
|- my_image
|  |- src
|  |  |- main.py
|  |- Dockerfile
|- components
|  |- custom_component
|     |- fondant_component.yaml
|        "image: my_image:tag"
|- pipeline.py
   "component_op = ComponentOp.from_spec(path="components/custom_component/fondant_component.yaml")
```

So while splitting the constructors kind of makes sense logically, it actually creates two ways to achieve the same thing. You could use both constructors to create either a custom or a generic component.

```python
component_op(component_dir="components/custom_component")
component_op.from_spec(component_spec_path="components/custom_component/fondant_component.yaml")
```

The only functionality for which we need more than the component specification, is to automatically build the images, which we currently only do in the local runner. 
(_I do think it would be nice to extend this behavior to other runners as well in the future though, and possibly even provide a `fondant build` command to help users build their components._)

That's the only reason why it might make sense to split the constructors, but this doesn't really matter for your pipeline definition, only execution. So I would move this responsibility to the runner and let it decide when to build (based on presence of a `Dockerfile`, command flags, ...).

So I would select only one of the two interfaces, and then I see the following (dis)advantages:
| `ComponentOp(component_dir="components/custom_component")`                                    | `ComponentOp.from_spec("components/custom_component/fondant_component.yaml")`                         |
|-----------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------|
| Less clear that users can use a generic component by only defining a `fondant_component.yaml` | More clear that only the `fondant_component.yaml` is needed to create a custom or generic component   |
| No need to add `fondant_component.yaml` since the name is fixed                               | Defining `fondant_component.yaml` each time seems superfluous, so then we might not want to keep it fixed |

I don't have a strong preference, but since we have to choose, I would go with:
```python
componentOp(component_dir="components/custom_component")
```
and clearly document the expected structure of the `component_dir`:
```
|- fondant_component.yaml (required)
|- Dockerfile (optional)
```
_(All the other files are actually don't matter to `fondant` at the moment, but are only used in the `Dockerfile`)_

Happy to hear other opinions.

src/fondant/pipeline.py
RobbeSneyders: I don't think we have to, since we can expect these components to be implemented for `fondant` specifically. So we can impose our own expectations. I don't think that's a bad thing to do here, since we adhere to the standard and keep things simple.

scripts/pre-build.sh
```diff
@@ -9,5 +9,5 @@ root_path=$(dirname "$scripts_path")
 
 pushd "$root_path"
 rm -rf src/fondant/components
-cp -r components/ src/fondant/
+find components/ -type f | grep -i yaml$ | xargs -i cp --parents {} src/fondant/
```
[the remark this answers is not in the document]
RobbeSneyders: Can you validate this on mac @GeorgesLorre?

docs/components.md
```diff
@@ -0,0 +1,148 @@
...
+
+component_op = ComponentOp.from_registry(
+  name="reusable_component",
+  arguments={
+    "arg": "value"
+  }
+)
+```
+
+??? "fondant.pipeline.ComponentOp.from_registry"
```
PhilippeMoussalli, line 40: are those meant to end up here? there are quite a few of them

GeorgesLorre:
Thanks for the detailed explanation and options:

I suggest we go for (also not a strong preference)
`componentOp(component_dir="components/custom_component")
`

The key will be documenting it very well and maybe expanding the logging with
`Found Dockerfile for component x: building image`

PhilippeMoussalli (approved):
Thanks Robbe! Looks good overall and the documentation is clear. 

Just a few minor things to adjust in the documentation

docs/components.md
```diff
@@ -0,0 +1,148 @@
...
+```
+
+You can add a generic component to your pipeline by creating a `ComponentOp` and passing in the path
+to the directory containing your custom `fondant_component.yaml`.
+
+```python title="pipeline.py"
+from fondant.pipeline import ComponentOp
+
+component_op = ComponentOp(
+  component_dir="components/custom_component",
```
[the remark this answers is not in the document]
RobbeSneyders: Good catch

docs/components.md
PhilippeMoussalli: looks neat ✨

PR #289, opened 2023-07-11
Add defaults to components

PhilippeMoussalli:
PR that add default arguments to component yaml. This makes it easier to get started with default values for reusable components.


RobbeSneyders (approved):
Thanks @PhilippeMoussalli! I think you need to update the failing test.

PR #291, opened 2023-07-12
Add kfp compiler

GeorgesLorre:
Untested -> looking for feedback on the structure not the content


src/fondant/pipeline.py
```diff
@@ -303,240 +303,3 @@ def _validate_pipeline_definition(self, run_id: str):
...
-            The pipeline id.
-        """
-        return self.client.get_pipeline_id(pipeline_name)
-
-    def get_pipeline_version_ids(self, pipeline_id: str) -> t.List[str]:
-        """Function that returns the versions of a pipeline given a pipeline id."""
-        pipeline_versions = self.client.list_pipeline_versions(pipeline_id).versions
-        return [version.id for version in pipeline_versions]
-
-    def delete_pipeline(self, pipeline_name: str):
```
PhilippeMoussalli, line 451: Where do you see all the other functionalities that are not related to compiling and running? user might still want to list pipeline runs, delete pipelines, ...

src/fondant/compiler.py
```diff
@@ -183,3 +184,61 @@ def _generate_spec(self, pipeline: Pipeline, extra_volumes: list) -> dict:
...
             "services": services,
         }
+
+
+class KubeFlowCompiler(Compiler):
+    """Compiler that creates a Kubeflow pipeline spec from a pipeline."""
+
+    def _resolve_imports(self):
+        """Resolve imports for the Kubeflow compiler."""
+        try:
```
PhilippeMoussalli, line 194: I think you can just do 

if is_kfp_available():
    import kfp
    from kfp import dsl

since the function will check for those things for you

src/fondant/compiler.py
```diff
@@ -183,3 +184,61 @@ def _generate_spec(self, pipeline: Pipeline, extra_volumes: list) -> dict:
...
+            from kfp import dsl
+        except ImportError:
+            raise ImportError(
+                "You need to install kfp to use the Kubeflow compiler, "
+                / "you can install it with `poetry install --extras kfp`"
+            )
+
+    def compile(
+        self,
+        pipeline: Pipeline,
```
PhilippeMoussalli, line 205: I think the `Pipeline` is the unifying interface between all runners right? if so, would add it to the abstract class

PhilippeMoussalli (commented):
I like the overall structure and separation between the Compiler and Runner! 

I think there are still some things that don't fall quite between like Client related functionalities. Not the highest priority but we should figure out where those land

src/fondant/compiler.py
RobbeSneyders: We should no longer need `is_kfp_available` with this approach. This is very nice change, since you'll no longer need to install kfp when using the local runner.

I think we can just move this to `__init__()` instead of this special function though?

src/fondant/compiler.py
```diff
@@ -183,3 +184,61 @@ def _generate_spec(self, pipeline: Pipeline, extra_volumes: list) -> dict:
...
+
+    def _resolve_imports(self):
+        """Resolve imports for the Kubeflow compiler."""
+        try:
+            import kfp
+            from kfp import dsl
+        except ImportError:
+            raise ImportError(
+                "You need to install kfp to use the Kubeflow compiler, "
+                / "you can install it with `poetry install --extras kfp`"
```
RobbeSneyders, line 200:
```suggestion
                / "you can install it with `pip install fondant[kfp]`"
```

`poetry install` is only for fondant developers. Users can also use poetry, but then they'll install their own project and need to indicate the fondant extras with the fondant dependency in their `pyproject.toml`.

src/fondant/runner.py
```diff
@@ -26,3 +29,30 @@ def run(cls, input_spec: str, *args, **kwargs):
...
+class KubeflowRunner(Runner):
+    @abstractmethod
+    def _resolve_imports(self):
+        """Resolve imports for the Kubeflow compiler."""
+        try:
+            import kfp
+        except ImportError:
+            raise ImportError(
+                "You need to install kfp to use the Kubeflow compiler, "
+                / "you can install it with `poetry install --extras kfp`"
```
RobbeSneyders, line 43:
```suggestion
                "You need to install kfp to use the Kubeflow runner, "
                / "you can install it with `pip install fondant[kfp]`"
```

src/fondant/pipeline.py
RobbeSneyders: Not sure if we need to wrap this functionality for every platform we want to run on. At least at this point in time, I would be ok with removing it and having the user work with the underlying platform directly.

src/fondant/runner.py
```diff
@@ -26,3 +29,30 @@ def run(cls, input_spec: str, *args, **kwargs):
...
+            )
+
+    def run(cls, input_spec: str, host: str, *args, **kwargs):
+        """Run a kubeflow pipeline."""
+        cls._resolve_imports()
+        client = kfp.Client(host=host)
```
RobbeSneyders, line 49: This probably doesn't change a lot to the current functionality, but it would make more sense to me to instantiate this in `__init__()`.

RobbeSneyders (commented):
Thanks @GeorgesLorre!

The lines along you split the behavior look good to me!

src/fondant/compiler.py
PhilippeMoussalli: https://github.com/ml6team/fondant/blob/86867a5e777c93a7ca0d982f263d66c38abc3e8f/src/fondant/import_utils.py#L24C1-L25C1

src/fondant/compiler.py
PhilippeMoussalli: Created a PR to remove obsolete packages

https://github.com/ml6team/fondant/pull/293

src/fondant/pipeline.py
PhilippeMoussalli: makes sense, we'll re-add it later if we notice that it's needed

src/fondant/compiler.py
```diff
@@ -207,3 +207,108 @@ def _generate_spec(
...
+
+    def _resolve_imports(self):
+        """Resolve imports for the Kubeflow compiler."""
+        try:
+            import kfp
+
+            self.kfp = kfp
+        except ImportError:
+            msg = """You need to install kfp to use the Kubeflow compiler,\n
+                     you can install it with `pip install --extras kfp`"""
```
RobbeSneyders, line 226:
```suggestion
                     you can install it with `pip install fondant[pipelines]`"""
```

Although we probably want to split this up for the different runners so it becomes fondant[kfp].

src/fondant/compiler.py
```diff
@@ -207,3 +207,108 @@ def _generate_spec(
...
+    ) -> None:
+        """Compile a pipeline to Kubeflow pipeline spec and save it to a specified output path.
+
+        Args:
+            pipeline: the pipeline to compile
+            output_path: the path where to save the Kubeflow pipeline spec
+        """
+        self.pipeline = pipeline
+        self.pipeline.sort_graph()
+        self.pipeline._validate_pipeline_definition("{{workflow.name}}")
```
RobbeSneyders, line 244: This should be done in the `DockerCompiler` as well, no?

src/fondant/pipeline.py
```diff
@@ -18,8 +17,6 @@
 from fondant.schema import validate_partition_number, validate_partition_size
 
 if is_kfp_available():
```
RobbeSneyders, line 19: Ideally we can get rid of this and make this file completely runner independent.

src/fondant/runner.py
```diff
@@ -26,3 +29,34 @@ def run(self, input_spec: str, *args, **kwargs):
...
+
+class KubeflowRunner(Runner):
+    def __init__(self):
+        self._resolve_imports()
+
+    @abstractmethod
+    def _resolve_imports(self):
+        """Resolve imports for the Kubeflow compiler."""
+        try:
+            global kfp
```
RobbeSneyders, line 42: Can we assign to `self.kfp` here as in the compiler?

src/fondant/runner.py
```diff
@@ -26,3 +29,34 @@ def run(self, input_spec: str, *args, **kwargs):
...
+        # client = kfp.Client(host=host)
+        # # TODO add logic to see if pipeline exists
+        # pipeline_spec = client.run_pipeline(
+        #     experiment_id=experiment.id,
+        #     job_name=run_name,
+        #     pipeline_package_path=pipeline.package_path,
+        # )
+
+        # pipeline_url = f"{self.host}/#/runs/details/{pipeline_spec.id}"
+        # logger.info(f"Pipeline is running at: {pipeline_url}")
```
RobbeSneyders, line 62: Is there a reason this is still commented out?

tests/test_pipeline.py
```diff
@@ -169,12 +165,12 @@ def test_invalid_pipeline_dependencies(
...
+        ("example_3", ["first_component", "second_component"]),
     ],
 )
 def test_invalid_pipeline_compilation(
     default_pipeline_args,
     invalid_pipeline_example,
-    tmp_path,
 ):
     """
     Test that an InvalidPipelineDefinition exception is raised when attempting to compile
```
RobbeSneyders, line 176: You're no longer compiling in the test :)

RobbeSneyders (commented):
Thanks @GeorgesLorre!

The general direction looks good. Left some comments.

src/fondant/components
RobbeSneyders, line 1: This should not be removed

src/fondant/runner.py
GeorgesLorre: should not have committed this, will be taken up in a separate PR addressing the kfp runner

src/fondant/pipeline.py
PhilippeMoussalli: This is only needed for setting `p_volumes` at the `componentOp` level, that was only needed when we were initially planning to download the Laion dataset locally to use it. I think we can remove both `ephemeral_storage_size` and `p_volumes` from the ComponenOp. We can always re-add them later if we notice they're needed

PhilippeMoussalli (commented):
Thanks Georges! Looks much cleaner already ✨

GeorgesLorre:
@RobbeSneyders updated!

src/fondant/compiler.py
```diff
@@ -207,3 +207,108 @@ def _generate_spec(
...
+
+    def __init__(self):
+        self._resolve_imports()
+
+    def _resolve_imports(self):
+        """Resolve imports for the Kubeflow compiler."""
+        try:
+            import kfp
+
+            self.kfp = kfp
```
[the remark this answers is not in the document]
GeorgesLorre: it avoids having to use a global, this is cleaner imo

GeorgesLorre:
@PhilippeMoussalli Ok I'll remove the `ephemeral_storage_size` and `p_volumes` in a new PR. This would indeed be cleaner.

RobbeSneyders (approved):
Thanks! Can you just update the error message for the kfp import before merging?

PR #294, opened 2023-07-13
Update pre-commit config with new folder structure

GeorgesLorre:


RobbeSneyders (commented):
Thanks @GeorgesLorre. We should change it for mypy as well (line 47).

GeorgesLorre:
> Thanks @GeorgesLorre. We should change it for mypy as well (line 47).

fixed

RobbeSneyders (approved):
Thanks!

PR #295, opened 2023-07-13
[LLM pipeline] MinHash generation for deduplication

mrchtr:
This component generates MinHashes of text. 
The MinHash similarity will be used to determine duplicated text passages.


components/minhash_generator/src/main.py
```diff
@@ -0,0 +1,58 @@
...
+from fondant.component import PandasTransformComponent
+from nltk.util import ngrams
+
+logger = logging.getLogger(__name__)
+
+
+class MinHashGeneratorComponent(PandasTransformComponent):
+    """Component generates minhashes of text."""
+
+    @staticmethod
```
PhilippeMoussalli, line 15: can we move those methods above and only have the class implement the `transform`?  This is a bit similar to we have it for other components [example](https://github.com/ml6team/fondant/blob/main/components/image_cropping/src/main.py#L15), it makes the layout a bit nicer than having to write static methods

components/minhash_generator/src/main.py
```diff
@@ -0,0 +1,58 @@
...
+    def create_shingles(text):
+        """Creates text shingles that will be used for the hash generation."""
+        # Split text into words
+        words = text.split()
+
+        # Generate shingles of size 3 using nltk's ngrams function
```
PhilippeMoussalli, line 21: any reason for the having a fixed size of 3? should this be a component parameter?

components/minhash_generator/src/main.py
```diff
@@ -0,0 +1,58 @@
...
+from nltk.util import ngrams
+
+logger = logging.getLogger(__name__)
+
+
+class MinHashGeneratorComponent(PandasTransformComponent):
+    """Component generates minhashes of text."""
+
+    @staticmethod
+    def create_shingles(text):
```
PhilippeMoussalli, line 16: Could you please add typehints here?

components/minhash_generator/src/main.py
```diff
@@ -0,0 +1,58 @@
...
+        """Creates text shingles that will be used for the hash generation."""
+        # Split text into words
+        words = text.split()
+
+        # Generate shingles of size 3 using nltk's ngrams function
+        return list(ngrams(words, 3))
+
+
+    @staticmethod
+    def compute_minhash(shingles):
```
PhilippeMoussalli, line 26: same comments as above (typehint, move the function above)

components/minhash_generator/tests/component_test.py
```diff
@@ -0,0 +1,38 @@
...
+    spec = ComponentSpec.from_file("../fondant_component.yaml")
+
+    component = MinHashGeneratorComponent(
+        spec,
+        input_manifest_path="./dummy_input_manifest.json",
+        output_manifest_path="./dummy_input_manifest.json",
+        metadata={},
+        user_arguments={},
+    )
+    component.setup()
```
PhilippeMoussalli, line 29: is this needed? does not seem like you have anything defined in the `setup()` method

PhilippeMoussalli (commented):
Thanks @mrchtr  :) 

Looks good overall and thanks for including tests. We should do this for the other components. 

Left a few minor comments.

components/minhash_generator/src/main.py
mrchtr: Size of 3 is common. But I agree that is indeed better to make it configurable.

PhilippeMoussalli (approved):
Thanks @mrchtr!

PR #297, opened 2023-07-17
Add component type to spec

PhilippeMoussalli:
PR that infers the component type based specification in order to explicitly define the component type: 

`load_component`:  produces output manifest only 
`transform_component`: consumes input manifest and produces output manifest 
`write_component`: consumes input manifest only

Shared fondant arguments between the component are the `metadata` and `component_spec`. The rest of the arguments are user defined arguments. 

I decided to implement this change since the kubeflow pipelines were failing in the write components because they had an `output_artifact` defined but no data was written to it. The expected arguments per component types are also more explicitly defined.

This also unlocks building more complex pipelines that don't assume that the first and last component are the read/write component (e.g. user might want to write their data to an external data source before the last step of the pipeline).


PhilippeMoussalli:
> Thanks @PhilippeMoussalli!
> 
> Is there a way to infer this without adding it to the component spec? I think we should try to keep it as simple as possible for the users.

Yeah indeed, I think we can do so by just looking at the component spec and the `consumes` and `produces` field. I'll update it accordingly

src/fondant/component_spec.py
```diff
@@ -102,6 +167,23 @@ class ComponentSpec:
...
+        """Function that returns the component type based on the component specification."""
+        consumes = self._specification.get("consumes", False)
+        produces = self._specification.get("produces", False)
+
+        if not consumes and produces:
+            _type = ComponentType.LOAD
+        elif consumes and produces:
+            _type = ComponentType.TRANSFORM
+        elif consumes and not produces:
+            _type = ComponentType.WRITE
```
RobbeSneyders, line 182: I think this is the case for filtering components as well. Eg. https://github.com/ml6team/fondant/blob/main/components/filter_comments/fondant_component.yaml

RobbeSneyders: Do we need to infer it from the spec only or can we also use the `Component` class? Because there it is already indicated if it is a Load, Transform, or Write component.

RobbeSneyders: Nvm, I don't think that's a good direction. You're using this info in the compiler and pipeline, where we're working with the spec only.

RobbeSneyders:
The only way I see to detect the difference between a write component and a filter component is by requiring the write component to specify `additionalSubsets:false` in its `produces` section.

We then get:
**Filter component**
```yaml
name: ...
description: ...
image: ...

consumes:
    ...

args:
    ...
```

**Write component**
```yaml
name: ...
description: ...
image: ...

consumes:
    ...

produces:
    additionalSubsets: false

args:
    ...
```

It's not great either, but the benefit is that you only have a special case for write components, which should be quite rare, instead of an extra field for every component.

You can discuss which way to go with the rest of the team.

PhilippeMoussalli:
Closed in favor of #347

PR #299, opened 2023-07-17
Add fsspec as explicit dependency

RobbeSneyders:
We use `fsspec` [directly](https://github.com/ml6team/fondant/blob/main/src/fondant/manifest.py#L10) in fondant, so we should add it as a dependency.


PhilippeMoussalli (approved):
Thanks Robbe!

Issue #301, opened 2023-07-18
Add component tests


RobbeSneyders:
Our reusable components should be tested at least for every release.

satishjasthi:
@RobbeSneyders, Shall I take up this issue and add test cases for all existing components

mrchtr:
# Problem Statement

Currently, there are no component tests for the reusable components, which can lead to the incorporation of unexpected
behavior into reusable components during a release.

To avoid this, we should add component tests that can be executed before each release.

Furthermore, by establishing a well-defined component test strategy, we can expedite the development process of
individual components.

The process of adding component tests will be documented to ensure clarity and consistency in the testing procedures.

# Proposed Approach

To ensure consistency across tests, I would propose adding an `AbstractComponentTest` or a `TestExecutor` class.
This shared utility will offer a common way to initialise/execute a component test.

During each test, we can focus on testing the transform method (the core business logic) of a single component.

By utilising this shared testing class, new tests for different components can be implemented more efficiently.

## Parameterize Test definition by using json fixtures

To simplify the process even more, we could define a default test case structure based on a
parameterized fixture. This approach will enable users to add JSON files for their test configurations, specifying the
expected input, output dataframe, and component arguments.

A `TestExecuter` could initialize the parameterized tests using the
JSON files, executing each test case automatically.

As a user, you will only need to provide the necessary JSON files with the required configurations for your tests. The
testing framework will take care of initializing the tests and comparing the actual output with the expected output for
each test case.

E.g.

```json
{
  "user_arguments": {
    "model_name": "..."
  },
  "input_data": <expected input dataframe>,
  "expected_output": <expected output dataframe>
}
```

Testing different components becomes straightforward and more accessible to users, reducing
the effort required to set up and execute tests for individual components.

This setup can probably not be used for complexer components.

## Run component test in the release pipeline

Since each component may have specific testing requirements, we may need to modify the component's Dockerfile to adapt
it to the officially [proposed way](https://docs.docker.com/language/java/run-tests/) of running tests.
This will help ensure that the testing environment aligns with the specific needs of each component's test suite.

```Dockerfile
FROM --platform=linux/amd64 python:3.8-slim as base

RUN apt-get update && \
...

FROM base as tests
COPY tests/ tests/
RUN ["python", "-m", "pytest", "tests/"]

FROM base
WORKDIR /component/src
ENTRYPOINT ["python", "main.py"]
```

This implementation would be downwards compatible and allows to continue building and running the components using the
[...]

Issue #300, opened 2023-07-18
Hide executor from users


RobbeSneyders:
Each component now has the following section at the bottom of their `src/main.py` file:

```python
if __name__ == "__main__":
    executor = PandasTransformExecutor.from_args()
    executor.execute(MyComponent)
```

Since you always need to match the right executor class with the implemented component class, this is something fondant can do for the users.

Ideally we can remove the whole `if __name__ == "__main__"` block by calling a fondant module / command as entrypoint instead of the `src/main.py` module.

PR #302, opened 2023-07-18
Split component implementation and execution

### LINKED ISSUE #203: Update components to implement the PandasTransformComponent

### LINKED ISSUE #257: Split general component behavior from user implementation
Currently users implement components by subclassing one of the provided component classes, which themselves provide a lot of general component behavior, such as data loading, validating, and writing.

Instead, we could split this by letting the user implement their components by implementing a class with a specific interface which is passed to a general component class.

Current:
```python
from fondant import PandasTransformComponent 

class MyComponent(PandasTransformComponent):

    def setup(self, kwarg1, kwarg2):
        ...

    def transform(self, dataframe):
        ...

component = MyComponent.from_args()
component.run()
```

Future:
```python
from fondant import Component

class MyComponent:

    def __init__(self, kwarg1, kwarg2):
        ...

    def transform(self, dataframe):
        ...

Component(MyComponent).run()
```

We can still provide an interface which the user can subclass, for instance using a [Protocol](https://peps.python.org/pep-0544/).

This has some advantages:
- Component implementations can use `__init__` instead of `setup` which will be more familiar to users
- It's easier to test the custom component implementation, as no dummy arguments need to be provided for the general component behavior (see https://github.com/ml6team/fondant/pull/232#discussion_r1250515346)

RobbeSneyders:
This PR follows up on the PoC presented in #268

---

Fixes #257 

It splits the implementation and execution of components, this has some advantages:

- Pandas components can use `__init__` instead of setup, which is probably more familiar to users
- Other components can use `__init__` as well instead of receiving all arguments to their transform or equivalent method, aligning implementation of different component types
- Component implementation and execution should be easier to test separately

I borrowed the executor terminology from KfP.

---

Fixes #203 

Since I had to update all the components, I also switched some of them to subclass `PandasTransformComponent` instead of `DaskTransformComponent`.

---

These changes open some opportunities for additional improvements, but I propose to tackle those as separate PRs as this PR is already quite huge due to all the changes to the components.

- [ ] #300
- [ ] #301

### DIFF
--- a/components/download_images/src/main.py
+++ b/components/download_images/src/main.py
@@ -12,6 +12,7 @@ import urllib
 
 import dask.dataframe as dd
 from fondant.component import DaskTransformComponent
+from fondant.executor import DaskTransformExecutor
 from resizer import Resizer
 
 logger = logging.getLogger(__name__)
@@ -88,21 +89,19 @@ def download_image_with_retry(
 class DownloadImagesComponent(DaskTransformComponent):
     """Component that downloads images based on URLs."""
 
-    def transform(
-            self,
-            dataframe: dd.DataFrame,
-            *,
-            timeout: int,
-            retries: int,
-            image_size: int,
-            resize_mode: str,
-            resize_only_if_bigger: bool,
-            min_image_size: int,
-            max_aspect_ratio: float,
-    ) -> dd.DataFrame:
-        """Function that downloads images from a list of URLs and executes filtering and resizing
+    def __init__(self,
+                 *_,
+                 timeout: int,
+                 retries: int,
+                 image_size: int,
+                 resize_mode: str,
+                 resize_only_if_bigger: bool,
+                 min_image_size: int,
+                 max_aspect_ratio: float,
+                 ):
+        """Component that downloads images from a list of URLs and executes filtering and resizing.
+
         Args:
-            dataframe: Dask dataframe
             timeout: Maximum time (in seconds) to wait when trying to download an image.
             retries: Number of times to retry downloading an image if it fails.
             image_size: Size of the images after resizing.
@@ -114,8 +113,9 @@ class DownloadImagesComponent(DaskTransformComponent):
         Returns:
             Dask dataframe
         """
-        logger.info("Instantiating resizer...")
-        resizer = Resizer(
+        self.timeout = timeout
+        self.retries = retries
+        self.resizer = Resizer(
             image_size=image_size,
             resize_mode=resize_mode,
             resize_only_if_bigger=resize_only_if_bigger,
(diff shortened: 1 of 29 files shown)


RobbeSneyders:
We need to update all the new components in open PRs as well. I just merged #295, so I will still update that component as part of this PR.

RobbeSneyders:
The new lines missing coverage are all `pass` or `RaiseNotImplemented` lines, so would ignore the coverage report. The combined behavior of the `Component` and `Executor` is still the same and tested.

components/download_images/src/main.py
```diff
@@ -88,21 +89,19 @@ def download_image_with_retry(
...
-            retries: int,
-            image_size: int,
-            resize_mode: str,
-            resize_only_if_bigger: bool,
-            min_image_size: int,
-            max_aspect_ratio: float,
-    ) -> dd.DataFrame:
-        """Function that downloads images from a list of URLs and executes filtering and resizing
+    def __init__(self,
+                 *_,
```
PhilippeMoussalli, line 93: what's the difference between `*_` and `*`?

docs/custom_component.md
```diff
@@ -39,31 +39,35 @@ The easiest way to implement a `TransformComponent` is to subclass the provided
...
 
-    def transform(self, dataframe):
-        """Implement your custom transformation logic in this single method
-        
+    def transform(self, dataframe: pd.DataFrame) -> pd.DataFrame:
+        """Implement your custom logic in this single method
         Args:
-            dataframe: A Pandas dataframe containing one partition of your data
-            
+            dataframe: A Pandas dataframe containing the data
```
PhilippeMoussalli, line 59: why omit that it's one partition? might make things a bit more clear to the user if they want to go more in-depth

PhilippeMoussalli (approved):
Thanks Robbe! 

Left a few comments but overall looks really good

mrchtr:
Looks good to me, as my previous comments are addressed by the new issues.

components/download_images/src/main.py
RobbeSneyders: - `*` doesn't capture any arguments, it just indicates that all following arguments are keyword arguments
- `*args` captures all remaining positional arguments, but throws a linting error if you don't add it to the docstring args.
- `*_` is the same as `*args` but indicates that the captured arguments are not used. This is a [Python convention](https://stackoverflow.com/questions/5893163/what-is-the-purpose-of-the-single-underscore-variable-in-python) that linters recognize.

docs/custom_component.md
RobbeSneyders: This wasn't a conscious change, will revert.

PR #303, opened 2023-07-18
[DataComp] Add image downloading component

NielsRogge:
This PR adds an image download component to the DataComp pipeline.

It's required by the subsequent text spotting components.

Some things to include in the docs:

- had to install `docker buildx` to make pushing of components work (run `docker buildx install`)
- small fix in the `build_components.sh` script to fix [this](https://stackoverflow.com/questions/29081799/sed-1-invalid-command-code-f)
- running the pipeline locally worked when running `fondant run pipeline:pipeline --local`

### DIFF
--- a/components/download_images/Dockerfile
+++ b/components/download_images/Dockerfile
@@ -11,7 +11,7 @@ RUN pip3 install --no-cache-dir -r requirements.txt
 
 # Install Fondant
 # This is split from other requirements to leverage caching
-ARG FONDANT_VERSION=main
+ARG FONDANT_VERSION=6d2251ea24645a05721d428a01e4a9d8c963baba
 RUN pip3 install fondant[aws,azure,gcp]@git+https://github.com/ml6team/fondant@${FONDANT_VERSION}
 
 # Set the working directory to the component folder
--- a/components/download_images/fondant_component.yaml
+++ b/components/download_images/fondant_component.yaml
@@ -1,15 +1,15 @@
 name: Download images
 description: Component that downloads images based on URLs
-image: ghcr.io/ml6team/download_images:dev
+image: ghcr.io/ml6team/download_images:main
 
 consumes:
-  images:
+  image:
     fields:
       url:
         type: string
 
 produces:
-  images:
+  image:
     fields:
       data:
         type: binary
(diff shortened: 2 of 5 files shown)


components/download_images/Dockerfile
```diff
@@ -11,7 +11,7 @@ RUN pip3 install --no-cache-dir -r requirements.txt
 
 # Install Fondant
 # This is split from other requirements to leverage caching
-ARG FONDANT_VERSION=main
+ARG FONDANT_VERSION=6d2251ea24645a05721d428a01e4a9d8c963baba
```
RobbeSneyders, line 14: Should be reverted before merging.

components/download_images/fondant_component.yaml
```diff
@@ -3,13 +3,13 @@ description: Component that downloads images based on URLs
 image: ghcr.io/ml6team/download_images:dev
 
 consumes:
-  images:
+  image:
```
RobbeSneyders, line 6: I assume this will break the other pipelines using this component. So we should either update them all, or stick to `images`. 

I would propose to stick to `images` since plural is the standard used in our current example pipelines and components (`text` is both singular and plural), and I think it sounds most natural.

scripts/build_components.sh
```diff
@@ -75,7 +75,7 @@ for dir in "${components_to_build[@]}"; do
 
   echo "Updating the image version in the fondant_component.yaml with:"
   echo "${full_image_names[0]}"
-  sed -i "s|^image: .*|image: ${full_image_names[0]}|" fondant_component.yaml
+  sed -i '' "s|^image: .*|image: ${full_image_names[0]}|" fondant_component.yaml
```
RobbeSneyders, line 78: This doesn't seem to work on linux.

RobbeSneyders (commented):
Thanks @NielsRogge! Some comments.

scripts/build_components.sh
RobbeSneyders: The platform independent command seems to be `sed -i -e` ([link](https://unix.stackexchange.com/questions/13711/differences-between-sed-on-mac-osx-and-other-standard-sed))

NielsRogge: This seems to work, however it also creates a yaml-e file besides yaml. So will add yaml-e to gitignore

NielsRogge:
Closing this PR as I have a newer version of this.

PR #304, opened 2023-07-18
Bugfix default 0 values

PhilippeMoussalli:
Fixes default values set at `0` that were wrongly evaluated to `None`


NielsRogge (approved):
Thanks for fixing

PR #306, opened 2023-07-18
Add --build-arg argument to compile and run commands

### LINKED ISSUE #305: Add `--build-arg` argument to runners
Add a `--build-arg` argument to the runners so build arguments can be passed in for the build step. This is for example used to define the fondant version that should be installed in the component.

RobbeSneyders:
Fixes #305 

Can be used to test a pipeline with fondant installed from a specific git hash available on github:
```bash
fondant run pipeline:pipeline --local --build-arg FONDANT_VERSION=551f7b1
```


src/fondant/cli.py
```diff
@@ -228,6 +234,7 @@ def compile(args):
             pipeline=args.pipeline,
             extra_volumes=args.extra_volumes,
             output_path=args.output_path,
+            build_args=args.build_arg,
```
NielsRogge, line 237: Any reason we have build_args (plural) here?

components/prompt_based_laion_retrieval/src/main.py
```diff
@@ -35,7 +35,7 @@ def __init__(
         """
         self.client = ClipClient(
             url=url,
-            indice_name="laion5B",  #TODO:revert back to laion5b-L-14 after backend correction
+            indice_name="laion5B-L-14",
```
PhilippeMoussalli, line 38: fyi: this will only work for the public service atm. @ChristiaensBert did you do the backed modifications?

src/fondant/cli.py
RobbeSneyders: There can be multiple build arguments, but the command line is `--build-arg` since you provide them one at a time.

Eg. `--build-arg FONDANT_VERSION=a8b9fd --build-arg CUSTOM_ARG=value` becomes `build_args = ["FONANT_VERSION=a8b9fd", "CUSTOM_ARG=value"]`

components/prompt_based_laion_retrieval/src/main.py
RobbeSneyders: Yes, I reverted the arguments to the public service as well. Seems most logical to include in our example so users can easily test it on a small scale.

src/fondant/compiler.py
```diff
@@ -67,14 +69,25 @@ def compile(
...
         logger.info(f"Compiling {pipeline.name} to {output_path}")
-        spec = self._generate_spec(pipeline=pipeline, extra_volumes=extra_volumes)
+        spec = self._generate_spec(
+            pipeline,
+            extra_volumes=extra_volumes,
+            build_args=build_args or [],
+        )
+
+        class NoAliasDumper(yaml.SafeDumper):
+            def ignore_aliases(self, data):
```
GeorgesLorre, line 85: Could you briefly explain why we need this ?

GeorgesLorre (approved):

PR #307, opened 2023-07-19
Bugfix: Data explorer local runner usage

mrchtr:
I have tested `fondant explore` by using local manifest on a M1. Therefore, I ran into some issues.

- I updated the data explorer Dockerfile according to this solution proposal (https://github.com/streamlit/streamlit/issues/4842#issuecomment-1309856767) which allows the execution of the streamlit container on a M1

- If a user has not named the local data directory as `artifacts`, the local path can not be resolved, and the data explorer can not load the parquet files. Therefore, I propose keeping the original name of the data directory when mounting it to the Docker container.

- Some smaller typo fixes of the data explorer documentation


GeorgesLorre (approved):
Thx!

RobbeSneyders (approved):

RobbeSneyders:
Thanks @mrchtr!

PR #308, opened 2023-07-19
Update script to work with macos

GeorgesLorre:


GeorgesLorre:
Tried it on MacOs and it seems to work fine now but need to make sure it is not broken for other systems now.

PR #309, opened 2023-07-19
Introduce repartitioning

PhilippeMoussalli:
PR that introduces the partitioning strategy discussed in #288 

1) The automatic behavior is as follows for all component types (dask, pandas)
* The written dataframe is re-partitioned to 250 Mb
* The loaded dataframe is re-partitioned depending on the current number of partitions and workers 

2) The behavior above can be overwritten by the end user in case they want to implement their own custom logic, this is done on the ComponentOp level as an additional flag parameters that can be passed. See added docs with this PR for more details 

I will handle adding the diagnostic tools and optimizing the downloader component in a separate PR.

### DIFF
--- a/docs/pipeline.md
+++ b/docs/pipeline.md
@@ -8,7 +8,7 @@ To build a pipeline, you need to define a set of component operations called `Co
 
 The component specifications include the location of the Docker image in a registry.
 
-The runtime configuration consists of the component's arguments and the definition of node pools and resources. For example, if a component requires GPU for model inference, you can specify the necessary GPU resources in the runtime configuration.
+The runtime configuration consists of the component's arguments and the definition of node pools, resources and custom partitioning specification. For example, if a component requires GPU for model inference, you can specify the necessary GPU resources in the runtime configuration.
 
 Here is an example of how to build a pipeline using Fondant:
 ```python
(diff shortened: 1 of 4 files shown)


src/fondant/data_io.py
```diff
@@ -17,6 +18,31 @@ def __init__(self, *, manifest: Manifest, component_spec: ComponentSpec) -> None
...
+            The partitioned dataframe.
+        """
+        n_partitions = dataframe.npartitions
+        n_workers = os.cpu_count()
+        logger.info(
+            f"The number of partitions of the input dataframe is {n_partitions}. The "
+            f"available number of workers is {n_workers}.",
+        )
+        if n_partitions < n_workers:
+            dataframe = dataframe.repartition(npartitions=n_partitions)
```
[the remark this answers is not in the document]
PhilippeMoussalli: woops it should be  `dataframe = dataframe.repartition(npartitions=n_workers)`
good catch

src/fondant/executor.py
```diff
@@ -39,13 +39,15 @@ def __init__(
         input_manifest_path: t.Union[str, Path],
         output_manifest_path: t.Union[str, Path],
         metadata: t.Dict[str, t.Any],
-        user_arguments: t.Dict[str, Argument],
+        user_arguments: t.Dict[str, t.Any],
+        output_partition_size: t.Optional[str] = "250MB",
```
RobbeSneyders, line 43: Shouldn't this be added to the parser as well, so it's extracted by `from_args`?

src/fondant/pipeline.py
```diff
@@ -75,6 +79,38 @@ def __init__(
...
 
+    def _set_arguments(
+        self,
+        arguments: t.Optional[t.Dict[str, t.Any]],
+    ) -> t.Dict[str, t.Any]:
+        """Set component arguments based on provided arguments and relevant ComponentOp
+        parameters.
+        """
+
+        def _validate_partition_size_arg(file_size):
```
RobbeSneyders, line 90: Can we add this as a `type` when registering it on the parser in the `Executor` class?

RobbeSneyders (commented):
Thanks @PhilippeMoussalli! Some comments.

docs/pipeline.md
```diff
@@ -46,7 +46,72 @@ Next, we define two operations: `load_from_hub_op`, which is a based from a reus
...
+
+caption_images_op = ComponentOp(  
+    component_dir="components/captioning_component",  
+    arguments={  
+        "model_id": "Salesforce/blip-image-captioning-base",  
+        "batch_size": 2,  
+        "max_new_tokens": 50,  
+    },  
+    input_partition_rows=100, 
+    output_partition_size="10MB",
```
NielsRogge, line 104: Any reason the input partitions are specified in terms of rows and the output partitions in terms of size?

satishjasthi: Ya I do feel using num rows is way more intuitive than size

GeorgesLorre (approved):
Nice work! Looking forward to see this in action

satishjasthi:
> Nice work! Looking forward to see this in action

@GeorgesLorre, I appreciate your patience and quick responses in resolution of intermediate issue during this PR review.😌

Issue #254, opened 2023-06-29
Improve data explorer UX


RobbeSneyders:
Currently the user needs to provide the path to a manifest file in the data explorer, but this path is based on the mounted data directory, which makes this not straightforward.

It would be ideal if we can detect the different manifest files and provide a dropdown or folder structure instead which the user can use to select the manifest.

Ideally, we can do this for remote paths as well by letting the user eg. provide a remote base path which we scan for manifests.

RobbeSneyders:
The data explorer throws an error when loading a subset without numeric columns. The `numeric_field` on the screenshot below is `None` in this case.

![Image](https://github.com/ml6team/fondant/assets/20990866/e430fdd4-5fd5-49aa-9afe-2fd71d09a8fd)

Traceback from terminal:

> Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/streamlit/runtime/scriptrunner/script_runner.py", line 552, in _run_script
    exec(code, module.__dict__)
  File "/app/main.py", line 48, in <module>
    build_numeric_analysis_plots(dataframe, numeric_fields)
  File "/app/widgets.py", line 180, in build_numeric_analysis_plots
    make_numeric_plot(dataframe, numeric_field, plot_type)
  File "/app/numeric_analysis.py", line 78, in make_numeric_plot
    data = dataframe[numeric_field].value_counts()
  File "/usr/local/lib/python3.8/site-packages/dask/dataframe/core.py", line 4851, in __getitem__
    raise NotImplementedError(key)
NotImplementedError: None

PhilippeMoussalli:
Additional improvements: 

* Include more schema for text data
* Find a better way to infer images columns

PhilippeMoussalli:
- [ ] Replace

PR #310, opened 2023-07-20
Bugfix: data explorer artifact mounting

mrchtr:
During the last fix I have incorporate a new bug in mounting the correct path to the docker container.
I have adapted the implementation.  The path in `fondant explore --data-directory <path>` can also be an relative path now. 
The path will be resolved to an absolute path. Inside the docker container the folder will be located in the root folder.


RobbeSneyders (approved):
Thanks! This works a lot nicer :)

Ran into another issue in the data explorer, but added it to https://github.com/ml6team/fondant/issues/254

PR #311, opened 2023-07-25
Add test cases for caption_images component and fixed bug in this com…

### LINKED ISSUE #301: Add component tests

satishjasthi:
This the first commit to add test cases to all existing components of fondant. This PR tries to resolve #301


mrchtr:
Thanks @satishjasthi! The test case is already a good starting point to build a unit test fundament!

During a quick offline discussion with @GeorgesLorre about the general test layout we came up with an idea. 
It would be great if we would have something like an `AbstractComponentTest` which can be used for all components. 

The basic idea is to make the implementation of unit tests as easy as possible for the users. In the best case scenario, users would only need to define a dataframe containing their input dummy data, the expected output data, and the component arguments. A default class could be created to handle tasks such as invoking the component's transform method and comparing the expected output dataframe with the actual result.

Would you be up for drafting such a class and use it to test the `caption_images` component?

satishjasthi:
Hey @mrchtr, Ya sure that sounds like a better idea, I would love to create such an abstract class for caption_images component.

satishjasthi:
> Thanks @satishjasthi! The test case is already a good starting point to build a unit test fundament!
> 
> During a quick offline discussion with @GeorgesLorre about the general test layout we came up with an idea. It would be great if we would have something like an `AbstractComponentTest` which can be used for all components.
> 
> The basic idea is to make the implementation of unit tests as easy as possible for the users. In the best case scenario, users would only need to define a dataframe containing their input dummy data, the expected output data, and the component arguments. A default class could be created to handle tasks such as invoking the component's transform method and comparing the expected output dataframe with the actual result.
> 
> Would you be up for drafting such a class and use it to test the `caption_images` component?

Hey @mrchtr, I have pushed commit for AbstractComponentTest class. My idea was to create a pytest class which can validate 
- Whether a new component created has base component as one of the components defined in fondant.component.py
- And to validate whether this new component has all the methods and its respective attributes that are defined in its base class

This can be used to test all the components, expect the actually functionality of the components which varies from one another, so user can add test cases as methods to this class to incorporate validations for those custom component behaviours.
Please let me know your views on this. Thank you

satishjasthi:
Hi @mrchtr, Thanks for the detailed explanation, I like this idea as it can easily extend to all components easily. I'll create a base class in this manner and submit new commit.

satishjasthi:
Hey @mrchtr, I have update the test case for CaptionImage with a new base class and used it for testing this component. Please review it and let me know your views.

components/caption_images/tests/test_caption_images.py
```diff
@@ -0,0 +1,91 @@
...
+            data={("captions", "text"): {0: "a motorcycle", 1: "a beetle"}},
+        )
+
+    @pytest.fixture(autouse=True)
+    def __setUp(self):
+        self.component = self.create_component()
+        self.input_data = self.create_input_data()
+        self.expected_output_data = self.create_output_data()
+
+    def test_transform(self):
```
mrchtr, line 84: Can we move this implementation to the super class, since it should be the same for most components?

components/caption_images/tests/test_caption_images.py
```diff
@@ -0,0 +1,91 @@
...
+        output = self.component.transform(self.input_data)
+        assert output.equals(self.expected_output_data)
+
+    def tearDown(self):
```
mrchtr, line 88: I guess we can remove this method. I have added the `tearDown` just as en example and I think we don't need it here explicitly.

components/caption_images/tests/test_caption_images.py
```diff
@@ -0,0 +1,91 @@
...
+            {"images": {"data": [requests.get(url).content for url in image_urls]}},
+        )
+
+    def create_output_data(self):
+        return pd.DataFrame(
+            data={("captions", "text"): {0: "a motorcycle", 1: "a beetle"}},
+        )
+
+    @pytest.fixture(autouse=True)
+    def __setUp(self):
```
mrchtr, line 79: Is it possible to remove this method as well? Ideally, the `setUp` of the super class is called.

mrchtr (commented):
Thanks a lot again @satishjasthi! I left some smaller comments. Got the feeling we are on the right track.

components/caption_images/tests/test_caption_images.py
satishjasthi: Yes we can, as this method has pretty much stand method of testing transform method for all classes.

components/caption_images/tests/test_caption_images.py
satishjasthi: Indeed, I felt the same but I retained it as its standard template in pytest testing with setup and teardown methods. I'll remove that.

components/caption_images/tests/test_caption_images.py
satishjasthi: Yes we can, as the code within this method is again remains same across components

satishjasthi:
@mrchtr, I have updated and pushed new commit with suggested changes, please review it.

src/fondant/abstract_component_test.py
```diff
@@ -0,0 +1,50 @@
...
+        self.input_data = self.create_input_data()
+        self.expected_output_data = self.create_output_data()
+
+    def test_transform(self):
+        """
+        Default test for the transform method.
+        Tests if the transform method executes without errors.
+        """
+        output = self.component.transform(self.input_data)
+        if not output.equals(self.expected_output_data):
```
mrchtr, line 46: Nitpick - pandas offers you already an assertion function: 
`pd.testing.assert_frame_equal(output,self.expected_output_data)`

mrchtr (approved):
Looks good to me. I have tested the code locally. 
Added on super small nitpick. 

I think we can merge it, and see it as a first blueprint for further component test. Thanks again!

src/fondant/abstract_component_test.py
satishjasthi: @mrchtr , I have pushed code with this change. Shall I create test cases for remaining components in the same PR or a new one?

Issue #313, opened 2023-07-25
Enable partial pipeline execution for local runner


PhilippeMoussalli:
Issue discussed in more details [here](https://github.com/ml6team/fondant/pull/292)

```[tasklist]
### Tasks
- [ ] https://github.com/ml6team/fondant/issues/314
- [ ] https://github.com/ml6team/fondant/issues/315
- [ ] https://github.com/ml6team/fondant/issues/321
- [ ] https://github.com/ml6team/fondant/issues/317
```

PR #318, opened 2023-07-26
Add method to estimate caching key

PhilippeMoussalli:
Related to #313 #292 

The cache key is a unique identifier for the component that will be used to decide whether a component should be executed or not. 

This is now estimated at compile time since it's easier to estimate the docker image digest outside of the component and pipelines (requires `docker` to be installed and authenticated). Only caveat there is that the image digest might change between the time the pipeline is compiled and the time it's run (e.g. image changes to a component is made after starting the run under a similar tag). However, I don't a straightforward  way to extract the image digest from within the component itself:

*  Connecting to the docker API with `curl` -> requires passing an authentication key 
*  Passing the digest as part of an environmental variable during the build process -> requires modification to the Docker image


GeorgesLorre:
In the local runner we often build custom components on the fly. How will we handle this ?

PhilippeMoussalli:
> In the local runner we often build custom components on the fly. How will we handle this ?

Missed that one, it's a bit tricky because of two reasons: 
* The digest is actually different from registry to registry, images built locally and remotely will have different digests https://blog.aquasec.com/docker-image-tags. This means that if you want to resume a pipeline that was run remotely on the local runner (e.g. by specifying the base path to be remote), caching will not work. 
* Even if we were to have the same digest, we're estimating the digest before running which is when the build happens with `docker-compose up`

Given those constrains and the fact that it's probably more efficient to estimate the digest during runtime based on the reasons mentioned in the description, I think an ideal solution would be to have a unique digest that can be retrieved during runtime. Either by 

* Including it as an env variable during build time (downside is that we have to change the structure of the Dockerfile to add an environment) 

*  Estimating it internally within the component (might require installing additional packages)

There is still the issue of having a unique digest, do you have any ideas on how to best approach this?

src/fondant/pipeline.py
```diff
@@ -153,6 +155,132 @@ def from_registry(
...
+    def get_image_manifest(image_ref: str):
+        """Retrieve the Docker image manifest.
+
+        Args:
+            image_ref: The Docker image reference (e.g., 'registry/image:tag').
+
+        Returns:
+            dict: The parsed JSON manifest.
+        """
+        cmd = ["docker", "manifest", "inspect", image_ref]
```
GeorgesLorre, line 168: This will make docker a hard requirement up until now it was more of a soft requirement for local running. If we do this we might want to use the docker python sdk

src/fondant/pipeline.py
```diff
@@ -153,6 +155,132 @@ def from_registry(
...
+        different operating systems, the hash of the amd64 CPU architecture and the linux
+        operating system is returned.
+
+        Args:
+            image_ref (str): The Docker image reference (e.g., 'registry/image:tag').
+
+        Returns:
+            str: The hash value (digest) of the Docker image.
+        """
+        manifest = self.get_image_manifest(image_ref)
```
GeorgesLorre, line 195: I guess getting the digest of local images is easy, it is the remote part that makes this difficult

src/fondant/pipeline.py
GeorgesLorre, line 168: I guess you need access to the registry to run this code ?

src/fondant/pipeline.py
```diff
@@ -153,6 +155,132 @@ def from_registry(
...
+
+            Args:
+            input_dict: The dictionary to be converted.
+
+            Returns:
+            A sorted JSON string representing the dictionary.
+            """
+            if isinstance(input_dict, dict):
+                return json.dumps(
+                    {k: sorted_dict_to_json(v) for k, v in sorted(input_dict.items())},
```
GeorgesLorre, line 244: Some recursion, cool!

GeorgesLorre:
I think overall the code looks fine. I'm just not a big fan of how we have to compare the docker images but I have no better proposition.

src/fondant/pipeline.py
PhilippeMoussalli: yeah indeed, I'm also not really happy with this but I don't think it's possible [link]. (https://stackoverflow.com/questions/56178911/how-to-obtain-docker-image-digest-from-tag).
For now the SDK seems to only support inspecting local images [link](https://docker-py.readthedocs.io/en/stable/api.html#:~:text=inspect_container(container)%C2%B6). The `docker manifest inspect` seems to a [new command](https://docs.docker.com/engine/reference/commandline/manifest_inspect/) so it might not have been integrated yet with the SDK. 

Perhaps we can use `curl` but then it requires to know the url of different registries [link](https://cloud.google.com/kubernetes-engine/docs/archive/using-container-images)

src/fondant/pipeline.py
PhilippeMoussalli: yeah It should be easy but we still have to somehow build the image -> get the digest -> compile
or we do it at runtime in which case we have to figure out how to pass the digest to the image itself. I checked a bit and it's not available in the image itself unless passed explicitly as an environmental variable. It's a bit of a chicken and an egg problem.

src/fondant/pipeline.py
PhilippeMoussalli: I think it should be fine for public images

RobbeSneyders:
Thanks @PhilippeMoussalli!

I see a couple of issues with this:
- This adds docker as a local dependency
- This requires local authentication to the registry, which might not be available
- The image digest might change after compilation

I'm wondering if we shouldn't simplify this further and calculate the hash key based on the image name and tag. 

This has the downside that the underlying image might change, but that shouldn't really happen anyway. And as long as we provide a flag to disable the caching, the user can always work around it. We could even automatically disable it for tags that are known to change like 'latest'. 

This downside is also already partially present in the current approach (see `the image digest might change after compilation` above).

But instead, we lose all the other disadvantages.

RobbeSneyders (commented):
Thanks @PhilippeMoussalli!

Looks good to me, 2 minor comments.

src/fondant/pipeline.py
```diff
@@ -139,6 +141,65 @@ def from_registry(
...
+            numbers, booleans, or None).
+
+            Args:
+            input_dict: The dictionary to be converted.
+
+            Returns:
+            A sorted JSON string representing the dictionary.
+            """
+            if isinstance(input_dict, dict):
+                return json.dumps(
```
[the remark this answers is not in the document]
PhilippeMoussalli: oh nice! the more you know :D

src/fondant/pipeline.py
```diff
@@ -139,6 +141,63 @@ def from_registry(
...
+            Args:
+                input_dict: The dictionary to be converted.
+
+            Returns:
+                A sorted JSON string representing the dictionary.
+            """
+            if isinstance(input_dict, dict):
+                return json.dumps(input_dict, sort_keys=True)
+
+            return input_dict
```
RobbeSneyders, line 171: I think we can now remove this whole function and replace it with the `json.dumps` call.

src/fondant/pipeline.py
```diff
@@ -139,6 +141,63 @@ def from_registry(
...
+        def get_nested_dict_hash(input_dict):
+            """Calculate the hash of a nested dictionary.
+
+            Args:
+                input_dict: The nested dictionary to calculate the hash for.
+
+            Returns:
+                The hash value (MD5 digest) of the nested dictionary.
+            """
+            sorted_json_string = sorted_dict_to_json(input_dict)
```
RobbeSneyders, line 182:
```suggestion
            sorted_json_string = json.dumps(input_dict, sort_keys=True)
```

RobbeSneyders (approved):

PR #319, opened 2023-07-26
Run Datacomp at scale

NielsRogge:
This PR includes updates made to the Fondant code base to make it run at scale for Datacomp

### DIFF
--- a/components/filter_image_resolution/Dockerfile
+++ b/components/filter_image_resolution/Dockerfile
@@ -12,7 +12,7 @@ RUN pip3 install --no-cache-dir -r requirements.txt
 
 # Install Fondant
 # This is split from other requirements to leverage caching
-ARG FONDANT_VERSION=28ec87862c160ead773eb15b57905ac61515f8cf
+ARG FONDANT_VERSION=f3f3925b8e8f634e2978e5c7fcefa72c53baba7c
 RUN pip3 install fondant[aws,azure,gcp]@git+https://github.com/ml6team/fondant@${FONDANT_VERSION}
 
 # Set the working directory to the component folder
--- a/components/filter_image_resolution/Dockerfile
+++ b/components/filter_image_resolution/Dockerfile
@@ -11,7 +11,7 @@ RUN pip3 install --no-cache-dir -r requirements.txt
 
 # Install Fondant
 # This is split from other requirements to leverage caching
-ARG FONDANT_VERSION=79df895e9d62d2010ccb8d40ee7e4fd4c68f117d
+ARG FONDANT_VERSION=41bd8f9d8f8003b41ffa375c1887869208e519ea
 RUN pip3 install fondant[aws,azure,gcp]@git+https://github.com/ml6team/fondant@${FONDANT_VERSION}
 
 # Set the working directory to the component folder
--- a/components/filter_image_resolution/Dockerfile
+++ b/components/filter_image_resolution/Dockerfile
@@ -11,7 +11,7 @@ RUN pip3 install --no-cache-dir -r requirements.txt
 
 # Install Fondant
 # This is split from other requirements to leverage caching
-ARG FONDANT_VERSION=main
+ARG FONDANT_VERSION=79df895e9d62d2010ccb8d40ee7e4fd4c68f117d
 RUN pip3 install fondant[aws,azure,gcp]@git+https://github.com/ml6team/fondant@${FONDANT_VERSION}
 
 # Set the working directory to the component folder
--- a/components/filter_image_resolution/Dockerfile
+++ b/components/filter_image_resolution/Dockerfile
@@ -12,7 +12,7 @@ RUN pip3 install --no-cache-dir -r requirements.txt
 
 # Install Fondant
 # This is split from other requirements to leverage caching
-ARG FONDANT_VERSION=1aaf6068e7135b917e5e5cb16b3a2bf9d58f4a4e
+ARG FONDANT_VERSION=f18dd510b1f106f421d433eb02b512177d5116a3
 RUN pip3 install fondant[aws,azure,gcp]@git+https://github.com/ml6team/fondant@${FONDANT_VERSION}
 
 # Set the working directory to the component folder
--- a/components/filter_image_resolution/Dockerfile
+++ b/components/filter_image_resolution/Dockerfile
@@ -3,7 +3,8 @@ FROM --platform=linux/amd64 python:3.8-slim
 # System dependencies
 RUN apt-get update && \
     apt-get upgrade -y && \
-    apt-get install git -y
+    apt-get install git -y && \
+    apt-get install graphviz -y
 
 # Install requirements
 COPY requirements.txt /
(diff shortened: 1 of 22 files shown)


components/load_from_hf_hub/src/main.py
```diff
@@ -44,12 +48,27 @@ def load(self) -> dd.DataFrame:
...
+            npartitions = self.n_rows_to_load // partition_length
+            dask_df = dask_df.head(self.n_rows_to_load, npartitions=npartitions)
+            dask_df = dd.from_pandas(dask_df, npartitions=npartitions)
+            # .reset_index(drop=True) # will reset it from 0 for every partition
+
+        # Set monotonically increasing index
+        logger.info("Setting the index...")
+        dask_df["id"] = 1
+        dask_df["id"] = dask_df.id.cumsum()
+        dask_df = dask_df.set_index("id", sort=True)
```
NielsRogge, line 71: This is based on https://stackoverflow.com/questions/47571715/dask-create-strictly-increasing-index

components/filter_image_resolution/fondant_component.yaml
```diff
@@ -1,14 +1,14 @@
...
 description: Component that filters images based on minimum size and max aspect ratio
-image: ghcr.io/ml6team/filter_image_resolution:latest
+image: ghcr.io/ml6team/filter_image_resolution:f3f3925b8e8f634e2978e5c7fcefa72c53baba7c
 
 consumes:
   image:
     fields:
       width:
-        type: int16
+        type: int64
```
PhilippeMoussalli, line 9: int16 maxes out at 32.767 which is why you may be having issues if your image's width was 36.000 
However, I think this should be resolved if you move to `int32`

components/filter_image_resolution/Dockerfile
```diff
@@ -3,15 +3,16 @@ FROM --platform=linux/amd64 python:3.8-slim
...
+    apt-get install graphviz -y
 
 # Install requirements
 COPY requirements.txt /
 RUN pip3 install --no-cache-dir -r requirements.txt
 
 # Install Fondant
 # This is split from other requirements to leverage caching
-ARG FONDANT_VERSION=main
+ARG FONDANT_VERSION=f3f3925b8e8f634e2978e5c7fcefa72c53baba7c
```
PhilippeMoussalli, line 15: @GeorgesLorre do we still need to revert the docker image to `main` or do they get automatically tagged as `dev` during merge?

components/load_from_hf_hub/fondant_component.yaml
```diff
@@ -23,4 +23,9 @@ args:
   n_rows_to_load:
     description: Optional argument that defines the number of rows to load. Useful for testing pipeline runs on a small scale
     type: int
-    default: None
\ No newline at end of file
+    default: None
+  dataset_length:
```
PhilippeMoussalli, line 27: I would avoid having the user specify the length of the dataset and read it from the hf metadata directly

examples/pipelines/datacomp/components/download_images/src/main.py
```diff
@@ -0,0 +1,159 @@
+"""
```
PhilippeMoussalli, line 1: What are the reasons for duplicating this component and not making the changes directly to the one in the registry?

components/filter_image_resolution/Dockerfile
```diff
@@ -3,15 +3,16 @@ FROM --platform=linux/amd64 python:3.8-slim
 # System dependencies
 RUN apt-get update && \
     apt-get upgrade -y && \
-    apt-get install git -y
+    apt-get install git -y && \
+    apt-get install graphviz -y
```
PhilippeMoussalli, line 7: You can omit `graphviz` for now, we'll tackle this in #331

pyproject.toml
```diff
@@ -46,6 +46,7 @@ dask = {extras = ["dataframe"], version = ">= 2023.4.1"}
 importlib-resources = { version = ">= 1.3", python = "<3.9" }
 jsonschema = ">= 4.18"
 pyarrow = ">= 11.0.0"
+graphviz = ">= 0.20.1"
```
PhilippeMoussalli, line 49: same comment here as above, maybe for testing you can just have it in the `requirements.txt` for now

scripts/build_components.sh
```diff
@@ -75,7 +75,7 @@ for dir in "${components_to_build[@]}"; do
 
   echo "Updating the image version in the fondant_component.yaml with:"
   echo "${full_image_names[0]}"
-  sed -i "s|^image: .*|image: ${full_image_names[0]}|" fondant_component.yaml
+  sed -i '' "s|^image: .*|image: ${full_image_names[0]}|" fondant_component.yaml
```
PhilippeMoussalli, line 78: what is the extra `''` for?

src/fondant/data_io.py
```diff
@@ -92,6 +93,9 @@ def _load_subset(self, subset_name: str, fields: t.List[str]) -> dd.DataFrame:
 
         subset_df = dd.read_parquet(remote_path, columns=fields)
 
+        logger.info(f"First few rows of subset {subset_name}:")
```
PhilippeMoussalli, line 96: I guess this is only for debugging and can be omitted

src/fondant/data_io.py
```diff
@@ -124,11 +128,16 @@ def load_dataframe(self) -> dd.DataFrame:
...
+        logging.info(f"Loading index...")
         dataframe = self._load_index()
+
         for name, subset in self.component_spec.consumes.items():
             fields = list(subset.fields.keys())
             subset_df = self._load_subset(name, fields)
             # left joins -> filter on index
+            # make sure that dataframe has same number of partitions
+            # as subset
+            dataframe = dataframe.repartition(npartitions=subset_df.npartitions)
```
PhilippeMoussalli, line 140: I'll have a look at this, eventually we can maybe even flag it as an issue to the `Dask` repo

src/fondant/data_io.py
```diff
@@ -196,6 +206,12 @@ def write_dataframe(self, dataframe: dd.DataFrame) -> None:
 
         dataframe.index = dataframe.index.rename("id").astype("string")
 
+        # logging.info("Visualizing task graph...")
```
PhilippeMoussalli, line 209: If I have custom code that I use for testing but don't want to commit I usually use 
```
git add -p <filename>
```
which allows you to only commit chunks of the file, can be also done in the IDE. 

Then if I want to switch to another branch and not lose the testing code, I stash the changes under a descriptive name

```
git stash save datacomp-scale-test
```

You can then fetch it back using 
```
git stash apply <stash_id>
```

src/fondant/schemas/common.json
```diff
@@ -7,6 +7,7 @@
         "int8",
         "int16",
         "int32",
+        "int64",
```
PhilippeMoussalli, line 10: This can be kept, I would just check that previous component can run at `int32` to reduce memory footprint

PhilippeMoussalli (commented):
Thanks Niels! 
I'll try to reproduce the steps and investigate the merging issue

scripts/build_components.sh
NielsRogge: This was required to run the script on MacOS: https://stackoverflow.com/questions/29081799/sed-1-invalid-command-code-f

NielsRogge:
Closing this PR in favor of a more up-to-date branch.

PR #320, opened 2023-07-26
Add disable caching argument

PhilippeMoussalli:
Related to #313 

Update [22/08]

This PR has been updated to add an argument at the componentOp level to disable caching. Previous implementation relied on passing an `execute_component` since we were estimating  whether a component is cached or not at compile time.

A component will only execute if `disable_caching` is not enabled and if previous matching executions are found (this will be implemented in a later PR)


src/fondant/compiler.py
```diff
@@ -143,17 +143,16 @@ def _generate_spec(
...
-                ],
-            )
+            execute_component = pipeline.execute_component(component_op)
+
+            command = [
+                "--metadata",
+                json.dumps(asdict(metadata)),
+                "--output_manifest_path",
+                f"{path}/{safe_component_name}/manifest.json",
+                "--execute_component",
```
GeorgesLorre, line 153: do we need this is the local compiler/runner ?

src/fondant/executor.py
```diff
@@ -195,12 +207,18 @@ def execute(self, component_cls: t.Type[Component]) -> None:
...
+            logging.info("Executing component")
+            component = component_cls(self.spec, **self.user_arguments)
+            output_df = self._execute_component(component, manifest=input_manifest)
+        else:
+            logging.info("Cached component run. Skipping component execution")
 
         output_manifest = input_manifest.evolve(component_spec=self.spec)
 
-        self._write_data(dataframe=output_df, manifest=output_manifest)
+        if self.execute_component:
```
GeorgesLorre, line 220: So a skipped component still technically runs it just uploads a manifest.

src/fondant/compiler.py
PhilippeMoussalli: It's gonna default for False now since caching is going to be disabled by default, but kept it for now for when we implement it in the future

src/fondant/executor.py
PhilippeMoussalli: yeah indeed, the `execution` here refers to loading/transforming the data. Perhaps we need a better name

src/fondant/executor.py
```diff
@@ -177,20 +189,43 @@ def _write_data(self, dataframe: dd.DataFrame, *, manifest: Manifest):
 
         data_writer.write_dataframe(dataframe)
 
+    def _load_cached_output_manifest(self) -> "Manifest":
```
PhilippeMoussalli, line 192: Those two methods will be implemented in a separate PR

src/fondant/compiler.py
```diff
@@ -147,16 +147,12 @@ def _generate_spec(
...
-                    "--output_manifest_path",
-                    f"{path}/{safe_component_name}/manifest.json",
-                ],
-            )
+            command = [
+                "--metadata",
+                json.dumps(asdict(metadata)),
+                "--output_manifest_path",
+                f"{path}/{safe_component_name}/manifest.json",
+            ]
```
RobbeSneyders, line 155: This will conflict with some of the changes in your other PRs.

RobbeSneyders (commented):
Thanks @PhilippeMoussalli!

Would just propose to use `cache` with default value of `True` instead of `disable_caching`. I think it's a bit more clear than using a negative parameter.

src/fondant/compiler.py
PhilippeMoussalli: True, will revert it back

src/fondant/component_spec.py
```diff
@@ -263,6 +263,12 @@ def from_fondant_component_spec(
                     "type": "String",
                     "default": "None",
                 },
+                {
+                    "name": "cache",
+                    "description": "Set to False to disable caching, True by default.",
+                    "type": "Boolean",
+                    "default": "True",
```
RobbeSneyders, line 270: Can we not pass a bool here instead of a string?

RobbeSneyders: https://github.com/kubeflow/pipelines/issues/5111

PhilippeMoussalli: All default arguments are currently defined as strings even `None`. We also have do do this in the component spec. Otherwise some weird behavior will occur. In this example, setting "`True"` or `True` will still result in True but setting "`False`" and `False` will result in False and True respectively. Something about the way kubeflow evaluates arguments.

RobbeSneyders (approved):

RobbeSneyders:
Can you resolve the conflicts and merge @PhilippeMoussalli?

PR #327, opened 2023-07-31
Add node pool label

shayorshay:
This PR adds a configurable node pool label in the pipeline specification. This is needed because the default label name differs per cloud provider.

### DIFF
--- a/src/fondant/pipeline.py
+++ b/src/fondant/pipeline.py
@@ -48,7 +48,7 @@ class ComponentOp:
 
     Note:
         - A Fondant Component operation is created by defining a Fondant Component and its input
-            arguments.
+          arguments.
         - The `number_of_gpus`, `node_pool_label`, `node_pool_name`,`p_volumes` and
           `ephemeral_storage_size` attributes are optional and can be used to specify additional
           configurations for the operation. More information on the optional attributes that can
--- a/src/fondant/pipeline.py
+++ b/src/fondant/pipeline.py
@@ -82,8 +82,10 @@ class ComponentOp:
         self.arguments.setdefault("component_spec", self.component_spec.specification)
 
         self.number_of_gpus = number_of_gpus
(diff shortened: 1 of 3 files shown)


src/fondant/pipeline.py
```diff
@@ -80,6 +82,7 @@ def __init__(
         self.arguments.setdefault("component_spec", self.component_spec.specification)
 
         self.number_of_gpus = number_of_gpus
+        self.node_pool_label = node_pool_label
         self.node_pool_name = node_pool_name
         self.p_volumes = p_volumes
         self.ephemeral_storage_size = ephemeral_storage_size
```
PhilippeMoussalli, line 88: Since both the label and the name depend on each other. Can you raise an error if one of them is not provided? 

```python
from fondant.exceptions import InvalidPipelineDefinition 
if bool(node_pool_label) != bool(node_pool_name):
        raise InvalidPipelineDefinition("Both 'node_pool_label' and 'node_pool_name' must be defined or both must be None.")
```

And can you also test that a valid error is returned by adding two example to `test_component_op`, one where the name is defined and the nodepool is not and the other way around: 

https://github.com/ml6team/fondant/blob/5b123bff57c933e3fbc24aac7a36f73bf4ddacd3/tests/test_pipeline.py#L35

PhilippeMoussalli (commented):
Thanks @shayorshay ! :) 
This will come in handy for different clouds. Left one small comments to handle edge cases and maybe add a test for it.

src/fondant/pipeline.py
shayorshay: I added the error checking and test. Could you please check? @PhilippeMoussalli

PhilippeMoussalli: Thanks :)

PhilippeMoussalli (approved):

PR #332, opened 2023-08-02
[Scripts] Fix build_components

NielsRogge:
This PR ensures that the `build_components.sh` script can run on both Linux and MacOS.

As suggested by @RobbeSneyders [here](https://github.com/ml6team/fondant/pull/303#discussion_r1269909421), thanks!


PhilippeMoussalli (approved):
Thanks!

PR #333, opened 2023-08-02
[Commoncrawl pipeline] Add webpage url filter component

shayorshay:
This PR adds a URL filter component to filter out webpages from a given category/categories of blacklisted urls. The database of blacklists is from https://dsi.ut-capitole.fr/blacklists.

### DIFF
--- a/examples/pipelines/commoncrawl/components/filter_webpage_urls/fondant_component.yaml
+++ b/examples/pipelines/commoncrawl/components/filter_webpage_urls/fondant_component.yaml
@@ -1,0 +1,25 @@
+name: Filter webpage urls
+description: Component that filters webpage urls based on given blacklist categories
+# image: ghcr.io/ml6team/filter_webpage_urls:latest
+image: 050243278346.dkr.ecr.us-east-1.amazonaws.com/filter_webpage_urls:latest
+
+consumes:
+  webpage:
+    fields:
+      url:
+        type: string
+      html:
+        type: string
+
+produces:
+  webpage:  
+    fields:
+      url:
+        type: string
+      html:
+        type: string
+
+args:
+  categories:
+    description: Categories of blacklisted webpages to filter out
+    type: list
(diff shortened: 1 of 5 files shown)


examples/pipelines/commoncrawl/components/filter_webpage_urls/src/main.py
```diff
@@ -0,0 +1,54 @@
...
+    result = []
+    for category in categories:
+        path = os.path.join(blacklist_dir, "blacklists", category, "domains")
+        urls = get_urls_from_file(path)
+        result.extend(urls)
+
+    return result
+
+
+class FilterWebpageUrls(DaskTransformComponent):
```
mrchtr, line 29: I think we could implement this as a `PandasTransformComponent`.

examples/pipelines/commoncrawl/components/filter_webpage_urls/src/main.py
```diff
@@ -0,0 +1,54 @@
...
+from utils.file_utils import download_tar_file, get_urls_from_file
+
+logger = logging.getLogger(__name__)
+
+BLACKLIST_DIR = "/tmp/blacklists"
+BLACKLIST_URL = "http://dsi.ut-capitole.fr/blacklists/download/blacklists.tar.gz"
+
+
+def get_blacklisted_domains(blacklist_dir: str, categories: List[str]) -> List[str]:
+    result = []
```
mrchtr, line 20: Could you use a `set` instead of `list`? Should be faster to check if an element is in the list. Especially if the collection contains millions of entries.

examples/pipelines/commoncrawl/components/filter_webpage_urls/fondant_component.yaml
```diff
@@ -0,0 +1,25 @@
...
+# image: ghcr.io/ml6team/filter_webpage_urls:latest
+image: 050243278346.dkr.ecr.us-east-1.amazonaws.com/filter_webpage_urls:latest
+
+consumes:
+  webpage:
+    fields:
+      url:
+        type: string
+      html:
+        type: string
```
RobbeSneyders, line 12: Do you actually need the html? I don't think you're using it.

shayorshay: no, we don't use it but we need to keep the whole dataset for the next component in the pipeline.

PR #335, opened 2023-08-03
[LLM pipeline] Update text normalization component

mrchtr:
Add minor improvements to the text normalization component. Mainly based on the work of [Penedo et al ](https://arxiv.org/pdf/2306.01116.pdf)

Quality can be improved by removing specific patterns in single lines:
> We analyse documents line-by-line, and
discard or edit the lines based on the following rules:
• If it is mainly composed of uppercase characters (discard);
• If it is only composed of numerical characters (discard);
• If it is a counter (e.g. 3 likes) (discard);
• If it only contains one word (discard);


components/text_normalization/tests/utils_test.py
```diff
@@ -0,0 +1,42 @@
...
+    line = "HELLO WORLD not upper SOMETHING ELSE IN UPPERCASE"
+    assert ~mainly_uppercase(line, threshold=0.9)
+
+def test_only_numerical():
+    line = "42"
+    assert only_numerical(line)
+
+def test_only_numerical_on_words():
+    line = "42 lorem ipsum"
+    assert ~only_numerical(line)
```
RobbeSneyders, line 24: Nit but `not` is more easily understood than `~`

components/text_normalization/fondant_component.yaml
```diff
@@ -9,12 +9,18 @@ consumes:
...
 
 args:
+  remove_additional_whitespaces:
+    description: If true remove all additional whitespace, tabs.
+    type: bool
   apply_nfc:
     description: If true apply nfc normalization
     type: bool
+  remove_bad_patterns:
+    description: If true remove bad patterns
```
RobbeSneyders, line 19: Could use a better description. It's not clear from this description what bad patterns are.

components/text_normalization/src/main.py
```diff
@@ -44,18 +75,25 @@ def transform(self, dataframe: pd.DataFrame) -> pd.DataFrame:
         Returns:
             Pandas dataframe
         """
+        if self.do_lowercase:
+            dataframe[("text", "data")] = dataframe[("text", "data")].apply(lambda x: x.lower())
+
+        if self.remove_bad_patterns:
+            dataframe[("text", "data")] = dataframe[("text", "data")].apply(remove_noisy_lines)
```
RobbeSneyders, line 82: Shouldn't these be swapped? If you're lower casing first, you're no longer removing the bad uppercase pattern. Or is this desired behavior?

components/text_normalization/src/utils.py
```diff
@@ -0,0 +1,65 @@
+import re
+
+
+def mainly_uppercase(line, threshold=0.7):
+    """
+    Checks if a line is mainly composed of uppercase characters.
+
+    Args:
+        line (str): The input line to check.
```
RobbeSneyders, line 9: Can you move the types in this file to the function signature?

components/text_normalization/tests/fixtures/apply_all.json
RobbeSneyders, line 1: I'm not yet convinced of the json fixtures. I don't see a lot of advantage compared to the python implementation used for the [`caption_images` tests](https://github.com/ml6team/fondant/blob/main/components/caption_images/tests/test_caption_images.py), while we do lose flexibility and add additional complexity.

components/text_normalization/tests/component_test.py
```diff
@@ -0,0 +1,50 @@
+import json
+import os
+import typing as t
+from glob import glob
+
+import pandas as pd
+import pytest
+from fondant.component_spec import ComponentSpec
+
+from components.text_normalization.src.main import TextNormalizationComponent
```
RobbeSneyders, line 10: Why do we import this from the repository root? I would expect us to be working inside the component folder here. So `from src.main import TextNormalizationComponent`.

We'll have to make sure this still works when running as part of the docker build as you proposed in https://github.com/ml6team/fondant/issues/301 though.

components/text_normalization/tests/conftest.py
RobbeSneyders, line 1: Why is this needed? Would like to get rid of this completely.

RobbeSneyders (commented):
Thanks @mrchtr!

Changes to the component look good. Left some comments on the testing strategy.

components/text_normalization/Dockerfile
```diff
@@ -1,18 +1,27 @@
...
+COPY src/ src/
 
+# Run component tests
+FROM base as test
+RUN pip3 install pytest # TODO add pytest to package setup
+ENV PYTHONPATH "${PYTHONPATH}:./src"
+COPY tests/ tests/
+RUN ["python", "-m", "pytest", "tests/"]
+
+FROM base
```
RobbeSneyders, line 24: Should we name this stage as well so we can explicitly invoke it?

```
FROM base as build
```

If we then call

```
docker build --stage=build
```

It should skip the test stage when using buildkit.

RobbeSneyders (commented):
Will you re-add a component test in this PR as well?

components/text_normalization/Dockerfile
mrchtr: We could do it. But we have to make sure to change the build script. Otherwise the next release will fail. 

If we keep it for now as it is, there would be no pressure to update all components immediately. We can still run `docker build .` and only the default stage (without the test stage) is executed.

RobbeSneyders: > We can still run docker build . and only the default stage (without the test stage) is executed.

Ok wasn't aware of this, then this is fine. Let's validate that tests do not run on every build to be sure though.

RobbeSneyders (commented):
Thanks @mrchtr!

For the tests, I would go for option 1. It is simple and readable. I don't think the limited value provided by the abstractions warrants the extra complexity / "magic".

mrchtr:
> For the tests, I would go for option 1. It is simple and readable. I don't think the limited value provided by the abstractions warrants the extra complexity / "magic".

Alright. I have reduced the code to option 1.

RobbeSneyders (approved):
Thanks!

PR #338, opened 2023-08-08
Add int64 dtype

NielsRogge:
When running the Datacomp pipeline at scale, I encountered an outlier image which had a width of about 37,000 pixels. As the component was using int16 as the dtype of the width column, this failed as it was outside the range. The error suggested to use int64 instead.

Then I noticed int64 is not supported yet, hence this PR adds it.

### DIFF
--- a/components/download_images/fondant_component.yaml
+++ b/components/download_images/fondant_component.yaml
@@ -14,9 +14,9 @@ produces:
       data:
         type: binary
       width:
-        type: int16
+        type: int32
       height:
-        type: int16
+        type: int32
     additionalFields: false
 
 args:
--- a/components/filter_image_resolution/fondant_component.yaml
+++ b/components/filter_image_resolution/fondant_component.yaml
@@ -6,9 +6,9 @@ consumes:
   image:
     fields:
       width:
-        type: int16
+        type: int64
       height:
-        type: int16
+        type: int64
 
 args:
   min_image_dim:
--- a/components/filter_image_resolution/fondant_component.yaml
+++ b/components/filter_image_resolution/fondant_component.yaml
@@ -6,9 +6,9 @@ consumes:
   image:
     fields:
       width:
-        type: int64
+        type: int32
       height:
-        type: int64
+        type: int32
 
 args:
   min_image_dim:
--- a/components/image_cropping/fondant_component.yaml
+++ b/components/image_cropping/fondant_component.yaml
@@ -14,9 +14,9 @@ produces:
       data:
         type: binary
       width:
-        type: int16
+        type: int32
       height:
-        type: int16
+        type: int32
 
 args:
   cropping_threshold:
--- a/components/image_resolution_extraction/fondant_component.yaml
+++ b/components/image_resolution_extraction/fondant_component.yaml
@@ -14,6 +14,6 @@ produces:
       data:
         type: binary
       width:
-        type: int16
+        type: int32
       height:
-        type: int16
+        type: int32
--- a/examples/pipelines/datacomp/components/cluster_image_embeddings/fondant_component.yaml
+++ b/examples/pipelines/datacomp/components/cluster_image_embeddings/fondant_component.yaml
@@ -8,9 +8,9 @@ consumes:
       url:
         type: string
       width:
-        type: int16
+        type: int32
       height:
-        type: int16
+        type: int32
       face_bboxes:
         type: array
         items:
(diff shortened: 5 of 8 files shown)


PhilippeMoussalli:
Thanks @NielsRogge, indeed we need to include this data type as an option. However, I think you can opt for `int32` in your component since it ranges from approx -2 Billion to +2 Billion, this will help reduce the size of your parquet files. Or even better a `uint16` since the image size is always positive (provided that you don't run into an outlier that's above 65k) 

https://www.topjavatutorial.com/mobile/swift/swift-data-types/#:~:text=x%20%3AInt%3D%2010-,Integer,-An%20integer%20is

PhilippeMoussalli (approved):
Thanks, some tests still need to be fixed

NielsRogge:
@PhilippeMoussalli I just encountered the issue:
```
File "/usr/local/lib/python3.8/site-packages/dask/dataframe/io/parquet/arrow.py", line 841, in _pandas_to_arrow_table
    raise ValueError(
ValueError: Failed to convert partition to expected pyarrow schema:
    `ArrowInvalid('Integer value 33081 not in range: -32768 to 32767', 'Conversion failed for column height with type int64')`
Expected partition schema:
    embedding: list<item: float>
      child 0, item: float
    face_bboxes: list<item: list<item: float>>
      child 0, item: list<item: float>
          child 0, item: float
    height: int16
    sha256: string
    url: string
    width: int16
    id: string
Received partition schema:
    embedding: list<item: double>
      child 0, item: double
    face_bboxes: list<item: list<item: double>>
      child 0, item: list<item: double>
          child 0, item: double
    height: int64
    sha256: string
    url: string
    width: int64
    id: string
```
So it suggests to use int64, not int32. Should we update to int64?

PhilippeMoussalli:
> @PhilippeMoussalli I just encountered the issue (on a different branch), this is what it looked like:
> 
> ```
> File "/usr/local/lib/python3.8/site-packages/fondant/data_io.py", line 223, in write_dataframe
>     dd.compute(*write_tasks)
>   File "/usr/local/lib/python3.8/site-packages/dask/base.py", line 599, in compute
>     results = schedule(dsk, keys, **kwargs)
>   File "/usr/local/lib/python3.8/site-packages/dask/threaded.py", line 89, in get
>     results = get_async(
>   File "/usr/local/lib/python3.8/site-packages/dask/local.py", line 511, in get_async
>     raise_exception(exc, tb)
>   File "/usr/local/lib/python3.8/site-packages/dask/local.py", line 319, in reraise
>     raise exc
>   File "/usr/local/lib/python3.8/site-packages/dask/local.py", line 224, in execute_task
>     result = _execute_task(task, data)
>   File "/usr/local/lib/python3.8/site-packages/dask/core.py", line 119, in _execute_task
>     return func(*(_execute_task(a, cache) for a in args))
>   File "/usr/local/lib/python3.8/site-packages/dask/optimization.py", line 990, in __call__
>     return core.get(self.dsk, self.outkey, dict(zip(self.inkeys, args)))
>   File "/usr/local/lib/python3.8/site-packages/dask/core.py", line 149, in get
>     result = _execute_task(task, cache)
>   File "/usr/local/lib/python3.8/site-packages/dask/core.py", line 119, in _execute_task
>     return func(*(_execute_task(a, cache) for a in args))
>   File "/usr/local/lib/python3.8/site-packages/dask/dataframe/io/parquet/core.py", line 170, in __call__
>     return self.engine.write_partition(
>   File "/usr/local/lib/python3.8/site-packages/dask/dataframe/io/parquet/arrow.py", line 880, in write_partition
>     t = cls._pandas_to_arrow_table(df, preserve_index=preserve_index, schema=schema)
>   File "/usr/local/lib/python3.8/site-packages/dask/dataframe/io/parquet/arrow.py", line 841, in _pandas_to_arrow_table
>     raise ValueError(
> ValueError: Failed to convert partition to expected pyarrow schema:
>     `ArrowInvalid('Integer value 33081 not in range: -32768 to 32767', 'Conversion failed for column height with type int64')`
> Expected partition schema:
>     embedding: list<item: float>
>       child 0, item: float
>     face_bboxes: list<item: list<item: float>>
>       child 0, item: list<item: float>
>           child 0, item: float
>     height: int16
>     sha256: string
>     url: string
>     width: int16
>     id: string
> Received partition schema:
>     embedding: list<item: double>
>       child 0, item: double
>     face_bboxes: list<item: list<item: double>>
>       child 0, item: list<item: double>
>           child 0, item: double
>     height: int64
>     sha256: string
>     url: string
>     width: int64
>     id: string
> ```

I think what's happening is that Dask converts the given number from python `int` (most likely defaults to int64) to pyarrow `int`. You're still using the old schema which sets the expected datatype to `int16` and the conversion fails. Try again with the new schema and see if you get the same error.

NielsRogge:
Yes it was on a different (outdated) branch, so merging this PR will fix it.

PR #339, opened 2023-08-08
[load_from_hf_hub] Add dataset_length, set_index

NielsRogge:
This PR adds 2 things to the `load_from_hf_hub` reusable component:

- a `dataset_length` argument, which is required in case the user specifies `n_rows_to_load`. The reason why I added this is because I hit an issue when `n_rows_to_load` was larger than the partition size. The current code loads only the first partition, so even though I specified `n_rows_to_load` to be 150k, I only got 69,000 rows. So to solve this I calculate the size of a single partition, then return approximately the requested `n_rows_to_load`.
- adds a monotonically increasing index as suggested by this Stackoverflow post, to solve the issue of duplicate indices due to every partition having indices that start at 0.

### DIFF
--- a/components/load_from_hf_hub/fondant_component.yaml
+++ b/components/load_from_hf_hub/fondant_component.yaml
@@ -1,6 +1,6 @@
 name: Load from hub
 description: Component that loads a dataset from the hub
-image: ghcr.io/ml6team/load_from_hf_hub:dev
+image: ghcr.io/ml6team/load_from_hf_hub:566a5b6fd3d422c98f7890a6e4101d8af10a42bf
 
 produces:
   dummy_variable:  #TODO: fill in here
@@ -22,9 +22,5 @@ args:
     default: None
   n_rows_to_load:
     description: Optional argument that defines the number of rows to load. Useful for testing pipeline runs on a small scale
-    type: int
-    default: None
-  dataset_length:
-    description: Optional argument that defines the length of the dataset. Required in case `n_rows_to_load` is specified.
     type: int
     default: None
--- a/components/load_from_hf_hub/fondant_component.yaml
+++ b/components/load_from_hf_hub/fondant_component.yaml
@@ -1,6 +1,6 @@
 name: Load from hub
 description: Component that loads a dataset from the hub
-image: ghcr.io/ml6team/load_from_hf_hub:566a5b6fd3d422c98f7890a6e4101d8af10a42bf
+image: ghcr.io/ml6team/load_from_hf_hub:dev
 
 produces:
   dummy_variable:  #TODO: fill in here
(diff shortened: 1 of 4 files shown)


PhilippeMoussalli:
Thanks @NielsRogge! 

I tried looking around a bit to see if we can fetch the number of rows from the hf_hub API directly without downloading the dataset to disk. I only found [this](https://huggingface.co/docs/huggingface_hub/v0.16.3/en/package_reference/file_download#huggingface_hub.hf_hub_url:~:text=about%20a%20file-,get_hf_file_metadata,-huggingface_hub.get_hf_file_metadata) method which returns some metadata but the number of rows is not guaranteed to be in there. 

I would opt for a different approach to calculate the required partitions (pseudo code to be tested): 

```python 
if self.n_rows_to_load is not None:
  partitions_length = 0 
  for npartitions, partition in enumerate(dataset.partitions):
       if partitions_length >= n_rows_to_load:
           break 
       logger.info(f"Required number of partitions to load {n_rows_to_load} is {npartitions})
       dataset_length += len(partition)
  dask_df = dask_df.head(n_rows_to_load, npartitions=npartitions)
  dask_df = dd.from_pandas(dask_df, npartitions=npartitions)
```

This has the following advantages: 
* The user does not have to lookup the number of rows from hf hub and input it manually 
* It returns a more approximate number of rows if the partitions are largely imbalanced w.r.t  to the n_rows 

The disadvantage is that you have to called compute however since we're computing only a few individual partitions and not the whole dataset, I expect the overhead to be minimal. 

Regarding the index, the approach sounds sensible. We can include it in the component for now but I will move it to the backend in a separate PR

NielsRogge:
Thanks @PhilippeMoussalli, I've verified your solution and it seems to work. 

Feel free to approve

components/load_from_hf_hub/fondant_component.yaml
```diff
@@ -1,6 +1,6 @@
 name: Load from hub
 description: Component that loads a dataset from the hub
-image: ghcr.io/ml6team/load_from_hf_hub:dev
```
PhilippeMoussalli, line 3: can you revert those back (also below)

examples/pipelines/datacomp/pipeline.py
```diff
@@ -65,17 +64,7 @@
 
 # add ops to pipeline
 pipeline.add_op(load_from_hub_op)
-pipeline.add_op(filter_image_resolution_op, dependencies=load_from_hub_op)
-pipeline.add_op(filter_complexity_op, dependencies=filter_image_resolution_op)
-pipeline.add_op(cluster_image_embeddings_op, dependencies=filter_complexity_op)
+# pipeline.add_op(filter_image_resolution_op, dependencies=load_from_hub_op)
```
PhilippeMoussalli, line 67: you can uncomment those

PhilippeMoussalli (approved):
Thanks! Some small comments to revert things and I think you need to run a pre-commit since some tests are failing

PR #340, opened 2023-08-08
[DataComp] Update pipeline name, remove DockerCompiler

NielsRogge:
This PR updates the pipeline name of DataComp, and makes sure it can run fine both locally and on GCP.

Switching between local vs GCP is currently done as follows:
- I need to manually comment out the base path I don't want to use
- it's either running `fondant run pipeline:pipeline --local` or `python pipeline.py`

Screenshot 2023-08-08 at 13 03 05

### DIFF
--- a/examples/pipelines/datacomp/pipeline.py
+++ b/examples/pipelines/datacomp/pipeline.py
@@ -54,13 +54,6 @@ filter_complexity_op = ComponentOp(
         "min_num_actions": 1,
     },
 )
-cluster_image_embeddings_op = ComponentOp(
-    component_dir="components/cluster_image_embeddings",
-    arguments={
-        "sample_ratio": 0.3,
-        "num_clusters": 3,
-    },
-)
 
 # add ops to pipeline
 pipeline.add_op(load_from_hub_op)
--- a/examples/pipelines/datacomp/pipeline.py
+++ b/examples/pipelines/datacomp/pipeline.py
@@ -13,10 +13,10 @@ logger = logging.getLogger(__name__)
 
 # Initialize pipeline and client
 pipeline = Pipeline(
-    pipeline_name="datacomp-filtering",
+    pipeline_name="datacomp-filtering-pipeline",
     pipeline_description="A pipeline for filtering the Datacomp dataset",
-    # base_path=PipelineConfigs.BASE_PATH,
-    base_path="/Users/nielsrogge/Documents/fondant_artifacts_datacomp",
+    base_path=PipelineConfigs.BASE_PATH,
+    # base_path="/Users/nielsrogge/Documents/fondant_artifacts_datacomp",
 )
 client = Client(host=PipelineConfigs.HOST)
 
@@ -66,5 +66,7 @@ cluster_image_embeddings_op = ComponentOp(
 pipeline.add_op(load_from_hub_op)
 pipeline.add_op(filter_image_resolution_op, dependencies=load_from_hub_op)
 pipeline.add_op(filter_complexity_op, dependencies=filter_image_resolution_op)
-pipeline.add_op(cluster_image_embeddings_op, dependencies=filter_complexity_op)
 # TODO add more ops
+
+
+client.compile_and_run(pipeline=pipeline)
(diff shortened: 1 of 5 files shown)


examples/pipelines/datacomp/pipeline.py
```diff
@@ -67,15 +66,7 @@
 pipeline.add_op(load_from_hub_op)
 pipeline.add_op(filter_image_resolution_op, dependencies=load_from_hub_op)
 pipeline.add_op(filter_complexity_op, dependencies=filter_image_resolution_op)
-pipeline.add_op(cluster_image_embeddings_op, dependencies=filter_complexity_op)
```
NielsRogge, line 70: The cluster image embeddings component won't be used for now, which is why it is removed here.

PhilippeMoussalli:
> Switching between local vs GCP is currently done as follows:
> 
> * I need to manually comment out the base path I don't want to use

Isn't this just changing the base path in `PipelineConfigs`? you can use a local base path or a remote one for the local runner. For the remote runner, it's always the remote one. 

> * it's either running `fondant run pipeline:pipeline --local` or `python pipeline.py`  

I think you can keep the interface and code of the pipeline as is, for now only `fondant run pipeline:pipeline --local` running the pipeline with kubeflow using a `fondant run` should be soon available with this PR #291 (@GeorgesLorre can you confirm?).

NielsRogge:
> Isn't this just changing the base path in PipelineConfigs? you can use a local base path or a remote one for the local runner. For the remote runner, it's always the remote one.

Yes but as I'm switching regularly between both, it's handy to specify both a local and a remote base path rather than having to type it everytime. cc @GeorgesLorre

examples/pipelines/datacomp/pipeline.py
```diff
@@ -51,20 +50,15 @@
...
 
 # add ops to pipeline
 pipeline.add_op(load_from_hub_op)
 pipeline.add_op(filter_image_resolution_op, dependencies=load_from_hub_op)
 pipeline.add_op(filter_complexity_op, dependencies=filter_image_resolution_op)
-pipeline.add_op(cluster_image_embeddings_op, dependencies=filter_complexity_op)
 # TODO add more ops
+
+
+client.compile_and_run(pipeline=pipeline)
```
RobbeSneyders, line 64: Can you put this in a `if __name__ == "__main__":` block so it's not executed when running `fondant run pipeline:pipeline`.

examples/pipelines/datacomp/pipeline.py
```diff
@@ -13,10 +13,10 @@
 
 # Initialize pipeline and client
 pipeline = Pipeline(
-    pipeline_name="datacomp-filtering",
+    pipeline_name="datacomp-filtering-pipeline",
     pipeline_description="A pipeline for filtering the Datacomp dataset",
-    # base_path=PipelineConfigs.BASE_PATH,
-    base_path="/Users/nielsrogge/Documents/fondant_artifacts_datacomp",
+    base_path=PipelineConfigs.BASE_PATH,
+    # base_path="/Users/nielsrogge/Documents/fondant_artifacts_datacomp",
```
RobbeSneyders, line 19: Would prefer not to commit this, can you leave this out?

RobbeSneyders (commented):
Thanks @NielsRogge!

GeorgesLorre:
> > Isn't this just changing the base path in PipelineConfigs? you can use a local base path or a remote one for the local runner. For the remote runner, it's always the remote one.
> 
> Yes but as I'm switching regularly between both, it's handy to specify both a local and a remote base path rather than having to type it everytime. cc @GeorgesLorre

I'm working on making the interface for docker and kfp more similar (TBD).

But you can alter your pipeline.py already to make this easier:
```
if __name__ == "__main__":
    # assuming kfp run
    client = Client(host=PipelineConfigs.HOST)
    # override the base_path
    pipeline.base_path = "...."
    client.compile_and_run(pipeline=pipeline)
```    
So if you call your pipeline.py it will assume it is a kfp run and  override the base_path and if you use `fondant run pipeline:pipeline --local` it should not run the code under the if statement.

RobbeSneyders (approved):

PR #341, opened 2023-08-08
Add kfp constraint

PhilippeMoussalli:
Add constrain to KFP version to avoid downloading latest by default (v2) which has a different API

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,7 +51,7 @@ fsspec = { version = ">= 2023.4.0", optional = true}
 gcsfs = { version = ">= 2023.4.0", optional = true }
 s3fs = { version = ">= 2023.4.0", optional = true }
 adlfs = { version = ">= 2023.4.0", optional = true }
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -51,7 +51,7 @@ fsspec = { version = ">= 2023.4.0", optional = true}
 gcsfs = { version = ">= 2023.4.0", optional = true }
 s3fs = { version = ">= 2023.4.0", optional = true }
 adlfs = { version = ">= 2023.4.0", optional = true }
-kfp = { version = ">= 1.8.19", optional = true }
+kfp = { version = ">= 1.8.19, <= 1.8.22", optional = true }
```
RobbeSneyders, line 54:
```suggestion
kfp = { version = ">= 1.8.19, < 2", optional = true }
```

Can we put the upper-bound at the major version instead?

RobbeSneyders (approved):

PR #343, opened 2023-08-08
Small improvements to tox configuration

RobbeSneyders:
Had issues running tox locally which in the end were unrelated to the configuration. But in the meantime I learned some better practices for tox, which I implemented in this PR.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -13,22 +13,28 @@ python =
     3.11: py311
 
 [testenv:pre-commit]
+skip_install=true
 allowlist_externals=
     pre-commit
 commands=pre-commit run --all-files --show-diff-on-failure
 
 [testenv:check-licenses]
+skip_install=true
 deps=
     liccheck>=0.7.3,<0.8
-commands=
+allowlist_externals=
+    poetry
+commands_pre=
     poetry lock
     poetry install --all-extras
     poetry show
     poetry export -f requirements.txt --output /tmp/requirements.txt
+commands=
     poetry run liccheck -s license_strategy.ini -r /tmp/requirements.txt -l PARANOID
 
-[testenv]
+[testenv:py{38,39,310,311}]
 setenv=PYTHONPATH = {toxinidir}:{toxinidir}
+skip_install=true
 deps=
     pytest>=7.4.0,<8
     pytest-cov>=4.1.0,<5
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -13,33 +13,41 @@ python =
...
+commands_pre=
     poetry lock
     poetry install --all-extras
     poetry show
     poetry export -f requirements.txt --output /tmp/requirements.txt
+commands=
     poetry run liccheck -s license_strategy.ini -r /tmp/requirements.txt -l PARANOID
 
-[testenv]
+[testenv:py{38,39,310,311}]
```
RobbeSneyders, line 35: No longer inherit this configuration in the other test environments, which lead to unnecessary dependency installation etc.

tox.ini
```diff
@@ -13,33 +13,41 @@ python =
...
     /usr/bin/bash
-commands=
+commands_pre=
     bash ./scripts/pre-build.sh
     poetry lock
     poetry install --all-extras
     poetry show
+commands=
     poetry run python -m pytest tests -vv --cov fondant --cov-report term-missing
+commands_post=
```
RobbeSneyders, line 52: Split pre and post commands so post script always run, even when command fails.

tox.ini
```diff
@@ -13,33 +13,41 @@ python =
...
 allowlist_externals=
     pre-commit
 commands=pre-commit run --all-files --show-diff-on-failure
 
 [testenv:check-licenses]
+skip_install=true
 deps=
     liccheck>=0.7.3,<0.8
-commands=
+allowlist_externals=
```
RobbeSneyders, line 25: No longer inherited

tox.ini
```diff
@@ -13,33 +13,41 @@ python =
     3.11: py311
 
 [testenv:pre-commit]
+skip_install=true
```
RobbeSneyders, line 16: Not necessary to install fondant.

mrchtr (approved):
Thanks @RobbeSneyders! I have tested your changes locally, and everything is working.

GeorgesLorre:
Should be faster as well no ? With less installing of packages.

RobbeSneyders:
> Should be faster as well no ? With less installing of packages.

Yes, running a single pytest environment takes ~20 seconds for me now.

PR #345, opened 2023-08-08
Add `index_column` and unique index creation to load_from_hf_hub component

PhilippeMoussalli:
Based on [this](https://stackoverflow.com/questions/47571715/dask-create-strictly-increasing-index)

PR that adds a monotonically increasing index to the initial loaded dataframe by default. This is beneficial in case the dataset does not come with a given reliable index.  Resetting the index should be avoided since it can assign duplicate indices across partitions. 

The user can always disable the automatic partitioning in case they have a pre-defined index column they want to set or they want to create their own initial custom index.


src/fondant/executor.py
```diff
@@ -265,7 +277,16 @@ def _execute_component(
...
         """
-        return component.load()
+        dask_df = component.load()
+
+        if self.disable_automatic_indexing is False:
+            # Set monotonically increasing index
+            logger.info("Setting the index...")
+            dask_df["id"] = 1
+            dask_df["id"] = dask_df.id.cumsum()
+            dask_df = dask_df.reset_index(drop=True).set_index("id", sort=True)
```
RobbeSneyders, line 287: Is the `reset_index` necessary before `set_index`?

RobbeSneyders (commented):
Thanks @PhilippeMoussalli!

Some high level questions / notes:
- What is the default index for the huggingface hub if this reindexing is disabled? Is it just a range index per partition, or does it inherit an index from huggingface?
- In the case that the index is not inherited from huggingface, is there a way for the user to indicate which column to use as the index?
- The link in your description mentions that `map_partitions` doesn't provide the partition number, but this is not / no longer correct. You can get this info via the [`partition_info`](https://docs.dask.org/en/stable/generated/dask.dataframe.DataFrame.map_partitions.html) argument. Not sure if this allows for a more efficient method.

PhilippeMoussalli:
> Thanks @PhilippeMoussalli!
> 
> Some high level questions / notes:
> 
> * What is the default index for the huggingface hub if this reindexing is disabled? Is it just a range index per partition, or does it inherit an index from huggingface?

I'm not sure if it's uniform across all HF datasets. So far, I just see it as the range index per partition since it has no set index (the index in datacomp is just a regular column). There might be other situations where the dataset is written with a set index but I have not run into those.

> * In the case that the index is not inherited from huggingface, is there a way for the user to indicate which column to use as the index?

I think we can maybe replace the argument from a boolean to a string indicating the column name. If it's not specified, then we can set the index internally. Wdyt? 

> * The link in your description mentions that `map_partitions` doesn't provide the partition number, but this is not / no longer correct. You can get this info via the [`partition_info`](https://docs.dask.org/en/stable/generated/dask.dataframe.DataFrame.map_partitions.html) argument. Not sure if this allows for a more efficient method.

oh did not realize this! 

I just tried it out 

```
df = pd.DataFrame(range(10), columns=["a"])

ddf = dd.from_pandas(df, npartitions=3)

def my_func(d, x, partition_info=None):
    print(partition_info)

ddf.map_partitions(my_func, 3, meta=df.head()).compute()
```

and it gives 

```
{'number': 0, 'division': 0}
{'number': 1, 'division': 4}
{'number': 2, 'division': 7}
```

We might just be able to paralelize setting the index that way if we know the divisions (increment starting from each division) but I would have to check this.

src/fondant/executor.py
PhilippeMoussalli: I noticed that if an index is already set, using `set_index` will not change the initial index. I will check if there is a way to find out if an index is set

PhilippeMoussalli:
Dask graph without map partitions (cumsum)
without_map


Dask graph with map partitions  
with_map

src/fondant/executor.py
PhilippeMoussalli: I removed it, it was because I was setting the index twice in a graph (as part of a test) so it was only considering the first one

RobbeSneyders:
> Dask itself seems to set a monotonically increasing index by itself even with multiple partitions **provided you don't called reset_index()**. See [link](https://www.coiled.io/blog/dask-set-index-dataframe#:~:text=you%20use%20it.-,set_index%20syntax,-Let%E2%80%99s%20create%20a)

So I assume this is not the case when loading a dataset from the HF hub?

> * However, you can still have a custom index that is set on the huggingface dataframe (tested this by setting an index, writing to hub and reading the dataset again). It's not very feasible to know if this is the case just by looking at the dataset in the hub or checking any flag on the dataset after it's loaded.

Can you share the link to such a dataset?

> * Since we can't tell if an index is set on the HF dataset, I think the most foolproof method is to set the index by default or provide the user the ability to set it based on one of the existing columns. In the latter case, the index should not be dropped in case the user wants to use it in subsequent components for other non-index related functionalities.

I see no reason not to drop it. You can always restore an index to a column (eg. with `reset_index`). I wouldn't keep duplicate information.

> I tested using the `paritions_info` argument and it only works in case of proper divisions, otherwise it returns duplicate values in cases with duplicated indices making it unusable for setting a unique index
> 
> ```
> p_info
> {'number': 0, 'division': 0}
> p_infop_info
> 
> {'number': 1, 'division': 7}
> {'number': 0, 'division': 0}
> p_info
> {'number': 1, 'division': 7}
> p_info
> {'number': 0, 'division': 0}
> p_info
> {'number': 1, 'division': 7}
> ```

I don't understand the inconsistency here. The `division` might not always line up, but the `number` should be unique. We could create a unique index with a format like `{number:03d}{cumsum:10d}`

PhilippeMoussalli:
> > Dask itself seems to set a monotonically increasing index by itself even with multiple partitions **provided you don't called reset_index()**. See [link](https://www.coiled.io/blog/dask-set-index-dataframe#:~:text=you%20use%20it.-,set_index%20syntax,-Let%E2%80%99s%20create%20a)
> 
> So I assume this is not the case when loading a dataset from the HF hub?

When the dataset is loaded from the hub, it won't have any index unless it's written. However, in the case a dataset has a written index to it. That index won't be set. 

> > * However, you can still have a custom index that is set on the huggingface dataframe (tested this by setting an index, writing to hub and reading the dataset again). It's not very feasible to know if this is the case just by looking at the dataset in the hub or checking any flag on the dataset after it's loaded.
>
> Can you share the link to such a dataset?

https://huggingface.co/datasets/philippemo/test_index

> > * Since we can't tell if an index is set on the HF dataset, I think the most foolproof method is to set the index by default or provide the user the ability to set it based on one of the existing columns. In the latter case, the index should not be dropped in case the user wants to use it in subsequent components for other non-index related functionalities.
> 
> I see no reason not to drop it. You can always restore an index to a column (eg. with `reset_index`). I wouldn't keep duplicate information.

Alright will remove it. 
 
> > I tested using the `paritions_info` argument and it only works in case of proper divisions, otherwise it returns duplicate values in cases with duplicated indices making it unusable for setting a unique index
> > ```
> > p_info
> > {'number': 0, 'division': 0}
> > p_infop_info
> > 
> > {'number': 1, 'division': 7}
> > {'number': 0, 'division': 0}
> > p_info
> > {'number': 1, 'division': 7}
> > p_info
> > {'number': 0, 'division': 0}
> > p_info
> > {'number': 1, 'division': 7}
> > ```
> 
> I don't understand the inconsistency here. The `division` might not always line up, but the `number` should be unique. We could create a unique index with a format like `{number:03d}{cumsum:10d}`

Yeah it was also a bit odd for me. Here is the script to reproduce this. I generated duplicate indices on purpose to check if the method can work against it. But you will see that this still results in duplicated indices

```python
nrange = 100

def generate_random_numbers(n):
    return np.random.randint(0, 1000, size=n)

df = pd.DataFrame(
    {"col1": generate_random_numbers(nrange),
     "col2": generate_random_numbers(nrange),
     "col3": generate_random_numbers(nrange),
     "col4": generate_random_numbers(nrange)}
)
ddf = dd.from_pandas(df, npartitions=10)
ddf = ddf.repartition(npartitions=10)
ddf = ddf.reset_index()
ddf = dd.from_pandas(ddf.compute(), npartitions=10)
print("number of partitions")
print(ddf.npartitions)
ddf["id"] = 1

def my_func(dataframe: pd.DataFrame, partition_info=None):
    print("p_info")
    print(partition_info)
    dataframe["id"] = dataframe.id.cumsum() + partition_info["division"]
    return dataframe

ddf = ddf.map_partitions(my_func,  meta=ddf.head())
ddf = ddf.set_index("id",drop=False)
print(ddf.compute())
```

RobbeSneyders:
I think it can work if we use the `number` from `partition_info` instead of `division`. Something like this:
```python
nrange = 100

def generate_random_numbers(n):
    return np.random.randint(0, 1000, size=n)

df = pd.DataFrame(
    {"col1": generate_random_numbers(nrange),
     "col2": generate_random_numbers(nrange),
     "col3": generate_random_numbers(nrange),
     "col4": generate_random_numbers(nrange)}
)
ddf = dd.from_pandas(df, npartitions=10)
ddf = ddf.repartition(npartitions=10)
ddf = ddf.reset_index()
ddf = dd.from_pandas(ddf.compute(), npartitions=10)
print("number of partitions")
print(ddf.npartitions)
ddf["id"] = 1

def my_func(dataframe: pd.DataFrame, partition_info=None):
    print("p_info")
    print(partition_info)
    dataframe["id"] = partition_info['number'] * 1000000000 + dataframe.id.cumsum()
    dataframe["id"] = dataframe["id"].apply(lambda x: str(x).zfill(10))
    return dataframe

ddf = ddf.map_partitions(my_func,  meta=ddf.head())
ddf = ddf.set_index("id", sort=False)
print(ddf.compute())
```

There's probably a better way to calculate and assign this though.

PhilippeMoussalli:
> I think it can work if we use the `number` from `partition_info` instead of `division`. Something like this:
> There's probably a better way to calculate and assign this though.

I think this solution can be valid:

```python
def generate_random_numbers(n):
    return np.random.randint(0, 1000, size=n)


df = pd.DataFrame(
    {"col1": generate_random_numbers(nrange),
     "col2": generate_random_numbers(nrange),
     "col3": generate_random_numbers(nrange),
     "col4": generate_random_numbers(nrange)}
)
ddf = dd.from_pandas(df, npartitions=10)
ddf = ddf.repartition(npartitions=10)
ddf = ddf.reset_index(drop=True)
ddf = dd.from_pandas(ddf.compute(), npartitions=10)
print("number of partitions")
print(ddf.npartitions)


def my_func(dataframe: pd.DataFrame, partition_info=None):
    dataframe["id"] = 1
    dataframe["id"] = str(partition_info['number']) + "_" + (dataframe.id.cumsum()).astype(str)
    dataframe.index = dataframe.pop("id")
    return dataframe


ddf = ddf.map_partitions(my_func, meta=ddf.head())
print(ddf.compute())
```

result: 
```
      col1  col2  col3  col4
id                          
0_1    943   961   427    45
0_2    283   592   596   985
0_3     42   177   466    25
0_4    297   716    36   460
0_5    786   959   334   314
...    ...   ...   ...   ...
9_6    146   464   704   370
9_7    788   841    83   218
9_8    654   178   154   181
9_9    156   901   505    52
9_10   676   424   631   842

[100 rows x 4 columns]
```

We don't have to make an assumption on the size of the dataset and the partition/row_number are more visible. 
What do you think about the format itself? should we keep it like this or include strings? 
e.g. (`p_<partition_number>-<row_number>`)

RobbeSneyders:
That makes sense indeed! :sweat_smile:

RobbeSneyders:
Related:
https://github.com/ml6team/fondant/issues/363

src/fondant/executor.py
```diff
@@ -254,7 +268,32 @@ def _execute_component(
...
+                    + "_"
+                    + (dataframe.id.cumsum()).astype(str)
+                )
+                dataframe.index = dataframe.pop("id")
+                return dataframe
+
+            dask_df = dask_df.map_partitions(_set_unique_index, meta=dask_df.head())
+        else:
+            logger.info(f"Setting `{self.index_column}` as index")
+            dask_df = dask_df.set_index(self.index_column, drop=True)
```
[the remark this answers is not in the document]
RobbeSneyders: That might make sense, but I feel like we're adding a lot of arguments like this which increase the complexity. Since we only want to add this for one component at the moment, it might make more sense to just add it there and check if / how we want to generalize in the future.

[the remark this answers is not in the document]
PhilippeMoussalli: alright that makes sense, made the changes

PR #348, opened 2023-08-08
[DataComp] Add download images component

NielsRogge:
This PR:

splits up the datacomp folder into 2 pipelines:

- a simple pipeline, just consisting of 3 components, serving as a simple baseline which could serve as a first submission
- a more advanced pipeline, which involves downloading images (using the reusable `download_images` component), and later on also text detection and text recognition 

and improves the `download_images` component to leverage Dask's `map_partitions`.

### DIFF
--- a/components/download_images/src/main.py
+++ b/components/download_images/src/main.py
@@ -89,7 +89,7 @@ def download_image_with_retry(
 def download_image_with_retry_partition(dataframe, timeout, retries, resizer):
     # process a single partition
     # TODO make column name more flexible
-    data = dataframe.image_url.apply(lambda x:
+    data = dataframe.images_url.apply(lambda x:
         download_image_with_retry(
             url=x, timeout=timeout, retries=retries, resizer=resizer,
         ),
@@ -164,9 +164,9 @@ class DownloadImagesComponent(DaskTransformComponent):
 
         # rename new columns to be conform the spec
         logger.info("Renaming columns...")
-        dataframe = dataframe.rename(columns={"data": "image_data",
-                                              "width": "image_width",
-                                              "height":"image_height"})
+        dataframe = dataframe.rename(columns={"data": "images_data",
+                                              "width": "images_width",
+                                              "height":"images_height"})
 
         # Remove images that could not be fetched
         logger.info("Dropping invalid rows...")
--- a/components/download_images/src/main.py
+++ b/components/download_images/src/main.py
@@ -86,6 +86,24 @@ def download_image_with_retry(
     return None, None, None
 
 
+def download_image_with_retry_partition(dataframe, timeout, retries, resizer):
+    # process a single partition
+    # TODO make column name more flexible
+    data = dataframe.image_url.apply(lambda x:
+        download_image_with_retry(
+            url=x, timeout=timeout, retries=retries, resizer=resizer,
+        ),
+    )
+
+    # use assign to add values as extra columns
+    dataframe = dataframe.assign(data=[example[0] for example in data],
+                   width=[example[1] for example in data],
+                   height=[example[2] for example in data],
+    )
+
+    return dataframe
+
+
 class DownloadImagesComponent(DaskTransformComponent):
     """Component that downloads images based on URLs."""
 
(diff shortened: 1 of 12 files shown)


components/download_images/src/main.py
```diff
@@ -123,35 +141,43 @@ def __init__(self,
...
+
+        # create meta
+        # needs to be a dictionary with keys = column names, values = dtypes of columns
+        # for each column in the output
+        meta = dict(zip(dataframe.columns, dataframe.dtypes))
+        meta["data"] = bytes
+        meta["width"] = int
+        meta["height"] = int
+
+        dataframe = dataframe.map_partitions(
```
RobbeSneyders, line 161: Is there still a reason to use a `DaskTransformComponent` for this component? If we use the `PandasTransformComponent`, we get the `map_partitions` for free.

examples/pipelines/datacomp/components/filter_text_complexity/src/main.py
```diff
@@ -27,6 +27,7 @@ def get_text_complexity(doc: spacy.tokens.doc.Doc):
     return complexity
 
 
+# this function can also be supported
 def get_num_actions(doc: spacy.tokens.doc.Doc):
```
RobbeSneyders, line 31: Would remove this function if it's not used. We have it in git, so can always re-add it when needed.

examples/pipelines/datacomp/components/load_from_hf_hub/fondant_component.yaml
```diff
@@ -27,6 +27,8 @@ produces:
   
   image_text:
     fields:
+      uid:
```
RobbeSneyders, line 30: Shouldn't this be used as the index? Is this missing functionality in the current `read_from_hub` component?

RobbeSneyders (commented):
Thanks @NielsRogge!

Did you notice any speedup by using `map_partitions` + `apply` instead of just `apply`?

Don't forget to revert all your local changes before merging.

examples/pipelines/datacomp/components/load_from_hf_hub/fondant_component.yaml
NielsRogge: Currently the `load_from_hf_hub` creates a monotonically increasing index: https://github.com/ml6team/fondant/blob/92942afa0d43e150ce77b15f7cc4039d75f222f1/components/load_from_hf_hub/src/main.py#L62-L66.

@PhilippeMoussalli 's PR will remove that and set it in Fondant: https://github.com/ml6team/fondant/pull/345

RobbeSneyders: Yes, but the `uid` is a field from the initial huggingface dataset we're loading? Then I would assume that ideally, we use that column as the index in Fondant?

NielsRogge: Ok but there's no way to specify that currently, right? Would we need to add an `index_column` argument to the `load_from_hf_hub` component?

RobbeSneyders: Yes, that's what I was steering towards. No need to add it in this PR, but important to know that that would be useful.

@PhilippeMoussalli related to my comment on #345

RobbeSneyders:
@NielsRogge LGTM if you can revert all the changes that should not be in git (the commit hashes etc.)

examples/pipelines/datacomp/components/load_from_hf_hub/fondant_component.yaml
```diff
@@ -1,9 +1,9 @@
 name: Load from hub
 description: Component that loads a dataset from the hub
-image: ghcr.io/ml6team/load_from_hf_hub:dev
+image: ghcr.io/ml6team/load_from_hf_hub:ecf9e620f1ff291236f67d752407fc01de1212cf
```
RobbeSneyders, line 3:
```suggestion
image: ghcr.io/ml6team/load_from_hf_hub:dev
```

RobbeSneyders (approved):

PR #350, opened 2023-08-10
Remove node_pool_name arguments in examples

RobbeSneyders:
Our current example pipelines don't work anymore out of the box since https://github.com/ml6team/fondant/pull/327 requires `node_pool_label` and `node_pool_name` to either both be set, or neither.

This PR removes the `node_pool_name` since it is very specific to our own kfp cluster anyway.


PhilippeMoussalli (approved):
Thanks Robbe!

PR #353, opened 2023-08-14
Add AWS credential arguments to commoncrawl download components. 

mrchtr:
Incorporating the `aws_access_key_id` and `aws_secret_access_key` as arguments into the common crawl download components. This enables access to the common crawl S3 buckets from any given environment.


RobbeSneyders (approved):
Thanks @mrchtr!

PR #355, opened 2023-08-14
Remove output_partition_size argument and logic

PhilippeMoussalli:
PR to remove automatic logic of repartitioning the output based on size


RobbeSneyders (approved):
Thanks @PhilippeMoussalli!

PR #356, opened 2023-08-14
Remove obsolete args from ComponentOp

PhilippeMoussalli:
PR that remove p_volumes and ephemeral storage arguments (previously needed on the assumption that the laion dataset was going to be downloaded in memory)


RobbeSneyders (approved):
Thanks @PhilippeMoussalli, can you resolve the conflict?

PR #358, opened 2023-08-15
[Commoncrawl pipeline] Add offset to load component

shayorshay:
This PR allows selecting segments by offsets. This is useful if we cannot run all segments at a go.


RobbeSneyders (approved):
Thanks @shayorshay!

PR #359, opened 2023-08-16
Implement kfp runner with tests

GeorgesLorre:
PR implementing the Kfp runner.

Had some issues with adding the pipeline name to the kfp spec (solved now) because we need it to launch a run with a sensible name.

Remarks:
- removed some logic around deleting existing pipelines that we have (which was not used afaik)
- I tested it on a toy pipeline and was able to push a run on the GCP kubeflow
- I will implement the CLI (for both the runner and compiler) in a separate PR 
- I will also make a separate PR for cleaning up the kfp imports and extras.

### DIFF
--- a/src/fondant/compiler.py
+++ b/src/fondant/compiler.py
@@ -226,7 +226,7 @@ class KubeFlowCompiler(Compiler):
             self.kfp = kfp
         except ImportError:
             msg = """You need to install kfp to use the Kubeflow compiler,\n
-                     you can install it with `pip install --extras kfp`"""
+                     you can install it with `pip install --extras pipelines`"""
             raise ImportError(
                 msg,
             )
--- a/src/fondant/compiler.py
+++ b/src/fondant/compiler.py
@@ -226,7 +226,7 @@ class KubeFlowCompiler(Compiler):
             self.kfp = kfp
         except ImportError:
             msg = """You need to install kfp to use the Kubeflow compiler,\n
-                     you can install it with `pip install --extras pipelines`"""
+                     you can install it with `pip install --extras kfp`"""
             raise ImportError(
                 msg,
             )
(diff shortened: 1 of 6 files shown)


src/fondant/compiler.py
```diff
@@ -226,7 +226,7 @@ def _resolve_imports(self):
             self.kfp = kfp
         except ImportError:
             msg = """You need to install kfp to use the Kubeflow compiler,\n
-                     you can install it with `pip install --extras pipelines`"""
+                     you can install it with `pip install --extras kfp`"""
```
PhilippeMoussalli, line 229: I think the extras is still called `pipelines`. You can however remove the `k8s` dependency from it since we're no longer using it and maybe rename `pipelines` to `kfp`

tests/test_runner.py
```diff
@@ -20,3 +26,54 @@ def test_docker_runner():
...
+
+
+def test_kubeflow_runner():
+    input_spec_path = str(VALID_PIPELINE / "kubeflow_pipeline.yml")
+    with mock.patch(
+        "kfp.Client",
+        new=MockKfpClient,
+    ):
+        runner = KubeflowRunner(host="some_host")
+        runner.run(input_spec=input_spec_path)
```
PhilippeMoussalli, line 57: What are we testing here exactly? I know it's difficult since we're mocking the pipeline run method. Can we maybe validate the `run_url`?

PhilippeMoussalli (approved):
Thanks Georges! Really nice implementation

Left a few minor comments but should be good to go

src/fondant/compiler.py
GeorgesLorre: I will just update this and rename the `pipelines` extra to `kfp` in a different PR

tests/test_runner.py
GeorgesLorre: We are testing nothing explicit but if the runner would error then the test would fail. I'll add an assert to see if the url is passed properly.

GeorgesLorre:
@PhilippeMoussalli resolved your comments

PhilippeMoussalli:
> @PhilippeMoussalli resolved your comments

Thanks :) good to merge

PR #360, opened 2023-08-16
Adjust bandit settings

mrchtr:
To avoid security breaches bandit should scan the pipeline example folders and tests as well.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -96,3 +96,6 @@ line-length = 100  # Only for comments, as code is handled by black at length 88
 
 [tool.ruff.pydocstyle]
 convention = "google"
+
+[tool.bandit]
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -96,3 +96,6 @@ line-length = 100  # Only for comments, as code is handled by black at length 88
 
 [tool.ruff.pydocstyle]
 convention = "google"
+
+[tool.bandit]
+skips = ["B101", "B108", "B307", "B404", "B601", "B602"]
```
RobbeSneyders, line 101: Can we fix or locally ignore some of these issues instead of skipping them entirely?

We should be able to skip B101 for non-test files only. See https://bandit.readthedocs.io/en/latest/plugins/b101_assert_used.html

PR #362, opened 2023-08-16
Hide executor from users

PhilippeMoussalli:
Draft PR that addresses this #300 

The new interface will be changed from this:

```python
if __name__ == "__main__":
    executor = PandasTransformExecutor.from_args()
    executor.execute(MyComponent)
```

to this: 

```python
if __name__ == "__main__":
    component_runner = ComponentRunner(MyComponent)
    component_runner.run()
```

I'v only changed one example in the tests to showcase how it would look like. All feedback is welcome


PhilippeMoussalli:
> Thanks @PhilippeMoussalli. My original idea was to get rid of the `if __name__ == "__main__":` block completely by executing a fondant entry point instead.
> 
> Eg. something like `fondant execute src/main.py:CustomComponent`

ah ok, I guess then this command will be specified in the Dockerfile entrypoint instead of  `python/main.py`. 
If I understand correctly, `CustomComponent` will refer to the component class name so it will not be the same for different components. Does that mean the Dockerfile entrypoint will have to be different depending on the component as well?

RobbeSneyders:
Yes, unless we detect it automatically by eg. finding a Component subclass in the file. This was just my original idea though, open to other suggestions.

CC @mrchtr I think you commented on my PR originally to hide the executor as well. Any opinion?

src/fondant/cli.py
```diff
@@ -63,13 +66,14 @@ def entrypoint():
...
     register_explore(subparsers)
+    register_execute(subparsers)
     register_compile(subparsers)
     register_run(subparsers)
 
     sys.path.append(".")
 
     # display help if no arguments are provided
-    args = parser.parse_args(sys.argv[1:] or ["--help"])
+    args, _ = parser.parse_known_args(sys.argv[1:] or ["--help"])
```
PhilippeMoussalli, line 76: this is needed to avoid parsing the kubeflow arguments when calling `fondant execute main`

src/fondant/executor.py
```diff
@@ -123,7 +123,8 @@ def _add_and_parse_args(cls, spec: ComponentSpec):
                 help=arg.description,
             )
 
-        return parser.parse_args()
+        args, _ = parser.parse_known_args()
```
PhilippeMoussalli, line 126: this is needed to avoid parsing the `ref` when running `fondant execute main` and only parse kubeflow arguments

PhilippeMoussalli:
> Yes, unless we detect it automatically by eg. finding a Component subclass in the file. This was just my original idea though, open to other suggestions.
> 
> CC @mrchtr I think you commented on my PR originally to hide the executor as well. Any opinion?

Made some additional changes and tested the current approach, it works fine. If we  agree on the current approach, i'll continue the rest of the PR (modify docs, make changes to rest of testes, change all the dockerfiles and components, ...)

src/fondant/cli.py
RobbeSneyders: This will ignore unknown arguments for other commands as well though, which might not be what we want. Maybe we can try something like:

```python
args, _ = parser.parse_known_args(sys.argv[1:] or ["--help"])
if args.func != execute:
    args = parser.parse_args(sys.argv[1:] or ["--help"])
```

src/fondant/cli.py
```diff
@@ -360,3 +393,54 @@ def pipeline_from_string(import_string: str) -> Pipeline:
...
+                )
+                raise ImportFromModuleError(msg)
+
+            component = cls
+
+    if component is None:
+        msg = f"No Component found in module {module_str}"
+        raise ImportFromModuleError(msg)
+
+    return component
```
RobbeSneyders, line 446: This will break if the user creates any base component classes to inherit from in their own project. We can probably make this a bit more robust by taking the component class with the max order. Something like this (untested):

```python
    component_classes = defaultdict(list)

    for name, cls in class_members:
        if issubclass(cls, BaseComponent):
            order = len(cls.__mro__)
            component_classes[order].append(cls)
            
    max_order = max(component_classes)
    found_components = component_classes[max_order]
    
    if len(found_components) == 0:
        msg = f"No Component found in module {module_str}"
        raise ImportFromModuleError(msg)
    if len(found_components) > 1:
        msg = (
            f"Found multiple components in {module_str}: {found_components}. Only one component "
            f"can be present"
        )
        raise ImportFromModuleError(msg)

    logger.info(f"Component `{name}` found in module {module_str}")
    return found_components[0]
```

RobbeSneyders (commented):
Thanks Philippe, this looks like a good approach to me. Left some comments.

src/fondant/cli.py
PhilippeMoussalli: Tried it out with small tweaks and it works

src/fondant/runner.py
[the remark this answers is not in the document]
PhilippeMoussalli: I tried that initially but it results in circular imports, will move it at the bottom of `executor` but open to other suggestions

RobbeSneyders, line 1: I think the changes in this file can be removed.

PhilippeMoussalli: woops forgot those ones, good catch

src/fondant/executor.py
```diff
@@ -424,3 +425,24 @@ def _write_data(self, dataframe: dd.DataFrame, *, manifest: Manifest):
...
+    "DaskWriteComponent": DaskWriteExecutor,
+    "PandasTransformComponent": PandasTransformExecutor,
+}
+
+
+class ComponentRunner:
+    def __init__(self, component: t.Type[Component]):
+        self.component = component
+
+    def _get_executor(self) -> Executor:
```
mrchtr, line 442: If we go for the factory approach, this method would be the only one inside of this class.
Basically, something like this:

```
def create_executer(self, component) -> Executer:
        component_type = self.component.__bases__[0].__name__
        return COMPONENT_EXECUTOR_MAPPING[component_type].from_args()
```

src/fondant/executor.py
```diff
@@ -424,3 +425,24 @@ def _write_data(self, dataframe: dd.DataFrame, *, manifest: Manifest):
...
+
+COMPONENT_EXECUTOR_MAPPING: t.Dict[str, t.Type[Executor]] = {
+    "DaskLoadComponent": DaskLoadExecutor,
+    "DaskTransformComponent": DaskTransformExecutor,
+    "DaskWriteComponent": DaskWriteExecutor,
+    "PandasTransformComponent": PandasTransformExecutor,
+}
+
+
+class ComponentRunner:
```
mrchtr, line 438: I would rename it to `ExecutorFactory`. In this case, the `executor.py` is the correct location where I would expect such a class.

src/fondant/cli.py
```diff
@@ -327,10 +335,49 @@ def run(args):
...
+    parser.add_argument(
+        "ref",
+        help="""Reference to the module containing the component to run""",
+        action="store",
+    )
+
+    parser.set_defaults(func=execute)
+
+
+def execute(args):
```
mrchtr, line 367: Finally, we could use it here e.g.

```
def execute(args):
    component = component_from_module(args.ref)
    executer = ExecuterFactory(component)
    executer.execute(component)
```

mrchtr (commented):
> Yes, unless we detect it automatically by eg. finding a Component subclass in the file. This was just my original idea though, open to other suggestions.
> 
> CC @mrchtr I think you commented on my PR originally to hide the executor as well. Any opinion?

Thanks @PhilippeMoussalli! The current state is already looking quite good. I like the simplicity of executing the component using the fondant cli command.

However, my initial thought was basically that we have to make sure a user always initialise the right component subclass with the right executer subclass. 

Rather than utilising the ComponentRunner, I would be move this functionality to the `ComponentExecutor` itself. Cause a `Component` is already part of an `Executor` which is responsible for managing the execution of the logic. Therefore, I think we just need to provide a common way to initialise the component and executer. 

Initially, I was considering the implementation of a factory that accepts a `XComponent` and returns an `XExecutor` based on the components class. We could migrate the logic from the `ComponentRunner.run` to the BaseExecutor
(this should be consistent across all executors, and if we find the necessity to implement an approaches for a new runner, we can still override the method). The factory could be called in your added cli command. 

I think the strongest pro argument for this approach is that we wouldn't need the name `ComponentRunner` (which does not align with the other runners) anymore. Instead we could call it e.g. `ExecuterFactory` and keep it in the `executer.py`

PhilippeMoussalli:
> > Yes, unless we detect it automatically by eg. finding a Component subclass in the file. This was just my original idea though, open to other suggestions.
> > CC @mrchtr I think you commented on my PR originally to hide the executor as well. Any opinion?
> 
> Thanks @PhilippeMoussalli! The current state is already looking quite good. I like the simplicity of executing the component using the fondant cli command.
> 
> However, my initial thought was basically that we have to make sure a user always initialise the right component subclass with the right executer subclass.
> 
> Rather than utilising the ComponentRunner, I would be keep this functionality in the `ComponentExecutor` itself. Cause a `Component` is already needed for the execution in the `Executor`. Which does already the right job (managing the execution of the logic itself). Therefore, I think we just need to provide a common way to initialise the component and executer.
> 
> Initially, I was considering the implementation of a factory that accepts a `XComponent` and returns an `XExecutor` based on the components class. We could migrate the logic from the `ComponentRunner.run` to the BaseExecutor (this should be consistent across all executors, and if we find the necessity to implement an approaches for a new runner, we can still override the method). The factory could be called in your added cli command.
> 
> I think the strongest pro argument for this approach is that we wouldn't need the name `ComponentRunner` (which does not align with the other runners) anymore. Instead we could call it e.g. `ExecuterFactory` and keep it in the `executer.py`

Thanks for the suggestion @mrchtr! Indeed implementing it as a factory seems to be cleaner than having it as a `runner` which is a rather restricted to pipelines. I'v implemented the suggestions above

RobbeSneyders (approved):
Great to have this, nice job @PhilippeMoussalli!

PR #366, opened 2023-08-17
Expand cli to support kfp compiling and running

GeorgesLorre:


tests/test_cli.py
```diff
@@ -46,23 +46,32 @@ def test_pipeline_from_string_error(import_string):
         pipeline_from_string(import_string)
 
 
-def test_compile_logic(tmp_path_factory):
+def test_local_compile(tmp_path_factory):
     """Test that the compile command works with arguments."""
     with tmp_path_factory.mktemp("temp") as fn:
         args = argparse.Namespace(
             local=True,
+            kubeflow=False,
```
RobbeSneyders, line 54: It might make more sense to merge these options (especially if we have more in the future):

```python
engine = self.parser.add_mutually_exclusive_group()
engine.add_argument('--local',
                    help='Run locally',
                    dest='engine',
                    action='store_const',
                    const='local')
engine.add_argument('--kfp',
                    help='Run on kfp',
                    dest='engine',
                    action='store_const',
                    const='kfp')
```

It's then available as `args.engine` with a value of either `"local"` or `"kfp"`

RobbeSneyders (approved):
Thanks @GeorgesLorre!

Issue #368, opened 2023-08-17
Redesign base path file structure for caching and data exploration


PhilippeMoussalli:
The current structure of the base path is as follows:

```bash
base_path
├── subset_1
│   └── <pipeline_name>-<pipeline_id>
│       └── data.0.parquet
├── subset_2
│   └── <pipeline_name>-<pipeline_id>
│       └── data.0.parquet
├── component_1
│   └── manifest_<hash_key_1>.json
└── component_2
    └── manifest_<hash_key_2>.json

```

Note: 
- hash key is still not included in the main branch 
- Keep in mind that we now have a unified structure between the local and remote runner 

We need to have a required file structure that makes it easy to: 

1) Find all the corresponding components of a given `pipeline_name` and `pipeline id`
	* This is needed for the data explorer where the user might select a pipeline name -> pipeline_id (drop down based on pipeline name) -> component and gets a list of all possible subsets to explore given the component's manifest 
2) Final all the manifests for a certain component name (see #325)
	* This is needed for the caching mechanism where we check for a cached pipeline run within a certain component folder based on the hash key appended to the manifest name.

The existing file structure only enables us to accomplish (2). For the data explore, the user has to now manually give the manifest path which is not very intuitive. It can also get very cluttered with an increase number of subsets and components. 

In order to enables both, I would propose to move to the following file structure: 

```bash
base_path/
├── pipelines/
│   └── <pipeline_name>/
│       └── <pipeline_id>/
│           ├── component_1/
│           │   ├── manifest_<hash_key_1>.json
│           │   └── subsets/
│           │       ├── <subset_1>/
│           │       │   └── part.0.parquet
│           │       └── <subset_2>/
│           │           └── part.0.parquet
│           └── component_2/
│               └── manifest_<hash_key_2>.json
└── manifest_references/
    ├── component_1/
    │   └── references.json
    └── component_2/
        └── references.json


```

The `pipelines` directory enables (1) and the `manifest_references` directory enables (2). `references.json` is a file that contains references to all the manifests of a certain component starting from the base path  it can be updated. 

`manifest_references/component_1/references.json`
```json
{ "manifest_<hash_key_1>": "pipelines/pipeline_name/pipeline_id/component_1/manifest_<hash_key_1>.json", 
 "manifest_<hash_key_2>": "pipelines/pipeline_name/pipeline_id/component_2/manifest_<hash_key_2>.json"
}
```

[...]

PR #372, opened 2023-08-21
Create separate class for metadata

PhilippeMoussalli:
PR that creates a metadata class, this will make it easier to implement #368 (was originally part of #325 but decided to break it down to make it easier to review).

Few other notable changes: 

- Refactored the runner to accept the pipeline as an attribute since it's the runner's interface 
- The `run_id` between both runners has now an identical format (name_timestamp), we no longer need the uid of kfp since it's just used to store the native output artifacts 
- The `safe_component_name` has been moved from the local runner to the component spec to avoid having to plug it everywhere

### DIFF
--- a/src/fondant/compiler.py
+++ b/src/fondant/compiler.py
@@ -2,23 +2,30 @@ import datetime
 import json
 import logging
 import typing as t
-from abc import ABC, abstractmethod
+from abc import ABC
 from dataclasses import asdict, dataclass
 from pathlib import Path
 
 import yaml
 
+from fondant.manifest import Metadata
 from fondant.pipeline import Pipeline
 
 logger = logging.getLogger(__name__)
 
 
 class Compiler(ABC):
-    """Abstract base class for a compiler."""
+    """Abstract base class for a pipeline compiler."""
 
-    @abstractmethod
-    def compile(self, *args, **kwargs) -> None:
-        """Abstract method to invoke compilation."""
+    def __init__(
+        self,
+        pipeline: Pipeline,
+    ):
+        self.pipeline = pipeline
+
+    def get_run_id(self) -> str:
+        timestamp = datetime.datetime.now().strftime("%Y%m%d%H%M%S")
+        return f"{self.pipeline.name}-{timestamp}"
 
 
 @dataclass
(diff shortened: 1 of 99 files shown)


src/fondant/compiler.py
```diff
@@ -2,23 +2,30 @@
...
+    """Abstract base class for a pipeline compiler."""
 
-    @abstractmethod
-    def compile(self, *args, **kwargs) -> None:
-        """Abstract method to invoke compilation."""
+    def __init__(
+        self,
+        pipeline: Pipeline,
+    ):
+        self.pipeline = pipeline
```
RobbeSneyders, line 24: I originally proposed to move the `pipeline` argument to the `compile` method, and I still think that's more logical.

> Why is the pipeline passed as an initialization argument instead of to the `compile` method. From a logical perspective, I wouldn't expect a compiler instance to be specific to a single pipeline.



_Originally posted by @RobbeSneyders in https://github.com/ml6team/fondant/pull/194#discussion_r1227781344_

src/fondant/compiler.py
```diff
@@ -124,37 +107,40 @@ def _patch_path(base_path: str) -> t.Tuple[str, t.Optional[DockerVolume]]:
...
         build_args: t.List[str],
     ) -> dict:
         """Generate a docker-compose spec as a python dictionary,
         loops over the pipeline graph to create services and their dependencies.
         """
-        timestamp = datetime.datetime.now().strftime("%Y%m%d%H%M%S")
-        path, volume = self._patch_path(base_path=pipeline.base_path)
-        run_id = f"{pipeline.name}-{timestamp}"
-        metadata = MetaData(run_id=run_id, base_path=path)
+        datetime.datetime.now().strftime("%Y%m%d%H%M%S")
```
RobbeSneyders, line 116: Is this line doing anything?

PhilippeMoussalli: No, it should be removed

src/fondant/compiler.py
PhilippeMoussalli: oh I see, makes sense. I'll revert the changes

src/fondant/manifest.py
```diff
@@ -112,20 +133,31 @@ def retrieve_from_filesystem(uri: str) -> Resource:
...
+        *,
+        pipeline_name: str,
+        base_path: str,
+        run_id: str,
+        component_id: str,
+    ) -> "Manifest":
         """Create an empty manifest.
 
         Args:
+            pipeline_name: the bane of the pipeline
```
GeorgesLorre, line 147:
```suggestion
            pipeline_name: the name of the pipeline
```

GeorgesLorre (approved):
Nice change, makes sense

RobbeSneyders (approved):

PR #373, opened 2023-08-21
Redesign base path file structure

PhilippeMoussalli:
Addresses #368

First review this #372


RobbeSneyders:
Thanks @PhilippeMoussalli!

Is the manifest now also saved in the same directory as the data when running on Kubeflow? Or is it only written as a Kubeflow artifact?

PhilippeMoussalli:
> Thanks @PhilippeMoussalli!
> 
> Is the manifest now also saved in the same directory as the data when running on Kubeflow? Or is it only written as a Kubeflow artifact?

In both actually, It was done in #322 

The kubeflow artifact one (which can't be modified) is needed to connect components through artifacts and the data one (in the base path) is for us

src/fondant/executor.py
```diff
@@ -212,9 +212,9 @@ def upload_manifest(self, manifest: Manifest, save_path: t.Union[str, Path]):
 
         if is_kubeflow_output:
```
RobbeSneyders, line 213: I'm not the biggest fan of having kubeflow specific code in here. Opened #376 to remove this.

RobbeSneyders (commented):
Ok thanks, that clears things up.

src/fondant/executor.py
```diff
@@ -212,9 +212,9 @@ def upload_manifest(self, manifest: Manifest, save_path: t.Union[str, Path]):
...
             # Save to the expected base path directory
-            safe_component_name = self.spec.name.replace(" ", "_").lower()
             save_path_base_path = (
-                f"{manifest.base_path}/{safe_component_name}/manifest.json"
+                f"{manifest.pipeline_name}/{manifest.run_id}/"
+                f"{manifest.component_id}/manifest.json"
```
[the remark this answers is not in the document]
PhilippeMoussalli: Good catch! I confused it with a subset where the `base_path` is appended to the location

src/fondant/executor.py
PhilippeMoussalli: I agree, though I don't see a clear-cut solution for this just yet

GeorgesLorre: will vertex have the same issue ?

PhilippeMoussalli: probably, ideally we have some control over the format and save path of the artifact.

RobbeSneyders (approved):
Thanks @PhilippeMoussalli!

PR #374, opened 2023-08-22
[DataComp] Add T-MARS

NielsRogge:
This PR adds the [T-MARS](https://tmars-clip.github.io/) pipeline to the Datacomp folder. Initial discussion at https://github.com/locuslab/T-MARS/issues/3.

Screenshot 2023-08-31 at 11 11 19

To do:
- remove commit hashes and replace with dev tag before merging


RobbeSneyders:
Do all the model and config files need to be included in the `detect_text` component?

examples/pipelines/datacomp/components/detect_text/src/main.py
```diff
@@ -0,0 +1,194 @@
...
+import logging
+import typing as t
+
+import numpy as np
+import pandas as pd
+import torch
+from fondant.component import PandasTransformComponent
+from PIL import Image
+
+logger = logging.getLogger(__name__)
```
PhilippeMoussalli, line 15: Can you move this after the imports and group all imports together?

examples/pipelines/datacomp/components/detect_text/src/main.py
```diff
@@ -0,0 +1,194 @@
...
+         polygon (ArrayLike): A polygon. In any form can be converted
+             to an 1-D numpy array. E.g. list[float], np.ndarray,
+             or torch.Tensor. Polygon is written in
+             [x1, y1, x2, y2, ...].
+
+     Returns:
+         np.array: The converted bounding box [x1, y1, x2, y2]
+    """
+    if len(polygon) % 2 != 0:
+        raise ValueError("Not ok")
```
PhilippeMoussalli, line 44: Perhaps a more descriptive error?

scripts/build_components.sh
```diff
@@ -97,6 +97,8 @@ for dir in "${components_to_build[@]}"; do
   docker build --push "${args[@]}" \
    --build-arg="FONDANT_VERSION=${tags[0]}" \
    --label org.opencontainers.image.source=https://github.com/${namespace}/${repo} \
+   --progress=plain \
+   --no-cache \
```
PhilippeMoussalli, line 101: This should definitely not be there. We want to keep caching when building components

scripts/build_components.sh
```diff
@@ -97,6 +97,8 @@ for dir in "${components_to_build[@]}"; do
   docker build --push "${args[@]}" \
    --build-arg="FONDANT_VERSION=${tags[0]}" \
    --label org.opencontainers.image.source=https://github.com/${namespace}/${repo} \
+   --progress=plain \
```
PhilippeMoussalli, line 100: what does this do?

examples/pipelines/datacomp/components/detect_text/src/main.py
```diff
@@ -0,0 +1,194 @@
...
+
+def detect_text_batch(
+    image_batch: pd.DataFrame,
+    *,
+    cfg,
+    model,
+) -> pd.Series:
+    """Detext text on a batch of images."""
+    imgs = torch.stack(list(image_batch), dim=0)
+    batch_size = imgs.shape[0]
```
PhilippeMoussalli, line 81: I don't think you need to infer this, you can just use `self.batch_size`

examples/pipelines/datacomp/components/detect_text/src/main.py
```diff
@@ -0,0 +1,194 @@
...
+    img_metas = {
+        "filename": [None for i in range(batch_size)],
+        "org_img_size": torch.ones((batch_size, 2)).long() * 256,
+        "img_size": torch.ones((batch_size, 2)).long() * 512,
+    }
+
+    data = dict()
+    data["imgs"] = imgs
+    data["img_metas"] = img_metas
+    data.update(dict(cfg=cfg))
```
PhilippeMoussalli, line 93:
```suggestion
data = {
    "imgs": imgs,
    "img_metas": img_metas,
    "cfg": cfg,
}
```

examples/pipelines/datacomp/components/detect_text/src/main.py
```diff
@@ -0,0 +1,194 @@
...
+
+    # forward
+    with torch.no_grad():
+        outputs = model(**data)
+
+    # get cropped images
+    boxes_batch = []
```
PhilippeMoussalli, line 100: Can you split `detect_text_image` into two functions similar to  `process_image`, one for detection and other for cropping? Helps in readability

examples/pipelines/datacomp/components/detect_text/src/main.py
```diff
@@ -0,0 +1,194 @@
...
+        logger.info(f"Device: {self.device}")
+
+        self.batch_size = batch_size
+        self.image_transform = SquarePadResizeNorm(
+            img_size=512,
+            norm_mean=(0.485, 0.456, 0.406),
+            norm_std=(0.229, 0.224, 0.225),
+        )
+
+        cfg = Config.fromfile("config/fast/tt/fast_tiny_tt_512_finetune_ic17mlt.py")
```
PhilippeMoussalli, line 127: I would remove all the other configs for the other models. It's a lot of files to upload. We can maybe think about developing the model later to be more generic then you should include the model name as a component argument

examples/pipelines/datacomp/components/detect_text/src/main.py
PhilippeMoussalli, line 1: Don't forget to remove commented out script

examples/pipelines/datacomp/components/detect_text/src/main.py
```diff
@@ -0,0 +1,194 @@
...
+    """Component that detects text in images using an mmocr model."""
+
+    def __init__(self, *args, batch_size: int) -> None:
+        self.device = "cuda" if torch.cuda.is_available() else "cpu"
+        logger.info(f"Device: {self.device}")
+
+        self.batch_size = batch_size
+        self.image_transform = SquarePadResizeNorm(
+            img_size=512,
+            norm_mean=(0.485, 0.456, 0.406),
```
PhilippeMoussalli, line 123: Are those normalization values the same across all models?

examples/pipelines/datacomp/components/detect_text/src/models/__init__.py
```diff
@@ -0,0 +1,9 @@
+# from .pan import PAN
+from .fast import FAST
+
+# from .psenet import PSENet
+from .builder import build_model
+
+# __all__ = ['PAN', 'PSENet', 'FAST']
```
PhilippeMoussalli, line 7: Can we remove this if we're only using a `FAST` model?

examples/pipelines/datacomp/components/detect_text/src/models/head/pa_head.py
PhilippeMoussalli, line 1: Would also remove those if we're only going to stick with a FAST model

examples/pipelines/datacomp/components/detect_text/src/models/head/psenet_head.py
PhilippeMoussalli, line 1: Same as above

examples/pipelines/datacomp/components/detect_text/src/models/loss/dice_loss.py
PhilippeMoussalli, line 1: Do we need the `loss` folder if we're not training?

examples/pipelines/datacomp/components/detect_text/src/models/neck/fpn.py
PhilippeMoussalli, line 1: same here

examples/pipelines/datacomp/components/load_from_hf_hub/fondant_component.yaml
```diff
@@ -32,6 +32,13 @@ produces:
       clipl14score:
         type: float32
 
+  textembedding:
```
PhilippeMoussalli, line 35: Wondering whether it won't be more intuitive to have the embeddings as a field rather than a subset. It makes more sense since different modalities (subsets) can have different embeddings, Wdyt?

PhilippeMoussalli (commented):
Thanks @NielsRogge! 

Looks good overall but I think there are many added files for the text detection component that can be removed (will make review easier). Maybe let's stick to one model for now and only keeps the configs necessary for that one

scripts/build_components.sh
NielsRogge: This was mainly for debugging purposes.

examples/pipelines/datacomp/components/load_from_hf_hub/fondant_component.yaml
NielsRogge: Yes I know, however image data and image embeddings are both heavy in terms of data, which is probably why the `subsets` were created in the first place. So if both image data and embeddings are part of the same subset, then Fondant will first always load both first, and then only provide the embeddings if the user specified this, right?

cc @RobbeSneyders

examples/pipelines/datacomp/pipeline.py
```diff
@@ -27,50 +30,83 @@
...
+)
+detect_text_op = ComponentOp(
+    component_dir="components/detect_text",
+    arguments={
+        "batch_size": 2,
+    },
+    node_pool_label="node_pool",
+    node_pool_name="model-inference-mega-pool",
+    number_of_gpus=1,
+    cache=False,
```
PhilippeMoussalli, line 70: You can remove `cache=False` now

scripts/build_components.sh
PhilippeMoussalli: This still needs to be addressed

examples/pipelines/datacomp/pipeline.py
```diff
@@ -16,8 +19,8 @@
     pipeline_name="datacomp-filtering-pipeline",
     pipeline_description="A pipeline for filtering the Datacomp dataset",
     base_path=PipelineConfigs.BASE_PATH,
+    # base_path="/Users/nielsrogge/Documents/fondant_artifacts_datacomp",
```
PhilippeMoussalli, line 22: This can be removed

examples/pipelines/datacomp/components/detect_text/src/main.py
PhilippeMoussalli: can you clarify this?

PhilippeMoussalli (commented):
Thanks Niels! 

Left a few additional comments. Can you please revert back the changes to the Dockerfiles/images? Otherwise the CI/CD pipelines will fail to update them

PhilippeMoussalli (approved):

PR #378, opened 2023-08-23
Modify kfp command

PhilippeMoussalli:
PR that modifies the KFP command to contain the new fondant cli command used to execute the component


NielsRogge (approved):
Thanks for fixing!

PR #380, opened 2023-08-23
Bugfix partitions Load from Hub

PhilippeMoussalli:
A fix for loading the dataset when `n_rows_to_load` is specified. An offset of 1 was missing which caused an issue when there was only 1 partition (npartitions from the enumerate started from 0 and an attempt to call `df.head()` with 0 partitions was made)


RobbeSneyders (approved):
Thanks @PhilippeMoussalli 

FYI, you could have used `enumerate(dask_df.partitions, start=1)` as well :slightly_smiling_face:

PhilippeMoussalli:
> Thanks @PhilippeMoussalli
> 
> FYI, you could have used `enumerate(dask_df.partitions, start=1)` as well 🙂

Oh that's nicer indeed, i'll add it

PR #382, opened 2023-08-23
Bugfix data-explorer images

PhilippeMoussalli:
PR that fixes the issue with rendering images in the explorer. Images currently have type `binary` and not `object`. This was changed to render them properly. 

Caveat: not all `binary` types will correspond to images so we will need to implement another logic for image detection but this fix should do for now. 

Fixed another small bug related to errors in rendering the numerical explorer when no numeric columns are present

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -74,6 +74,7 @@ wheels/
 *.egg-info/
 .installed.cfg
 *.egg
+.tox
 MANIFEST
 
 # DS_Store (MacOS)
@@ -85,8 +86,12 @@ MANIFEST
 # kubeflow artifacts
 *.tgz
 
+# Coverage files
+*.coverage
+
 # docker artifacts
 examples/pipelines/**/docker-compose.yml
+examples/pipelines/**/pipeline.yaml
 
 # yaml-e files
 *.yaml-e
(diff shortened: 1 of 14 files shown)


ChristiaensBert (approved):
lgtm

PhilippeMoussalli:
> Looks good, are the images rendered decently fast? When did we make the change between binary and object?

Only tested for 20 images but it was ok. Are there methods to optimize this? 

Not sure if we every made the switch between the two types (might have happened a while ago). In general, this is the schema that is checked with the manifest: 

https://github.com/ml6team/fondant/blob/50f3a97878ac81670ebe624039ff0fcec0542e4f/src/fondant/schema.py#L41C1-L42C1

PhilippeMoussalli:
> I'm still getting a lot of issues just running `fondant explore` on this branch. Do I need to rebuild an image and update the version to get the latest code?


I am using the fondant explore with the `-tag` flag with this commit hash `71214bb9de30cd4302d09af97acf117d4b0e9231` (latest from this PR)

It does require first building it with `scripts/build_explorer.sh --tag 71214bb9de30cd4302d09af97acf117d4b0e9231`

I assume this won't be an issue when it's merged to main (if it's part of the CI/CD)

PR #383, opened 2023-08-23
Strip url in download_images before downloading

RobbeSneyders:
This should solve a lot of the "invalid non-printable ASCII character in URL" errors


NielsRogge (approved):
Thanks!

PR #384, opened 2023-08-24
Generic common crawl download component

mrchtr:
This PR adds a common crawl download component which download specific webpages after filtering the common crawl columnar index.


components/common_crawl_download/tests/cc_download_test.py
```diff
@@ -0,0 +1,20 @@
+from main import CommonCrawlDownloadComponent
+
+
+def test_load_warc_file_content():
+    """Test components transform method."""
+    # TODO implement a real test case
```
mrchtr, line 6: We still have to fix the test case before merging.

mrchtr, line 6: We still have to fix the test before merging to main.

components/common_crawl_download/src/main.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def load(self) -> dd.DataFrame:
+        """
+        Returns:
+            Dask dataframe.
+        """
+        ddf = self.filter_common_crawl_index()
+
+        # Repartitioning to utilize all cores
+        ddf = ddf.repartition(npartitions=os.cpu_count())
```
mrchtr, line 80: Not entirely sure if the repartitioning is needed here.

components/common_crawl_download/src/main.py
```diff
@@ -0,0 +1,108 @@
...
+        ddf = ddf[['url', 'content']]
+
+        ddf.columns = [
+            "webpage_url",
+            "webpage_content",
+        ]
+
+        if self.n_records_to_download is not None:
+            ddf_n_partitions = ddf.partitions
+            ddf = ddf.head(self.n_records_to_download)
```
mrchtr, line 97: We should consider a different approach for limiting the number of records. Actually this triggers a compute and still a whole download of the defined webpages.

mrchtr (commented):
I have merged the source code from both @shayorshay and myself, and tested the component locally. I have documented some questions that arose during this process.
Would be nice to get your opinion on these @RobbeSneyders

components/common_crawl_download/src/main.py
```diff
@@ -0,0 +1,161 @@
...
+    def load(self) -> dd.DataFrame:
+        index_ddf = self.load_index()
+
+        if self.n_records_to_download is not None:
+            n_partitions = max(os.cpu_count(), self.n_records_to_download // 10000)  # type: ignore
+            logging.info(f"Repartitioning to {n_partitions} partitions.")
+            index_ddf = dd.from_pandas(index_ddf.head(self.n_records_to_download,
+                                                      npartitions=-1), npartitions=n_partitions)
+        else:
+            n_partitions = len(self.index_files) * 1000
```
mrchtr, line 137: Are there specific reasons why you used `1000` and `10000`?

components/common_crawl_download/src/main.py
```diff
@@ -0,0 +1,161 @@
...
+        return content_df
+
+    def load(self) -> dd.DataFrame:
+        index_ddf = self.load_index()
+
+        if self.n_records_to_download is not None:
+            n_partitions = max(os.cpu_count(), self.n_records_to_download // 10000)  # type: ignore
+            logging.info(f"Repartitioning to {n_partitions} partitions.")
+            index_ddf = dd.from_pandas(index_ddf.head(self.n_records_to_download,
+                                                      npartitions=-1), npartitions=n_partitions)
```
mrchtr, line 135: `npartitions=-1` Would that limit n_records_to_download to the size of the last partition?

mrchtr (commented):
Thanks @RobbeSneyders. Two small remaining questions about the initial partitioning:

components/common_crawl_download/src/main.py
RobbeSneyders: There's around 10M pages in each index file. So this should end up with ~10k pages per partition in both cases (depending on filtering of course).

Each page in the output of my test was ~25KB (text extracted), so 10k pages should be around the Dask recommendation of 250MB per partition.

Will add a comment.

components/common_crawl_download/src/main.py
RobbeSneyders: [It uses all partitions](https://docs.dask.org/en/latest/generated/dask.dataframe.DataFrame.head.html).

Have to admit that I don't know what the impact on performance is.

RobbeSneyders (approved):

mrchtr:
@RobbeSneyders shall we close this one? I think the dask writing issue is a blocker to continue here.

RobbeSneyders:
Yes we can close for now.

PR #386, opened 2023-08-24
Set image pull policy to always

PhilippeMoussalli:
Addresses #365

### DIFF
--- a/src/fondant/compiler.py
+++ b/src/fondant/compiler.py
@@ -279,6 +279,10 @@ class KubeFlowCompiler(Compiler):
                     component_task,
                     component_op,
                 )
+
+                # Set image pull policy to always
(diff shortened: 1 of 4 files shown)


RobbeSneyders (approved):
Thanks @PhilippeMoussalli!

We might want to add this for the local runner as well.

PhilippeMoussalli:
> Thanks @PhilippeMoussalli!
> 
> We might want to add this for the local runner as well.

Isn't it already the case? 
https://github.com/ml6team/fondant/commit/c1806869180cecf10311eaaeb1b84831608c6bb3

RobbeSneyders:
Apparently it is. Now @GeorgesLorre again won't believe that I write my own code. :sweat_smile:

PR #387, opened 2023-08-24
Implement caching

PhilippeMoussalli:
PR that implements caching: 

* The component cache is estimated at compile time based on the images tag, input parameters, component spec and component resources (GPU, nodepool). Caching is enabled by default at the componentOp level and disabled if the image has the `latest` tag

* At runtime, the cache key is checked against the base path to check for matching executions based on a the cache key (currently appended to the manifest name). 

*  If multiple matching executions are found, we return the latest one. This makes it easier to develop. For example, if we're working with the `dev` tag and we disable caching to force the execution of a new version of the component (scenario where the spec and parameters are the same). Then when we enable caching again, we would only load the manifest from the latest execution. 
 
* A component is cached only if it's enabled at the Op level, a matching manfiest is found and previous component's run is cached. To check whether the previous component is cached or not. We check if the run-id of the previous component (input manifest) has the same run-id as the current component. If they match, it means that the previous component had to execute (manifest not loaded from previous pipeline run). Note that this implies that we don't change the loaded manifest's run-id to that of the current pipeline run. This might not be ideal for lineage tracking but I see no other way of doing this check.

### DIFF
--- a/src/fondant/executor.py
+++ b/src/fondant/executor.py
@@ -262,6 +262,22 @@ class Executor(t.Generic[Component]):
         )
         return True
 
+    def _run_execution(
+        self,
+        component_cls: t.Type[Component],
+        input_manifest: Manifest,
+    ) -> Manifest:
+        logging.info("Executing component")
+        component = component_cls(self.spec, **self.user_arguments)
+        output_df = self._execute_component(
+            component,
+            manifest=input_manifest,
+        )
+        output_manifest = input_manifest.evolve(component_spec=self.spec)
+        self._write_data(dataframe=output_df, manifest=output_manifest)
+
+        return output_manifest
+
     def execute(self, component_cls: t.Type[Component]) -> None:
         """
         Execute a component.
@@ -269,19 +285,6 @@ class Executor(t.Generic[Component]):
         Args:
             component_cls: The class of the component to execute
         """
-
-        def _run_execution(inner_input_manifest: Manifest) -> Manifest:
-            logging.info("Executing component")
-            component = component_cls(self.spec, **self.user_arguments)
-            output_df = self._execute_component(
-                component,
-                manifest=inner_input_manifest,
-            )
-            inner_output_manifest = input_manifest.evolve(component_spec=self.spec)
-            self._write_data(dataframe=output_df, manifest=inner_output_manifest)
-
-            return inner_output_manifest
-
         input_manifest = self._load_or_create_manifest()
 
         if self.cache and self._is_previous_cached(input_manifest):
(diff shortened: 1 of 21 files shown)


RobbeSneyders:
Thanks @PhilippeMoussalli 

Still have to look at the code, but already two questions:

- > A component is cached only if it's enabled at the Op level, a matching manfiest is found and previous component's run is **not** cached.

  I assume this "not" is a mistake, and we only cache if the previous component was also cached?

- Can you add some documentation on the caching as well?

PhilippeMoussalli:
> Thanks @PhilippeMoussalli
> 
> Still have to look at the code, but already two questions:
> 
> * > A component is cached only if it's enabled at the Op level, a matching manfiest is found and previous component's run is **not** cached.
>   
>   
>   I assume this "not" is a mistake, and we only cache if the previous component was also cached?
> * Can you add some documentation on the caching as well?

Yeah indeed, was a mistake. 

I will add documentation

src/fondant/filesystem.py
```diff
@@ -0,0 +1,40 @@
...
+import fsspec
+
+logger = logging.getLogger(__name__)
+
+FSSPEC_SCHEME_DICT = {
+    "file": "file",
+    "s3": "s3",
+    "gs": "gcs",
+    "abfs": "abfs",
+}
```
RobbeSneyders, line 13: I don't think we need this mapping. `fsspec.filesystem()` works with the protocol directly.

src/fondant/executor.py
```diff
@@ -189,43 +191,109 @@ def _write_data(self, dataframe: dd.DataFrame, *, manifest: Manifest):
...
-        else:
-            logger.info("No previous matching execution found")
+        def _run_execution(inner_input_manifest: Manifest) -> Manifest:
+            logging.info("Executing component")
+            component = component_cls(self.spec, **self.user_arguments)
+            output_df = self._execute_component(
+                component,
+                manifest=inner_input_manifest,
+            )
+            inner_output_manifest = input_manifest.evolve(component_spec=self.spec)
```
RobbeSneyders, line 280:
```suggestion
            inner_output_manifest = inner_input_manifest.evolve(component_spec=self.spec)
```

Maybe it's better to move this into a separate method to prevent these kind of issues.

RobbeSneyders (commented):
Thanks @PhilippeMoussalli. Looks good to me apart from some small comments.

src/fondant/filesystem.py
PhilippeMoussalli: you're right, adjusted it

RobbeSneyders (approved):

RobbeSneyders:
Thanks @PhilippeMoussalli. Enormous effort which improves Fondant a lot!

Issue #244, opened 2023-06-29
Pass unconsumed columns when additionalFields is True


RobbeSneyders:
[As described in our documentation](https://fondant.readthedocs.io/en/latest/component_spec/#additionalsubsets-additionalfields), fondant should propagate subset columns when `additionalFields` is True. This is currently not the observed behavior, and the user needs to consume and produce each column they want in the output.

To fix this, we should:
- Read all the consumed and propagated columns
- Pass the consumed columns to the user
- Merge the data returned by the user with the propagated columns
- Write the results

RobbeSneyders:
We should think about how we want to tackle this, since this would remove the benefit of only loading the columns that the user specifies.

RobbeSneyders:
Achieved by redesign in #672

PR #390, opened 2023-08-28
Add url deduplication for license free image use case

mrchtr:
Add component which deduplicates image urls by reset the index to the image url column.


examples/pipelines/commoncrawl/components/image_url_deduplication/fondant_component.yaml
```diff
@@ -0,0 +1,9 @@
+name: Deduplicate images based on image source url
+description: Component that deduplicate images based on source url
+image: ghcr.io/ml6team/image_url_deduplication:latest
+
+consumes:
```
RobbeSneyders, line 5: Since #244 is still open, we should consume and produce all the fields that we need

examples/pipelines/commoncrawl/components/image_url_deduplication/src/main.py
```diff
@@ -0,0 +1,37 @@
...
+logger = logging.getLogger(__name__)
+
+
+def deduplicate_dask_dataframe(dataframe: dd.DataFrame, column: str):
+    logger.info(f"Start exact deduplication of column {column}.")
+
+    num_partitions = dataframe.npartitions
+
+    # Sort the DataFrame on the column
+    dataframe = dataframe.sort_values(column)
```
RobbeSneyders, line 16: I would calculate a `url_surtkey` for each row, and then set it as index, which will sort on it, before deduplicating.

examples/pipelines/commoncrawl/components/image_url_deduplication/src/main.py
```diff
@@ -0,0 +1,37 @@
...
+
+    # Set index to column
+    # dataframe = dataframe.set_index(column)
+
+    # Repartition on column
+    dataframe = dataframe.repartition(npartitions=num_partitions)
+
+    # Dedup partition wise on exact match
+    return dataframe.drop_duplicates(subset=[column], keep="last")
```
RobbeSneyders, line 25: Does this leverage the fact that the data is sorted? I think this is a global operation, while we sort first to make it a local operation.

RobbeSneyders (commented):
Thanks @mrchtr.

Some comments. I believe we should calculate a `url_surtkey`, sort on it, and then do local deduplication.

examples/pipelines/commoncrawl/components/image_url_deduplication/src/main.py
RobbeSneyders: We also might want to implement the local operation ourselves so we can choose a strategy to keep a certain license. I see two options:

- Iterate through the rows while comparing each `url_surtkey` with the previous one, and then checking the license to decide which one to keep.
- Sort each partition on a second column as well, which should sort the licenses on strictness, so we can use `pandas.drop_duplicates` with the `keep` argument.

examples/pipelines/commoncrawl/components/image_url_deduplication/fondant_component.yaml
RobbeSneyders: Nvm, this is only filtering.

examples/pipelines/commoncrawl/components/image_url_deduplication/src/main.py
mrchtr: I think we can't set the `url_surtkey` as index. Fondant expects as index column the column `id`, if I see it correctly:
`dataframe.index = dataframe.index.rename("id")`.

examples/pipelines/commoncrawl/components/image_url_deduplication/src/main.py
mrchtr: Good point. I have implemented the second approach.

examples/pipelines/commoncrawl/components/image_url_deduplication/fondant_component.yaml
shayorshay, line 1: we need the `license_type` in the subset too

examples/pipelines/commoncrawl/components/image_url_deduplication/src/main.py
```diff
@@ -0,0 +1,49 @@
...
+            df: A dask dataframe with the image urls
+        Returns:
+            A dask dataframe with the image urls
+        """
+        num_partitions = dataframe.npartitions
+
+        # Sort the DataFrame on the column
+        dataframe = dataframe.sort_values(
+            by=["image_image_url", "image_license_type"],
+            key=lambda col: col.apply(get_license_strictness),
```
RobbeSneyders, line 43: Doesn't this sorting key get executed on the image_url as well?

examples/pipelines/commoncrawl/components/image_url_deduplication/tests/image_url_dedup_test.py
RobbeSneyders, line 1: Can we create the data in Python instead of using a parquet file? That would be more transparent.

examples/pipelines/commoncrawl/components/image_url_deduplication/src/main.py
RobbeSneyders: This just renames the index, it says nothing about its values. I still think a `url_surtkey` would be the most logical index.

examples/pipelines/commoncrawl/components/image_url_deduplication/src/main.py
```diff
@@ -0,0 +1,49 @@
...
+        )
+
+        # Repartition on column
+        dataframe = dataframe.repartition(npartitions=num_partitions)
```
RobbeSneyders, line 47: I don't think this ensures that duplicates are in the same partition as long as we don't have the url as the index.

RobbeSneyders (commented):
I still think this has some issues which I explain in my comments.

I would go for the following approach:
- Create `url_surtkey` column
- Set `url_surtkey` column as index (will sort on index)
- `map_partitions` to start working locally
  - pandas sort_values on index and license, taking care that sort key is applied to both columns
  - pandas drop duplicates

The `url_surtkey` is also the most logical index for the final dataset.

RobbeSneyders:
Closing for now.

PR #391, opened 2023-08-29
Preserve divisions when writing and reading

RobbeSneyders:
Apparently we are currently not preserving the divisions of the data when writing and reading to parquet. We are preserving the partitions when reading initially, but since divisions are no longer known, the partitions will be shuffled when we merge the index and subsets together.

Let's create a dataframe with known partitions and write it to parquet:
```python
ddf = dd.from_dict({"data": list(range(5))}, npartitions=5)
ddf.known_divisions
Out: True
ddf.divisions
Out: (0, 1, 2, 3, 4, 4)

ddf.to_parquet("/tmp/data")
```

If we read it back, we see that the partitions are preserved, but the division information is not.
```python
index_df = dd.read_parquet("/tmp/data")
index_df = index_df.drop("data", axis=1)

index_df.map_partitions(len).compute()
Out: 
0    1
1    1
2    1
3    1
4    1
dtype: int64

index_df.known_divisions
Out: False
index_df.divisions
Out: (None, None, None, None, None, None)
```

We read it back as a subset as well and double check that we get the same behavior, we do:
```python
subset_df = dd.read_parquet("/tmp/data")

subset_df.map_partitions(len).compute()
Out: 
0    1
1    1
2    1
3    1
4    1
dtype: int64

subset_df.known_divisions
Out: False
subset_df.divisions
Out: (None, None, None, None, None, None)
```

If we now merge these 2 dataframes together, we see that the partitions are shuffled, because Dask thinks the divisions are unknown.
```python
dataframe = dd.merge(index_df, subset_df, left_index=True, right_index=True, how="left")
dataframe.map_partitions(len).compute()
Out: 
0    3
1    1
2    1
3    0
4    0
dtype: int64
```

Running this again with the changes in this PR will preserve the division information and the partitions after merging.

### DIFF
--- a/src/fondant/data_io.py
+++ b/src/fondant/data_io.py
@@ -93,7 +93,11 @@ class DaskDataLoader(DataIO):
 
         logger.info(f"Loading subset {subset_name} with fields {fields}...")
 
-        subset_df = dd.read_parquet(remote_path, columns=fields)
+        subset_df = dd.read_parquet(
+            remote_path,
+            columns=fields,
+            calculate_divisions=True,
+        )
 
         # add subset prefix to columns
         subset_df = subset_df.rename(
@@ -115,7 +119,7 @@ class DaskDataLoader(DataIO):
         remote_path = index.location
 
         # load index from parquet, expecting id and source columns
-        return dd.read_parquet(remote_path)
+        return dd.read_parquet(remote_path, calculate_divisions=True)
 
     def load_dataframe(self) -> dd.DataFrame:
         """
@@ -137,7 +141,7 @@ class DaskDataLoader(DataIO):
                 subset_df,
                 left_index=True,
                 right_index=True,
-                how="inner",
+                how="left",
             )
 
         dataframe = self.partition_loaded_dataframe(dataframe)
@@ -159,7 +163,7 @@ class DaskDataWriter(DataIO):
     def write_dataframe(self, dataframe: dd.DataFrame) -> None:
         write_tasks = []
 
-        dataframe.index = dataframe.index.rename("id").astype("string")
+        dataframe.index = dataframe.index.rename("id")
 
         # Turn index into an empty dataframe so we can write it
         index_df = dataframe.index.to_frame().drop(columns=["id"])
@@ -257,6 +261,7 @@ class DaskDataWriter(DataIO):
             schema=schema,
             overwrite=False,
             compute=False,
+            write_metadata_file=True,
         )
         logging.info(f"Creating write task for: {location}")
         return write_task


src/fondant/data_io.py
```diff
@@ -137,7 +141,7 @@ def load_dataframe(self) -> dd.DataFrame:
                 subset_df,
                 left_index=True,
                 right_index=True,
-                how="inner",
+                how="left",
```
RobbeSneyders, line 144: Don't know why this was "inner" before. "left" is what we want to do (filter on the index).

src/fondant/data_io.py
```diff
@@ -159,7 +163,7 @@ def __init__(
     def write_dataframe(self, dataframe: dd.DataFrame) -> None:
         write_tasks = []
 
-        dataframe.index = dataframe.index.rename("id").astype("string")
+        dataframe.index = dataframe.index.rename("id")
```
RobbeSneyders, line 166: An integer columns is no longer sorted if we convert it to string. We should keep the index sorted.

src/fondant/data_io.py
PhilippeMoussalli: The reason for that was due to an error that Niels was facing which got fixed when changing the join type. I'll link you to the chat. 

I am not a DE expert but I read that inner joins should generally be faster and since an inner join only joins matching ids it should also filter on the index.

src/fondant/data_io.py
RobbeSneyders: No but I don't think there's a reason for us to cast it? We can just keep the type.

src/fondant/data_io.py
RobbeSneyders: I think left join is what we want functionally. The index keeps track of which records are in the dataset, so we should use its ids. If data for a certain id is missing in a subset, that will become None instead, which I think is quite logical.

I think the issue Niels had before might be solved by the changes in this PR as well, let's try.

As far as I know, inner joins are only more performant because of their different functionality. If all ids are present in the right index (which should usually be the case), I believe there is no difference. If ids are missing in the right index, we're not getting the functionality we want.

src/fondant/data_io.py
PhilippeMoussalli: Makes sense, I thought for some reason that we had a fixed schema for the index at the ComponentSpec level but that's not the case

src/fondant/data_io.py
PhilippeMoussalli: > I If data for a certain id is missing in a subset, that will become None instead, which I think is quite logical.

Won't this require dropping all the Nons as well before passing the dataframe to avoid having to handle cases where the row does not contain the expected data? Otherwise the user transform function will have to handle this.

RobbeSneyders: I think it's better not to drop it, because it would be silent and hard to debug. Instead it now fails loudly, or the user code in the component needs to handle Nones.

PhilippeMoussalli (approved):

PR #392, opened 2023-08-29
Add commoncrawl pipeline that starts from warc paths

RobbeSneyders:
This PR adds a commoncrawl pipeline which contains 2 components:
- A component that downloads `warc.paths.gz` files and outputs all the paths as separate partitions
- A component that downloads a single warc file and extracts the images and its licenses


mrchtr (approved):
Thanks @RobbeSneyders! Looks good to me.

PR #397, opened 2023-08-30
Standarize fsspec file access

PhilippeMoussalli:
PR that standarizes ffspec file access to use the `Abstractfilesystem` class for all filesystem related functionalities. 

Previously we were using the `Abstractfilesystem` and custom functions `fs_open()` to read/write/list the manifest(s). This caused some issue since `fs_open` expects the protocol of the file system to be present in the URL. However, files returns by `Abstractfilesystem().ls` returned a normal file system url. When passed to `fs_open()` it causes it to fail to read the file properly.

Still some issues with using glob patterns. More details here: https://github.com/ml6team/fondant/issues/368

### DIFF
--- a/src/fondant/executor.py
+++ b/src/fondant/executor.py
@@ -224,7 +224,7 @@ class Executor(t.Generic[Component]):
             else:
                 manifest_file = matching_manifests[0]
 
-            return Manifest.from_file(manifest_file)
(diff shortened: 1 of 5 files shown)


RobbeSneyders (approved):
Thanks @PhilippeMoussalli!

Have you been able to test this with remote storage?

Issue #375, opened 2023-08-22
Automatically detect pipeline attribute when running / compiling


RobbeSneyders:
In #362, we introduced a `component_from_module` function to automatically detect a Fondant component in a Python file. We can do the same to automatically detect a Fondant pipeline in a Python file.

This way users can run:
```
fondant run pipeline.py
```

instead of:
```
fondant run pipeline:pipeline
```

PR #398, opened 2023-08-30
Detect pipeline attribute during compile/run

PhilippeMoussalli:
Addresses #375


GeorgesLorre:
Nice change ! Makes it easier the import string was indeed a bit confusing.

Some notes:
1. did you check the documentation for cli examples that might need updating ?
2. In the original implementation when there was an error in you pipeline code the cli would just say that it could not find your pipeline but there was never any indication (traceback or special message) that there was something wrong is this easier in this implementation ?

PhilippeMoussalli:
> Nice change ! Makes it easier the import string was indeed a bit confusing.
> 
> Some notes:
> 
> 1. did you check the documentation for cli examples that might need updating ?
> 2. In the original implementation when there was an error in you pipeline code the cli would just say that it could not find your pipeline. There was never any indication (traceback or special message) that there was something wrong, is this easier in this implementation ?

Thanks Georges! 

1) Just update the docs. 
2) If there is no pipeline found it will return a valid error. I think the error that you're referring to was because of the default output path (which was wrongly set to None). This shouldn't occur anymore. It's either the default for kfp/docker-compose or a user custom one.

GeorgesLorre (approved):

docs/getting_started.md
```diff
@@ -287,7 +287,7 @@ We add the component to our pipeline definition and specify that it depends on t
 We can now easily run our new pipeline:
 
 ```bash
-fondant run pipeline:my_pipeline --local
+fondant run pipeline --local
```
mrchtr, line 290: I think one command is outdated here. Either the one in line 142 `fondant run pipeline.py --local` or this one.

docs/getting_started.md
```diff
@@ -139,10 +139,10 @@ Fondant has a feature rich CLI that helps you with these steps. Let's start by r
 First of all make sure you have [Docker Compose](https://docs.docker.com/compose/) installed on your system
 
 ```bash
-fondant run pipeline:my_pipeline --local
+fondant run pipeline.py --local
```
mrchtr, line 142: Shouldn't this be `fondant compile pipeline --local`? Is the `.py` needed here?

mrchtr (approved):
Thanks @PhilippeMoussalli, lgtm. I've just found two small typos in the documentation.

PR #402, opened 2023-08-31
Bugfix filesystem kfp

PhilippeMoussalli:
PR that fixes a bug with the current filesystem implementation. We recently switched from using fsspec methods (`fsspec.fs_open`) which accepts any urlpaths (protocol+path) towards using the filesystem itself (`fsspec.filesystem`) which offer more functionalities like `glob` and date fetching which were needed for the caching approach.  (https://filesystem-spec.readthedocs.io/en/latest/api.html) 

The issue was that we were basing our filesystem based on the `base_path` which does not follow the filesystem required in the manifest in kfp: 

Local runner: 
- Remote base path follows remote manifest reference (e.g. both `gcs`)
- Local base path follows local manifest reference (e.g. both `/artifacts/...`)

 Remote runner: 
- Remote base path and a local manifest reference due to the way kubeflow artifact paths are resolved before mapping them to minio (/tmp/output_manifest/data). 

This PR checks the filesystem explicitly before writing. Note that this is just a temporarily workaround. Now that we don't need the additional functionalities of fsspec.filesystem, we can revert back to just using `fs_open`. This will make it easier to resolve all the different possible combination of URLs. 
https://github.com/ml6team/fondant/issues/368#issuecomment-1700824541

### DIFF
--- a/components/load_from_hf_hub/fondant_component.yaml
+++ b/components/load_from_hf_hub/fondant_component.yaml
@@ -1,6 +1,6 @@
 name: Load from hub
 description: Component that loads a dataset from the hub
-image: ghcr.io/ml6team/load_from_hf_hub:dev
+image: ghcr.io/ml6team/load_from_hf_hub:898484086417c9124e87e00ba7f7dc21ca73a26e
 
 produces:
   dummy_variable:  #TODO: fill in here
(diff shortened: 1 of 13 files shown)


NielsRogge (approved):
Thanks for fixing!

PR #403, opened 2023-09-03
Improve commoncrawl components

RobbeSneyders:
This is the version currently running (check the commits in the specs)


src/fondant/data_io.py
```diff
@@ -261,7 +261,6 @@ def _create_write_task(
             schema=schema,
             overwrite=False,
             compute=False,
-            write_metadata_file=True,
```
mrchtr, line 264: We should keep in mind to investigate the impact of this change later. In particular to answer the question which impact the writing metadata file affects memory release.

@PhilippeMoussalli do you think this change has an impact on existing components/pipelines?

mrchtr (approved):
Thanks @RobbeSneyders. Lgtm, beside the change regarding metadata writing. Maybe we can exclude this change before we merge the PR.

shayorshay (approved):

PR #407, opened 2023-09-04
Update caching strategy

PhilippeMoussalli:


src/fondant/executor.py
```diff
@@ -300,6 +289,32 @@ def execute(self, component_cls: t.Type[Component]) -> None:
 
         self.upload_manifest(output_manifest, save_path=self.output_manifest_path)
 
+    def _upload_manifest_reference(
```
GeorgesLorre, line 292: the naming in this function is a bit confusing cache key vs manifest reference

GeorgesLorre (approved):
Looks good, simpler so more elegant

src/fondant/executor.py
PhilippeMoussalli: Good call, i'll change it

PR #408, opened 2023-09-05
Add option to setup preemptible VMs

PhilippeMoussalli:
PR that adds an option to set a preemptible VM on the component spec level. 
This can be done by setting `preemptible` to True at the `Op` level. Note that you also need to assign a preemptible nodepool. I have already setup some preemptible nodepools in [this PR](https://github.com/ml6team/Express-infra/compare/add-preemtibles-node-pools?expand=1) 

Caveat: seems like kfp currently only enables setting up preemptibles on a GCP cluster since this functionaliteis seems to be tightly integrated with GKE. More info [here](
Preemptible nodepools can only run up to 24 hours [link](https://cloud.google.com/compute/docs/instances/preemptible#:~:text=Compute%20Engine%20always%20stops%20preemptible%20instances%20after%20they%20run%20for%2024%20hours.%20Certain%20actions%20reset%20this%2024%2Dhour%20counter.)

We can think later on incorporating a retry mechanism once the issue with Dask is fixed. For now those VMs can be used to reduce development costs.

### DIFF
--- a/docs/pipeline.md
+++ b/docs/pipeline.md
@@ -48,7 +48,16 @@ Next, we define two operations: `load_from_hub_op`, which is a based from a reus
     Currently Fondant supports linear DAGs with single dependencies. Support for non-linear DAGs will be available in future releases.
 
 ## Setting Custom node pool parameters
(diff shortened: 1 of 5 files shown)


GeorgesLorre (approved):
will be very helpful for future bigger runs

mrchtr (approved):
Thanks @PhilippeMoussalli!

PR #409, opened 2023-09-05
change meta estimation

PhilippeMoussalli:
PR that fixes an issue with the load from hub component when setting an index dynamically. Previous implementation relied on estimating the `meta/schema` dynamically but this lead to some errors since sometimes the wrong type was inferred. 

This PR fixes this issue by reading in the schema from the component spec

### DIFF
--- a/components/load_from_hf_hub/src/main.py
+++ b/components/load_from_hf_hub/src/main.py
@@ -2,24 +2,31 @@
 import logging
 import typing as t
 
+import dask
 import dask.dataframe as dd
 import pandas as pd
 from fondant.component import DaskLoadComponent
+from fondant.component_spec import ComponentSpec
 
 logger = logging.getLogger(__name__)
+
+dask.config.set({"dataframe.convert-string": False})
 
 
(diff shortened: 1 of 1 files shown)


components/load_from_hf_hub/src/main.py
```diff
@@ -74,14 +82,24 @@ def _set_unique_index(dataframe: pd.DataFrame, partition_info=None):
                 """Function that sets a unique index based on the partition and row number."""
                 dataframe["id"] = 1
                 dataframe["id"] = (
-                    str(partition_info["number"])
-                    + "_"
-                    + (dataframe.id.cumsum()).astype(str)
+                        str(partition_info["number"])
+                        + "_"
+                        + (dataframe.id.cumsum()).astype(str)
```
GeorgesLorre, line 87: we set the id as `str` here. Was there no problem with inferring divisions with this ?

GeorgesLorre: makes sense, thx!

GeorgesLorre (approved):

Issue #395, opened 2023-08-29
Use the Dask.distributed.LocalCluster


RobbeSneyders:
The clusters provided by `dask.distributed` are more performant and offer more control than the schedulers used by vanilla `dask`. It also looks like it is the main focus of current and future development in `dask`.

We should investigate if we can use the `LocalCluster` from `dask.distributed`, which still runs on a single machine, but provides a lot of benefits.

mrchtr:
I have taken a look at how we could implement this. I think the right place to incorporate this usage would be in the `Executor`.

I would add the `client` to the class attributes of the `Executor`. This would give us the possibility to configure the client based on initialization arguments, which can be defined in the spec file later on.

Technically, it is possible to have several clients in one Python session, but only one client can be active at a time. To prevent the overwriting of these settings, I would propose modifying the `DaskDataLoader`. I would suggest assigning the client to the dataset so that, for every computation, the defined client of the executor will be used.
Assignment of a client can be done be calling: `dataframe = dataframe.persist(client=client)`


Below, you can find pseudocode illustrating how I see the changes.

```python
from distributed import Client
import dask.dataframe as dd
from dask.distributed import LocalCluster, performance_report

dummy_dataframe = dd.from_dict({"id": [0, 1, 2, 3], "url": ["a", "b", "c", "d"]}, npartitions=2)

class DaskDataLoader():
    def load_dataframe(self, client=None):
        # fondant code
        # ...
        dataframe = dummy_dataframe

        if client:
            return dataframe.persist(client=client)

class Executer():
    def __init__(self):
        # Set arguments for localcluster, provide definition in component spec?
        worker_kwargs = {
            "n_workers": 1,
            "memory_limit": 2e9,
            "threads_per_worker": 1
        }

        self.client = Client(LocalCluster(**worker_kwargs))

    def _execute_component(self):
        data_loader = DummyDaskDataLoader(
           # manifest=manifest,
           # component_spec=self.spec,
           # input_partition_rows=self.input_partition_rows,
        )

        dataframe = data_loader.load_dataframe(self.client)

        # ...

        return dataframe

if __name__ == '__main__':
    executer = DummyExecuter()
    dataframe = executer._execute_component()
    dataframe = dummy_dataframe
    dataframe["column"] = "hello world"
    
    # If we want to store the dashboard as static html we need to use the context manager wrapping around the final execution
    with performance_report(filename="export.html"):
        dataframe.compute()
```

[...]

Issue #406, opened 2023-09-04
Out of memory caused by partition write operations


mrchtr:
# Issue 
The issue arises when we are writing the dask dataframe to parquet files. The process involves creating a separate Parquet file for each partition and column. We construct an execution graph where the final steps involve these write operations. 
However, dask does not execute the graph partitions sequentially, leading to inefficient memory usage and out-of-memory errors (related to [issue 10480](https://github.com/dask/dask/issues/10480)). The write step in the dask computation graph has dependencies on the columns of every partition.

![image](https://github.com/ml6team/fondant/assets/15777729/93f845de-9d6d-42bb-b280-5614bc82877b)

We need to optimize and declutter the write operations for better memory management and stable execution of fondant components.

## Option 1: Declutter execution graph

We could consider using delayed objects to simplify the execution graph. This involves wrapping each column write process of each partition into a delayed object, which could potentially enable sequential execution. 

![image](https://github.com/ml6team/fondant/assets/15777729/282bc710-8847-4290-ad12-2ac0f65955e1)

Pseudo code for wrapping the write process within a delayed object:
```python
def write_partition_to_parquet(partition, partition_index, output_dir):
        partition.to_parquet(os.path.join(output_dir, f'partition_{partition_index}.parquet'))

# this needs to be done for each column, e.g. COLUMN_df.to_delayed()
for partition_index, partition in enumerate(index_df.to_delayed()):
        delayed_task = delayed(write_partition_to_parquet)(partition, partition_index, output_dir + "/id")
        delayed_tasks.append(delayed_task)
```

However, I didn't find a practical solution for this yet. The execution of the graph is still not handled sequentially by dask. 
Also the graph looks still not optimal, cause we have `number_column x number_of_partitions` computations steps at the end. 

Maybe this could be a short-term solution if we can successfully execute the subgraphs sequentially.

## Option 2: Scalable workaround

I have been exploring a potentially more stable solution by leveraging the dask client in combination with a LocalCluster. 
Instead of using the `dataframe.compute()` method, I have been experimenting with `dataframe.persist()`. 
This approach triggers asynchronous partition-wise computation and immediately returns a list of pointers containing dask futures.

A dask future is a representation of an asynchronous computation. It can have different statuses, such as "pending" when waiting to complete and "finished" when the computation has successfully finished. 

To utilise this approach, I would iterate over the pointers of finished futures, retrieve the computed pandas dataframe, and then write it column-wise to a single parquet file. 
[...]

PR #410, opened 2023-09-06
Incorporate dask client

mrchtr:
PR addressing #395.
- Incorporate the usage of LocalCluster and dask client as default 
- Expose the default dask dashboard port in the docker-compose file, enabling the opening of the dashboard for debugging purposes


src/fondant/compiler.py
```diff
@@ -173,10 +173,14 @@ def _generate_spec(
             if extra_volumes:
                 volumes.extend(extra_volumes)
 
+            ports: t.List[t.Union[str, dict]] = []
+            ports.append("8787:8787")
```
mrchtr, line 177: Perhaps we can use a more dynamic approach. For now, I have just used the default ports of the dashboard.

src/fondant/data_io.py
```diff
@@ -189,7 +191,8 @@ def write_dataframe(self, dataframe: dd.DataFrame) -> None:
 
         with ProgressBar():
             logging.info("Writing data...")
-            dd.compute(*write_tasks)
+            # alternative implementation possible: futures = client.compute(...)
+            dd.compute(*write_tasks, scheduler=dask_client)
```
mrchtr, line 195: Alternative implementation: 

```python
futures = dask_client.compute(write_tasks)
wait(futures)
...
```

Here, we still have to investigate how we are addressing the oom issue. See proposal in #406

mrchtr (commented):
Some additional notes/questions:

src/fondant/compiler.py
PhilippeMoussalli: if we go for the client approach and don't revert back to the standard approach then we can always just expose the default port. Any reason why you might pick a different or multiple ports?

src/fondant/__init__.py
```diff
@@ -1,6 +1,12 @@
...
+import dask
+
 logging.basicConfig(
     format="[%(asctime)s | %(name)s | %(levelname)s] %(message)s",
     level=logging.INFO,
 )
+
+
+dask.config.set({"dataframe.convert-string": False})
+dask.config.set({"distributed.worker.daemon": False})
```
PhilippeMoussalli, line 12: what does this do?

src/fondant/__init__.py
```diff
@@ -1,6 +1,12 @@
...
 
+import dask
+
 logging.basicConfig(
     format="[%(asctime)s | %(name)s | %(levelname)s] %(message)s",
     level=logging.INFO,
 )
+
+
+dask.config.set({"dataframe.convert-string": False})
```
PhilippeMoussalli, line 11: I am not 100% if this guarantees that it's applied to the rest of the modules. I may have been wrong before. The best place for it indeed might be in the executor since that's the starting point and it will apply globally to all other modules called within it. 

More details here: 
image

src/fondant/data_io.py
```diff
@@ -160,7 +158,11 @@ def __init__(
     ):
         super().__init__(manifest=manifest, component_spec=component_spec)
 
-    def write_dataframe(self, dataframe: dd.DataFrame) -> None:
+    def write_dataframe(
+        self,
+        dataframe: dd.DataFrame,
+        dask_client: t.Optional[Client] = None,
```
PhilippeMoussalli, line 164: what happens if it's None, should it default to the LocalCluster? Ideally the default here would follow the one in the ComponentSpec

src/fondant/executor.py
```diff
@@ -33,7 +35,32 @@
 
 
 class Executor(t.Generic[Component]):
-    """An executor executes a Component."""
+    """
+    An executor executes a Component.
+
+    Args:
```
PhilippeMoussalli, line 41: Thanks for adding this! 
- Can you indent? 
- No need to add the datatype between two brackets. The convention we follow is to just have it in the function arguments

src/fondant/__init__.py
mrchtr: This config defines if a `distributed.worker` will run as a daemon/background process or not. 
So in our setup the processes will run in the foreground. I wasn't able to start the processes in the background inside of a docker container. I wasn't able to run the workers of the client as daemon processes. 
We just have to make sure that we apply this configuration before we are initialising the client. 
However, I think this change is indeed just needed for the usage of the LocalCluster/client. I will move it to the `Executer` class right before we initialising the client.

src/fondant/__init__.py
mrchtr: Yes indeed. See my comment above. Will move both configuration to the executer.

src/fondant/data_io.py
mrchtr: I wanted to avoid failing pipelines if we merge this one. 
We will call `dd.compute(*write_tasks, scheduler=None)` what will have either tries to load the global client (if available) or using the `default client` our existing approach.
We can revisit it again when we are fixing the oom issue.

src/fondant/compiler.py
mrchtr: I have separated the code and created a meaningful variable. This makes it clearer what the port represents.

src/fondant/pipeline.py
```diff
@@ -66,7 +70,20 @@ def __init__(
...
+        self.arguments = self._set_arguments(arguments)  #
+
+        self._add_component_argument(
+            "input_partition_rows",
+            input_partition_rows,
+            validate_partition_number,
+        )
+        self._add_component_argument("cache", cache)
+        self._add_component_argument("cluster_type", cluster_type)
+        self._add_component_argument("client_kwargs", client_kwargs)
```
mrchtr, line 85: For the first time reading the code, it wasn't clear from the initial perspective where the ComponentOp arguments are defined. The magic is actually hidden inside the `_set_arguments` function. I would propose a slightly different method, such as `_add_component_argument`. In my opinion, this would make it a bit clearer to understand what's happening just by looking at the class constructor.
@PhilippeMoussalli wdyt?

src/fondant/data_io.py
PhilippeMoussalli: ok that's clear, thanks!

src/fondant/executor.py
```diff
@@ -54,13 +76,45 @@ def __init__(
...
+            }
+
+        if cluster_type == "local":
+            logger.info(f"Initialize local dask cluster with arguments {client_kwargs}")
+
+            # Additional dask configuration have to be set before initialising the client
+            # worker.daemon is set to false because creating a worker process in daemon
+            # mode is not possible in our docker container setup.
+            dask.config.set({"distributed.worker.daemon": False})
+            dask.config.set({"dataframe.convert-string": False})
```
PhilippeMoussalli, line 93: does this one also just apply for the `local` cluster?

src/fondant/executor.py
```diff
@@ -54,13 +76,45 @@ def __init__(
...
+
     @classmethod
     def from_args(cls) -> "Executor":
         """Create an executor from a passed argument containing the specification as a dict."""
         parser = argparse.ArgumentParser()
         parser.add_argument("--component_spec", type=json.loads)
         parser.add_argument("--cache", type=ast.literal_eval)
         parser.add_argument("--input_partition_rows", type=validate_partition_number)
+        parser.add_argument("--cluster_type", type=ast.literal_eval)
+        parser.add_argument("--client_kwargs", type=json.loads)
```
PhilippeMoussalli, line 117: This might fail is `client_kwargs` is none, look at the function `validate_partition_number` to see how this is handled. 

FYI there is some cleanup to do in the whole argument parsing stuff so feel free to give suggestions

src/fondant/pipeline.py
PhilippeMoussalli: I think it makes sense indeed, looks more clear

src/fondant/component_spec.py
```diff
@@ -306,6 +318,7 @@ def from_fondant_component_spec(
                         *cls._dump_args(fondant_component.args.values()),
                         "--output_manifest_path",
                         {"outputPath": "output_manifest_path"},
+                        # TODO: check if it is needed to add cluster_type and client_kwargs
```
PhilippeMoussalli, line 321: this is needed indeed

Issue #393, opened 2023-08-29
Create Vertex runner


RobbeSneyders:
Create a runner that allows users to run Fondant pipelines on Vertex AI pipelines.
```[tasklist]
### Tasks
- [ ] https://github.com/ml6team/fondant/issues/412
- [ ] https://github.com/ml6team/fondant/issues/413
- [ ] https://github.com/ml6team/fondant/issues/414
- [ ] https://github.com/ml6team/fondant/issues/423
- [ ] https://github.com/ml6team/fondant/issues/422
- [ ] https://github.com/ml6team/fondant/issues/420
- [ ] https://github.com/ml6team/fondant/issues/419
- [ ] https://github.com/ml6team/fondant/issues/417
- [ ] https://github.com/ml6team/fondant/issues/416
- [ ] https://github.com/ml6team/fondant/issues/434
- [ ] https://github.com/ml6team/fondant/issues/437
- [ ] https://github.com/ml6team/fondant/issues/494
- [ ] https://github.com/ml6team/fondant/issues/495
- [ ] https://github.com/ml6team/fondant/pull/533
```

GeorgesLorre:
Vertex runs kubeflow pipelines in a more managed way since we already have a kubeflow runner (and compiler), vertex should be a logical next step.

For Vertex we [should use](https://cloud.google.com/vertex-ai/docs/pipelines/build-pipeline#sdk) the "new" kubeflow pipelines v2 which we currently do not support.

There are 3 options to make this work:

- we keep our current kfp runner v1 and create a new runner that uses kfp v2 **--> not possible since we can't have 2 versions of kfp installed at the same time** 

- we use the (v2-compatibility-mode)[https://www.kubeflow.org/docs/components/pipelines/v1/sdk-v2/v2-compatibility/#compiling-and-running-pipelines-in-v2-compatibility-mode] to both support v1 and v2. This means importing from `kfp.v2` for v2 functionality. This works in versions of `kfp==1.8.0` and up.

**--> seems like the logica way forward but the `kfpv1.v2` and `kfpv2` do not have the same interfaces and the documentation is very lacking** 
**--> feels like a temporary fix**

- we only use v2 and port our current code of the kubeflow runner to v2
**--> this means that we need to rewrite the kubeflow compiler and create the vertex compiler but there is a lot of overlapping code. The compiler can be the same and the runnen could be specific AFAIK.**
[...]

PR #411, opened 2023-09-06
Feature/vertex compiler

GeorgesLorre:


src/fondant/component_spec.py
```diff
@@ -230,76 +228,149 @@ class KubeflowComponentSpec:
     def __init__(self, specification: t.Dict[str, t.Any]) -> None:
         self._specification = specification
 
+    @staticmethod
+    def convert_arguments(fondant_component):
+        python2kubeflow_type = {
```
GeorgesLorre, line 233: I'll remove this since its defined at the top of the file

pyproject.toml
```diff
@@ -41,7 +41,7 @@ classifiers = [
 ]
 
 [tool.poetry.dependencies]
-python = ">= 3.8"
+python = ">= 3.8 < 3.12"
```
GeorgesLorre, line 44: kfp(v2) only supports up to python 3.12

src/fondant/compiler.py
```diff
@@ -305,3 +305,83 @@ def _set_configuration(self, task, fondant_component_operation):
...
+                    text=component_op.component_spec.kubeflow_specification.to_string(),
+                )
+
+                # Execute the Kubeflow component and pass in the output manifest path from
+                # the previous component.
+                component_args = component_op.arguments
+                metadata = json.dumps(
+                    {
+                        "base_path": self.pipeline.base_path,
+                        "run_id": "{{workflow.name}}",
```
GeorgesLorre, line 362: TODO: this workflow.name placeholder does not get replaced

src/fondant/component_spec.py
```diff
@@ -230,76 +228,149 @@ class KubeflowComponentSpec:
...
-            "name": fondant_component.name,
-            "description": fondant_component.description,
-            "inputs": [
-                {
-                    "name": "input_manifest_path",
-                    "description": "Path to the input manifest",
-                    "type": "String",
+    def from_fondant_component_spec(cls, fondant_component: ComponentSpec):
+        """Generate a Kubeflow component spec from a ComponentOp."""
+        input_definitions = {
```
GeorgesLorre, line 256: split now between artifacts and parameters

src/fondant/component_spec.py
```diff
@@ -33,15 +33,13 @@ def kubeflow2python_type(type_: str) -> t.Any:
...
-# TODO: Change after upgrading to kfp v2
-# :https://www.kubeflow.org/docs/components/pipelines/v2/data-types/parameters/
 python2kubeflow_type = {
-    "str": "String",
-    "int": "Integer",
-    "float": "Float",
-    "bool": "Boolean",
-    "dict": "JsonObject",
-    "list": "JsonArray",
+    "str": "STRING",
```
PhilippeMoussalli, line 37: Perhaps we can import this dict and use it? 

https://github.com/kubeflow/pipelines/blob/2b05ec867fad84e24fe73ef7515e3b5849297e79/sdk/python/kfp/dsl/types/type_utils.py#L56

We can maybe do the same for V1 as well instead of defining our own custom one 
https://github.com/kubeflow/pipelines/blob/2b05ec867fad84e24fe73ef7515e3b5849297e79/sdk/python/kfp/dsl/types/type_utils.py#L86

FYI I just stumbled on this when trying to look out for the v2 types since I couldn't find them in the docs

src/fondant/compiler.py
```diff
@@ -305,3 +305,83 @@ def _set_configuration(self, task, fondant_component_operation):
...
+                    component_task = kubeflow_component_op(
+                        input_manifest_path=manifest_path,
+                        metadata=metadata,
+                        **component_args,
+                    )
+                    component_task.after(previous_component_task)
+
+                else:
+                    component_task = kubeflow_component_op(
+                        metadata=metadata,
```
PhilippeMoussalli, line 378: Why not follow the same approach as the KFP compiler and set the input manifest as optional?

src/fondant/compiler.py
```diff
@@ -305,3 +305,83 @@ def _set_configuration(self, task, fondant_component_operation):
...
+        @self.kfp.dsl.pipeline(name=pipeline.name, description=pipeline.description)
+        def kfp_pipeline():
+            previous_component_task = None
+            manifest_path = None
+            for component_name, component in self.pipeline._graph.items():
+                logger.info(f"Compiling service for {component_name}")
+
+                component_op = component["fondant_component_op"]
+                # convert ComponentOp to Kubeflow component
+                kubeflow_component_op = self.kfp.components.load_component_from_text(
```
PhilippeMoussalli, line 352: It seems like this is the only difference between the two compiler (the SDK changed for this function). Can we perhaps reuse the same function instead of defining it twice. Seems like there is quite a bit of overlap

src/fondant/component_spec.py
PhilippeMoussalli, line 233: This seems to be defined twice

src/fondant/executor.py
```diff
@@ -197,35 +197,14 @@ def upload_manifest(self, manifest: Manifest, save_path: t.Union[str, Path]):
...
 
         """
-        is_kubeflow_output = (
-            str(save_path) == "/tmp/outputs/output_manifest_path/data"  # nosec
-        )
-
-        if is_kubeflow_output:
-            # Save to the expected base path directory
-            safe_component_name = self.spec.name.replace(" ", "_").lower()
-            save_path_base_path = (
```
PhilippeMoussalli, line 215: Are we then no longer saving the manifest relative to the basepath? We need it for data exploration
https://github.com/ml6team/fondant/issues/368#issuecomment-1700824541

PhilippeMoussalli (commented):
Thanks Geroges! Really nice work, looking clean ✨

Still a bit concerned on how this will end up integrating with KFP. Just out of curiosity, what would break with your current implementation if you switch back to v1 and import from v2?

src/fondant/component_spec.py
GeorgesLorre: I'll add it as a task

GeorgesLorre:
> Thanks Geroges! Really nice work, looking clean ✨
> 
> Still a bit concerned on how this will end up integrating with KFP. Just out of curiosity, what would break with your current implementation if you switch back to v1 and import from v2?

I'll add it as a task

src/fondant/executor.py
GeorgesLorre: Added as task

src/fondant/components/embedding_based_laion_retrieval/fondant_component.yaml
```diff
@@ -0,0 +1,31 @@
+name: LAION retrieval
```
PhilippeMoussalli, line 1: Why are different component spcs added?

src/fondant/component_spec.py
```diff
@@ -343,31 +416,54 @@ def to_string(self) -> str:
...
     def input_arguments(self) -> t.Mapping[str, Argument]:
         """The input arguments of the component as an immutable mapping."""
-        return types.MappingProxyType(
-            {
-                info["name"]: Argument(
-                    name=info["name"],
-                    description=info["description"],
-                    type=info["type"],
-                    default=info["default"] if "default" in info else None,
+        args = {}
```
PhilippeMoussalli, line 419: A lof of repeated code between input and output arguments, can be one put in one function

src/fondant/compiler.py
```diff
@@ -263,28 +264,32 @@ def kfp_pipeline():
...
-                    **component_args,
-                )
-                # Set optional configurations
-                component_task = self._set_configuration(
-                    component_task,
-                    component_op,
-                )
-
-                # Set image pull policy to always
-                component_task.container.set_image_pull_policy("Always")
```
PhilippeMoussalli, line 286: might be nice to keep the image pull policy (unless it's another function in v2 in which case we can tackle it separately or create a task for it)

tests/test_compiler.py
```diff
@@ -83,7 +83,7 @@ def now(cls):
 @pytest.fixture(params=TEST_PIPELINES)
 def setup_pipeline(request, tmp_path, monkeypatch):
     pipeline = Pipeline(
-        pipeline_name="test_pipeline",
+        pipeline_name="testpipeline",
```
PhilippeMoussalli, line 86: any special reason for this name change?

tests/test_compiler.py
```diff
@@ -232,30 +234,13 @@ def test_kubeflow_compiler(setup_pipeline, tmp_path_factory):
...
-    pipeline.add_op(component_1)
-    compiler = KubeFlowCompiler()
-    with tmp_path_factory.mktemp("temp") as fn:
-        output_path = str(fn / "kubeflow_pipeline.yml")
-        compiler.compile(pipeline=pipeline, output_path=output_path)
-        with open(output_path) as src, open(
-            VALID_PIPELINE / "kubeflow_pipeline.yml",
-        ) as truth:
-            assert yaml.safe_load(src) == yaml.safe_load(truth)
+# @pytest.mark.usefixtures("_freeze_time")
```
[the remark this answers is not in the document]
PhilippeMoussalli: ahh ok never mind it's because of the new API, would tackle it in another ticket

src/fondant/component_spec.py
```diff
@@ -230,85 +229,159 @@ class KubeflowComponentSpec:
...
+                                "{{$.inputs.parameters['component_spec']}}",
+                                "--input_partition_rows",
+                                "{{$.inputs.parameters['input_partition_rows']}}",
+                                "--cache",
+                                "{{$.inputs.parameters['cache']}}",
+                                *cls._dump_args(fondant_component.args.values()),
+                                "--output_manifest_path",
+                                "{{$.outputs.artifacts['output_manifest_path'].uri}}",
+                            ],
+                            "command": ["python3", "main.py"],
```
mrchtr, line 334: Shouldn't we use here `fondant execute main` instead?

GeorgesLorre: ah really ? Where do we do that ? I just took over the way we did it for kfpv1.

src/fondant/compiler.py
GeorgesLorre: added as a task: https://github.com/ml6team/fondant/issues/393

src/fondant/component_spec.py
GeorgesLorre: Agreed but this will probably be removed once we use the component spec as base for building the arguments and not the KubeflowComponentSpec.

https://github.com/ml6team/fondant/issues/393

src/fondant/component_spec.py
mrchtr: We changed while resolving this: #362

src/fondant/component_spec.py
```diff
@@ -230,85 +229,159 @@ class KubeflowComponentSpec:
...
+                        "schemaVersion": "0.0.1",
+                    },
                     "description": "Path to the output manifest",
-                    "type": "String",
                 },
-            ],
-            "implementation": {
-                "container": {
-                    "image": fondant_component.image,
-                    "command": [
```
PhilippeMoussalli, line 292: I think this is the command that you replaced, we're using this now for the components based on this PR https://github.com/ml6team/fondant/commits/main?before=82b5131ac8e72c620917d65c31c80b0b903e4d83+35&branch=main&qualified_name=refs%2Fheads%2Fmain

Can you replace and check if all still runs fine?

PhilippeMoussalli: @mrchtr

src/fondant/component_spec.py
PhilippeMoussalli: See comment above

src/fondant/component_spec.py
GeorgesLorre: updated

PhilippeMoussalli (approved):
Thanks Georges!

Issue #424, opened 2023-09-11
Update data explorer


PhilippeMoussalli:
Currently the data explorer expects the user to pass a local directory to mount. Sometimes we want to explore the dataset from remote pipelines. Also, the user has to manually pass in the manifest instead of specifying a pipelines run. 

To make the data explorer user-friendly, the user should ideally just pass in a base path and then the available pipelines and runs will be made available for them to explore 

```[tasklist]
### Tasks
- [ ] https://github.com/ml6team/fondant/issues/426
- [ ] https://github.com/ml6team/fondant/issues/432
```

PR #427, opened 2023-09-11
handle different base paths explorer

PhilippeMoussalli:
Linked to #426 #424 

Previous implementation of the local runner could only work with a local runner and not the remote one. This PR enables the explorer to handle both. If the base path is local, it will be mounted to the explorer. Otherwise, the base path will be passed as a container argument inside the component 

Notes: 
* Cloud credentials are optional for the local runner (manifest is local but can  point to a remote ) but mandatory with the remote base path


data_explorer/app/main.py
```diff
@@ -18,6 +19,15 @@
...
     # make sidebar with input fields for manifest path, subset and fields
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--base_path",
+        type=str,
+        help='Mounted or remote base path',
+    )
+    args = parser.parse_args()
+
+    # TODO: add remote path as an argument (Later PR)
```
mrchtr, line 30: I think it would be nice if you could create an issue for this task, so that we don't forget about fixing it later on.

data_explorer/app/filesystem.py
GeorgesLorre, line 1: double of the src/fondant/filesystem.py ?

PhilippeMoussalli: nice catch, indeed this is not needed

data_explorer/app/widgets.py
```diff
@@ -4,11 +4,12 @@
 import os
 from typing import Dict, List, Optional, Tuple, Union
 
+from fondant.filesystem import get_filesystem
```
GeorgesLorre, line 7: this import also needs a change I think ?

PhilippeMoussalli: no we still have src/filesystem.py so this should be fine

GeorgesLorre (approved):

PR #429, opened 2023-09-12
Add vertex runner

GeorgesLorre:
adresses: https://github.com/ml6team/fondant/issues/417

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,7 +41,7 @@ classifiers = [
 ]
 
 [tool.poetry.dependencies]
-python = ">= 3.8 < 3.12"
+python = ">= 3.8 < 3.11"
 dask = {extras = ["dataframe"], version = ">= 2023.4.1"}
 importlib-resources = { version = ">= 1.3", python = "<3.9" }
 jsonschema = ">= 4.18"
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -53,13 +53,14 @@ s3fs = { version = ">= 2023.4.0", optional = true }
 adlfs = { version = ">= 2023.4.0", optional = true }
 kfp = { version = "2.0.1", optional = true }
 pandas = { version = ">= 1.3.5", optional = true }
+google-cloud-aiplatform = "^1.32.0"
```
PhilippeMoussalli, line 56: Can you add it as optional similar to the above?

PhilippeMoussalli (approved):
Thanks Georges!

pyproject.toml
GeorgesLorre: good spot, updated!

src/fondant/runner.py
```diff
@@ -85,3 +85,36 @@ def get_name_from_spec(self, input_spec: str):
...
             return spec["pipelineInfo"]["name"]
+
+
+class VertexRunner(Runner):
+    def __resolve_imports(self):
+        import google.cloud.aiplatform as aip
+
+        self.aip = aip
+
+    def __init__(self, project_id: str, project_region: str, service_account: str = ""):
```
PhilippeMoussalli, line 96:
```suggestion
    def __init__(self, project_id: str, project_region: str, service_account: t.Optional[str]=None):
```

https://github.com/googleapis/python-aiplatform/blob/0fa47aea08b635159e9c6dc64f7d330f047dfb5d/google/cloud/aiplatform/preview/jobs.py#L232

src/fondant/runner.py
```diff
@@ -85,3 +85,36 @@ def get_name_from_spec(self, input_spec: str):
...
+    def run(self, input_spec: str, *args, **kwargs):
+        job = self.aip.PipelineJob(
+            display_name=self.get_name_from_spec(input_spec),
+            template_path=input_spec,
+            enable_caching=False,
+        )
+        if self.service_account:
+            job.submit(service_account=self.service_account)
+        else:
+            job.submit()
```
PhilippeMoussalli, line 114:
```suggestion
         job.submit(service_account=self.service_account)
```

with the above suggestion, since the service account defined in `submit()` can be optional

GeorgesLorre: updated

PR #436, opened 2023-09-18
Fix v2 defaults

PhilippeMoussalli:
PR that fixes the v2 default /optional types. KFP v2 has two different optional types: 

- Optional values that default to None -> defined by setting `isOptional` to True and not passing the default value
- Optional values that default to another values  -> defined by setting the value in  `defaultValue` and having `isOptional` False or empty. 

This PR adjusts the componentSpec accordingly. Another nice win is that we do not have to define bools/dicts/lists as strings in the component spec 

Few things no note: 

* Due to strict type. Mixing different data types is not allowed -> we had to change the type of `input_partition_rows` to int. Which means it does not accept the value `disable` anymore. Need to fix this in a seperate PR 
* 'inf` as float is not allowed

### DIFF
--- a/src/fondant/component_spec.py
+++ b/src/fondant/component_spec.py
@@ -55,12 +55,14 @@ class Argument:
         description: argument description
         type: the python argument type (str, int, ...)
         default: default value of the argument (defaults to None)
+        optional: whether an argument is optional or not (defaults to False)
     """
 
     name: str
     description: str
     type: str
     default: t.Optional[str] = None
+    optional: t.Optional[bool] = False
 
 
 class ComponentSubset:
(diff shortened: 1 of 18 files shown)


src/fondant/component_spec.py
```diff
@@ -55,12 +55,14 @@ class Argument:
...
         type: the python argument type (str, int, ...)
         default: default value of the argument (defaults to None)
+        optional: whether an argument is optional or not (defaults to False)
     """
 
     name: str
     description: str
     type: str
     default: t.Optional[str] = None
+    optional: t.Optional[bool] = False
```
GeorgesLorre, line 65: Nice! We needed this.

GeorgesLorre (approved):

PR #438, opened 2023-09-19
Set dask local scheduler as default

mrchtr:
Still, there might be some issues with the usage of the dask client and local cluster. To make it more intuitive, we should revert to using the old way as the default. This PR sets the old way as the default option, and the client usage will now need to be set explicitly.


PhilippeMoussalli (approved):
Thanks Matthias!

PR #439, opened 2023-09-19
Feature/args from spec

GeorgesLorre:
feature for https://github.com/ml6team/fondant/issues/412

### DIFF
--- a/src/fondant/executor.py
+++ b/src/fondant/executor.py
@@ -121,9 +121,12 @@ class Executor(t.Generic[Component]):
             if arg.name in cls.optional_fondant_arguments():
                 input_required = False
                 default = None
-            elif arg.default is not None or arg.optional is True:
+            elif arg.default is not None and arg.optional is False:
                 input_required = False
                 default = arg.default
+            elif arg.default is not None and arg.optional is True:
+                input_required = False
+                default = None
             else:
                 input_required = True
                 default = None
@@ -137,6 +140,10 @@ class Executor(t.Generic[Component]):
             )
 
         args, _ = parser.parse_known_args()
+        args.__dict__ = {
+            k: v if v != "None" else None for k, v in args.__dict__.items()
+        }
+
         return args
 
     @staticmethod
@@ -335,7 +342,7 @@ class DaskLoadExecutor(Executor[DaskLoadComponent]):
 
     @staticmethod
     def optional_fondant_arguments() -> t.List[str]:
-        return ["input_manifest_path"]
+        return ["input_manifest_path", "input_partition_rows"]
 
     def _load_or_create_manifest(self) -> Manifest:
         return Manifest.create(
@@ -399,6 +406,10 @@ class DaskTransformExecutor(TransformExecutor[DaskTransformComponent]):
 
 
 class PandasTransformExecutor(TransformExecutor[PandasTransformComponent]):
+    @staticmethod
+    def optional_fondant_arguments() -> t.List[str]:
+        return ["input_manifest_path", "input_partition_rows"]
+
     @staticmethod
     def wrap_transform(transform: t.Callable, *, spec: ComponentSpec) -> t.Callable:
         """Factory that creates a function to wrap the component transform function. The wrapper:
@@ -503,7 +514,7 @@ class DaskWriteExecutor(Executor[DaskWriteComponent]):
 
     @staticmethod
     def optional_fondant_arguments() -> t.List[str]:
-        return ["output_manifest_path"]
+        return ["input_partition_rows", "output_manifest_path"]
 
     def _load_or_create_manifest(self) -> Manifest:
         return Manifest.from_file(self.input_manifest_path)
--- a/tests/example_specs/components/arguments/component.yaml
+++ b/tests/example_specs/components/arguments/component.yaml
@@ -65,5 +65,4 @@ args:
   optional_arg:
     description: optional argument
     type: str
-    optional: True
     default: None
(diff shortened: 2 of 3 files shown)


src/fondant/component_spec.py
```diff
@@ -18,52 +17,50 @@
...
-    "int": "NUMBER_INTEGER",
-    "float": "NUMBER_DOUBLE",
-    "bool": "BOOLEAN",
-    "dict": "STRUCT",
-    "list": "LIST",
-}
-
 
 @dataclass
 class Argument:
```
GeorgesLorre, line 22: I feel like we could do more with this class:
- handle argument conversion to argparse and kfp
- do type checking (sane defaults)
- handle default/optional logics

src/fondant/executor.py
```diff
@@ -130,7 +130,7 @@ def _add_and_parse_args(cls, spec: ComponentSpec):
```
PhilippeMoussalli, line 130: I think we need to add another conditional to avoid returning `None` as a string (should be fixed in the tests below)

could you try replacing the above with this and retrying the tests with actual `None` type? 

```python
for arg in component_arguments.values():
    if arg.name in cls.optional_fondant_arguments():
        input_required = False
        default = None
    elif arg.default is not None and arg.optional is False:
        input_required = False
        default = arg.default
    elif arg.default is not None and arg.optional is True:
        input_required = False
        default = None
    else:
        input_required = True
        default = None
```

src/fondant/executor.py
```diff
@@ -154,9 +154,8 @@ def _get_component_arguments(spec: ComponentSpec) -> t.Dict[str, Argument]:
             Input and output arguments of the component.
         """
         component_arguments: t.Dict[str, Argument] = {}
-        kubeflow_component_spec = spec.kubeflow_specification
```
PhilippeMoussalli, line 157: Suggestions to add this after line 139

```python
 args.__dict__ = {k: v if v != "None" else None for k, v in args.__dict__.items()}
```

Currently we check if types are `optional` by checking the value in the `default` field (if it's None or another value). However, sometimes the value in the default is a string that we want to override with `None`. 

So either we: 

* Add an additional field to the user component spec where they have to specifically mention if a field is optional -> advantage more explicit, but more things for the user to implement 
* Implement the suggestion above, less things for the user to worry about, might lead to an edge case where the user actually wants to have "None" as a string

tests/test_component.py
```diff
@@ -130,15 +126,15 @@ def _process_dataset(self, manifest: Manifest) -> t.Union[None, dd.DataFrame]:
         "list_default_arg": ["foo", "bar"],
         "dict_default_arg": {"foo": 1, "bar": 2},
         "string_default_arg": "foo",
-        "string_default_arg_none": None,
-        "integer_default_arg_none": None,
-        "float_default_arg_none": None,
-        "bool_default_arg_none": None,
-        "list_default_arg_none": None,
-        "dict_default_arg_none": None,
+        "string_default_arg_none": "None",
```
PhilippeMoussalli, line 129: This should be Nontype and not string (see suggestions above). 
Same for the last two arguments

tests/example_specs/components/arguments/component.yaml
```diff
@@ -38,31 +38,32 @@ args:
...
-    description: argument with default None value type that can be overriden with a valid python type
-    type: float
-    default: None
   override_default_arg_with_none:
     description: argument with default python type that can be overriden with None
     type: str
+  optional_arg:
+    description: optional argument
+    type: str
+    optional: True
```
PhilippeMoussalli, line 68: I don't think specifying `optional` is needed. This is currently hidden from the user and inferred based on the value in the `default` field

PhilippeMoussalli (commented):
Thanks Georges! 
This was much needed and looks much cleaner already. Left a few remarks concerning the parsing of default types and overriding default args

src/fondant/compiler.py
```diff
@@ -302,33 +291,47 @@ def kfp_pipeline():
...
-                    )
-                    component_task.after(previous_component_task)
+                output_manifest_path = (
+                    f"{pipeline.base_path}/{pipeline.name}/"
+                    f"{run_id}/{component_name}/manifest.json"
+                )
+                # Set the execution order of the component task to be after the previous
+                # component task.
+                if component["dependencies"]:
+                    for dependency in component["dependencies"]:
```
RobbeSneyders, line 309: This won't work for more then one dependency, right? Maybe we should make this explicit by checking that the amount of dependencies is 1 and selecting the first one.

src/fondant/compiler.py
```diff
@@ -404,7 +407,6 @@ def compile(
         @self.kfp.dsl.pipeline(name=pipeline.name, description=pipeline.description)
         def kfp_pipeline():
```
RobbeSneyders, line 408: This looks identical to the one above. Should we lift it to a base class?

tests/example_pipelines/compiled_pipeline/example_1/kubeflow_pipeline.yml
```diff
@@ -1,13 +1,10 @@
+# PIPELINE DEFINITION
+# Name: testpipeline
+# Description: description of the test pipeline
```
RobbeSneyders, line 3: Why is this commented at the top of each pipeline.yml?

RobbeSneyders (commented):
Thanks @GeorgesLorre!

Nice improvements indeed. Lacking some context, but did my best to review anyway :)

src/fondant/compiler.py
GeorgesLorre: I have a feeling that the compiler for vertex and kfpV2 can be the same. I would like to test both more and then we can merge (or abstract)

tests/example_pipelines/compiled_pipeline/example_1/kubeflow_pipeline.yml
GeorgesLorre: its generated by kfp

Issue #440, opened 2023-09-19
LocalRunner: Initial pipeline run will fail if no cache dir is available


mrchtr:
A initial pipeline run will fail when there is no cache directory present in the artifact directory.

In the executor.py file, we are writing the cache key to the 'cache' directory using the following code:

```
with fs_open(manifest_reference_path, mode="wt", encoding="utf-8") as file_:
    file_.write(str(manifest_save_path))
```

During the initial pipeline run, the cache directory is not created, and the file write operation fails with the error message: `FileNotFoundError: [Errno 2] No such file or directory: ...`

The issue can be resolved manually by creating a `cache` dir inside the artifact directory.

PR #442, opened 2023-09-20
Use auto_mkdir in fs_open calls

mrchtr:
There was a change of the default behaviour in the recent fsspec release (2023.9.1)
Use the `auto_mkdir` in fs_open to utilize automatic directory creation. 

Solves #440


PhilippeMoussalli (approved):
Thanks Matthias!

PR #444, opened 2023-09-21
Feature/no artifacts

GeorgesLorre:


src/fondant/compiler.py
```diff
@@ -302,33 +291,47 @@ def kfp_pipeline():
...
 
                 else:
                     component_task = kubeflow_component_op(
                         metadata=metadata.to_json(),
+                        output_manifest_path=output_manifest_path,
                         **component_args,
                     )
 
-                # Set optional configurations
+                # Set optional arguments
```
PhilippeMoussalli, line 329: I think this should still be configurations

PhilippeMoussalli (approved):
Beautiful ✨

PR #461, opened 2023-09-25
Bugfix/sample pipeline cc 25m

shayorshay:
This PR:

- adds the missing fields `surt_url` and `top_level_domain` needed to run the sample CC pipeline
-  adjusts `n_rows_to_load` to 100k to match the HF dataset card

### DIFF
--- a/examples/pipelines/filter-cc-25m/components/load_from_hf_hub/fondant_component.yaml
+++ b/examples/pipelines/filter-cc-25m/components/load_from_hf_hub/fondant_component.yaml
@@ -1,6 +1,6 @@
 name: Load from hub
 description: Component that loads a dataset from the hub
-image: ghcr.io/ml6team/load_from_hf_hub:dev
+image: ghcr.io/ml6team/load_from_hf_hub:latest
 
 produces:
   images:


examples/pipelines/filter-cc-25m/pipeline.py
```diff
@@ -30,22 +30,19 @@ def create_directory_if_not_exists(path):
...
     component_dir="components/load_from_hf_hub",
     arguments={
         "dataset_name": "fondant-ai/fondant-cc-25m",
         "column_name_mapping": load_component_column_mapping,
-        "n_rows_to_load": 1000,  # Here you can modify the number of images you want to download.
+        "n_rows_to_load": 100000,  # Here you can modify the number of images you want to download.
     },
 )
 
-# Filter mime type component
```
PhilippeMoussalli, line 44: Why are we removing this one again?

shayorshay: its not used here but in `filter_pipeline.py`

PhilippeMoussalli: ohh ok thanks :)

PhilippeMoussalli (approved):

mrchtr (approved):
Thanks @shayorshay!

PR #469, opened 2023-09-27
Fix caching writes

PhilippeMoussalli:
PR that fixes issues related to caching subsequent components. 

Since cached manifests contain the same run id as the pipeline where they originated from, they can sometimes override existing data if the component needs to run again (e.g. cached first component loaded into second component which has different arguments and thus needs to execute). This PR makes sure that:

* The run-id of the loaded **input manifest** matches the one from the **current pipeline run**, this ensures that all new subsets from the component being executed are written under the corresponding path of that pipeline run 
*  The run-id of the **output manifest** is that of the **cached manifest** (if it applies). This ensures that subsequent components are able to tell whether the previous component execution was cached or not [link to code](https://github.com/ml6team/fondant/blob/39436e7faaa3e5acdd5eb347ce965ad33d669f73/src/fondant/executor.py#L294)


RobbeSneyders:
So If I'm understanding correctly, the logic should be as follows:

- If the component is cached, it outputs the cached manifest with the run_id of the previous run
- If the component is not cached, it outputs a new manifest with the run_id of the current run
- A component checks if the previous component was cached by comparing the run_id in the manifest it receives with the run_id of the current run (from the metadata).

But the problem is that the component propagates the run_id from the input manifest to the output manifest.

If that is correct, can we change it where it's propagated instead? So in either:
- [Executor.run_execution()](https://github.com/ml6team/fondant/blob/62b6658e28843497af45c9d5cc3fb98b617241b7/src/fondant/executor.py#L337)
- [Manifest.evolve()](https://github.com/ml6team/fondant/blob/39436e7faaa3e5acdd5eb347ce965ad33d669f73/src/fondant/manifest.py#L257)

I believe a change there would be simpler and easier to understand.

RobbeSneyders:
I think it can still be further simplified. Just pass the `run_id` to `.evolve()` and set it in the metadata at the start. That `run_id` is then used for the paths automatically as well, and then we don't need to update the metadata separately.

PhilippeMoussalli:
> I think it can still be further simplified. Just pass the `run_id` to `.evolve()` and set it in the metadata at the start. That `run_id` is then used for the paths automatically as well, and then we don't need to update the metadata separately.

Thanks for the feedback Robbe! I implemented the suggestion. Please double check to see if that's what you were implying. 

Also added other small fixes mainly related to propogating the `cache` arguments from the op to the pipeline spec

src/fondant/manifest.py
```diff
@@ -257,21 +257,27 @@ def remove_subset(self, name: str) -> None:
...
         component_spec: ComponentSpec,
+        write_run_id: str,
     ) -> "Manifest":
         """Evolve the manifest based on the component spec. The resulting
         manifest is the expected result if the current manifest is provided
         to the component defined by the component spec.
+
+        Args:
+            component_spec: the component spec
+            write_run_id: the run id used to define the write subset path.
```
RobbeSneyders, line 268:
```suggestion
            run_id: the run id to include in the evolved manifest. If no run id is provided, the run id from the original manifest is propagated.
```

src/fondant/manifest.py
```diff
@@ -257,21 +257,27 @@ def remove_subset(self, name: str) -> None:
     def evolve(  # noqa : PLR0912 (too many branches)
         self,
         component_spec: ComponentSpec,
+        write_run_id: str,
```
RobbeSneyders, line 260:
```suggestion
        *,
        run_id: t.Optional[str],
```

src/fondant/manifest.py
```diff
@@ -257,21 +257,27 @@ def remove_subset(self, name: str) -> None:
...
+
+        Args:
+            component_spec: the component spec
+            write_run_id: the run id used to define the write subset path.
         """
         evolved_manifest = self.copy()
 
         # Update `component_id` of the metadata
         component_id = component_spec.name.lower().replace(" ", "_")
+        evolved_manifest.update_metadata(key="run_id", value=write_run_id)
```
RobbeSneyders, line 274:
```suggestion
        if run_id is not None:
            evolved_manifest.update_metadata(key="run_id", value=write_run_id)
```

src/fondant/manifest.py
```diff
@@ -257,21 +257,27 @@ def remove_subset(self, name: str) -> None:
...
         evolved_manifest.update_metadata(key="component_id", value=component_id)
 
         # Update index location as this is currently always rewritten
         evolved_manifest.index._specification[
             "location"
-        ] = f"/{self.pipeline_name}/{self.run_id}/{component_id}/index"
+        ] = f"/{self.pipeline_name}/{write_run_id}/{component_id}/index"
```
RobbeSneyders, line 280:
```suggestion
        ] = f"/{self.pipeline_name}/{evolved_manifest.run_id}/{component_id}/index"
```

src/fondant/manifest.py
```diff
@@ -322,7 +328,7 @@ def evolve(  # noqa : PLR0912 (too many branches)
                 # Update subset location as this is currently always rewritten
                 evolved_manifest.subsets[subset_name]._specification[
                     "location"
-                ] = f"/{self.pipeline_name}/{self.run_id}/{component_id}/{subset_name}"
+                ] = f"/{self.pipeline_name}/{write_run_id}/{component_id}/{subset_name}"
```
RobbeSneyders, line 331:
```suggestion
                ] = f"/{self.pipeline_name}/{evolved_manifest.run_id}/{component_id}/{subset_name}"
```

src/fondant/pipeline.py
```diff
@@ -444,7 +444,7 @@ def _validate_pipeline_definition(self, run_id: str):
                             raise InvalidPipelineDefinition(
                                 msg,
                             )
-            manifest = manifest.evolve(component_spec)
+            manifest = manifest.evolve(component_spec, write_run_id=run_id)
```
RobbeSneyders, line 447:
```suggestion
            manifest = manifest.evolve(component_spec, run_id=run_id)
```

tests/test_manifest_evolution.py
RobbeSneyders, line 1: The changes in this file are no longer needed, but we might want to add a test to test that the `run_id` is updated.

tests/test_manifest_evolution.py
```diff
@@ -28,7 +28,10 @@ def examples():
 def test_evolution(input_manifest, component_spec, output_manifest):
     manifest = Manifest(input_manifest)
     component_spec = ComponentSpec(component_spec)
-    evolved_manifest = manifest.evolve(component_spec=component_spec)
+    evolved_manifest = manifest.evolve(
+        component_spec=component_spec,
+        write_run_id=manifest.run_id,
```
RobbeSneyders, line 33: Can we update this test to provide a different run_id and validate that it's set correctly?

tests/test_manifest_evolution.py
```diff
@@ -42,7 +45,10 @@ def test_component_spec_location_update():
 
     manifest = Manifest(input_manifest)
     component_spec = ComponentSpec(specification)
-    evolved_manifest = manifest.evolve(component_spec=component_spec)
+    evolved_manifest = manifest.evolve(
+        component_spec=component_spec,
+        write_run_id=manifest.run_id,
```
RobbeSneyders, line 50: This change can be reverted.

RobbeSneyders (commented):
Thanks @PhilippeMoussalli!

Almost there :slightly_smiling_face:  I would make the argument optional and rename it to just `run_id`. See my suggestions.

RobbeSneyders (approved):
Thanks @PhilippeMoussalli!

PR #470, opened 2023-09-27
Improve hit rate of download images component

RobbeSneyders:
This PR contains two changes to improve the hit rate of the download images component:

- Introduce a Semaphore to limit the amount of concurrent requests. This should help with the timeout errors. If this slows down the speed too much, we can increase the number. If the changes in this PR work, we probably want to make this an argument to the component.
- Configure httpx to follow redirects, [which it doesn't do by default](https://www.python-httpx.org/compatibility/#redirects), but requests / urllib3 does.

### DIFF
--- a/components/download_images/fondant_component.yaml
+++ b/components/download_images/fondant_component.yaml
@@ -28,6 +28,10 @@ args:
     description: Number of times to retry downloading an image if it fails.
     type: int
     default: 0
+  n_connections:
+    description: Number of concurrent connections opened per process. Decrease this number if you are running into timeout errors.
+    type: int
+    default: 1000
   image_size:
     description: Size of the images after resizing.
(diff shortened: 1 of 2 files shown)


components/download_images/fondant_component.yaml
```diff
@@ -28,6 +28,10 @@ args:
     description: Number of times to retry downloading an image if it fails.
     type: int
     default: 0
+  n_connections:
+    description: Number of concurrent connections opened per process. Decrease this number if you are running into timeout errors.
```
PhilippeMoussalli, line 32: Can you also mention that lower connections improve the success rate but might result in slower downloads?

PhilippeMoussalli (approved):
Thanks Robbe!

PR #474, opened 2023-09-28
Create load from parquet

PhilippeMoussalli:
PR that creates a generic component to load data stored as a parquet from remote storage 

Largely based on the load from hub component with some modification. Currently needed for the datacomp pipeline. 

Future modification could include loading data from local storage though we'd need to make sure that the data is mounted locally

### DIFF
--- a/components/load_from_parquet/Dockerfile
+++ b/components/load_from_parquet/Dockerfile
@@ -1,0 +1,23 @@
+FROM --platform=linux/amd64 python:3.8-slim
+
+# System dependencies
+RUN apt-get update && \
+    apt-get upgrade -y && \
+    apt-get install git -y
+
+# Install requirements
+COPY requirements.txt /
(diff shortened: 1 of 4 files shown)


PhilippeMoussalli:
> The component does probably the job for now. I am just curious about if we can reuse the load from hub component completely. In the load_from_hub component we do something like this:
> 
> ```
> dask_df = dd.read_parquet(f"hf://datasets/{self.dataset_name}")
> ```
> 
> If I see it correctly, the only part that is different here is `hf://` vs. `gs//` or `s3://`. What do you think about to adjust the load from hub component to a more generic one?

Fairpoint, there is not a lot of difference between them and it can be generalized. I would tackle this in another PR though since it might affect other pipelines (need to change the load from hub component name, examples, docs, ...) [#475](https://github.com/ml6team/fondant/issues/475)

mrchtr (approved):

PR #477, opened 2023-09-29
KfpV2 and Vertex 

GeorgesLorre:


examples/pipelines/datacomp/pipeline.py
```diff
@@ -41,64 +41,66 @@
...
+)
+detect_text_op = ComponentOp(
+    component_dir="components/detect_text",
+    arguments={
+        "batch_size": 2,
+    },
+    node_pool_label="node_pool",
+    node_pool_name="model-inference-mega-pool",
+    number_of_accelerators=1,
+    accelerator_name="GPU",
```
RobbeSneyders, line 70: Not something we need to tackle in this PR, but we might want to make the infrastructure arguments dynamic instead by eg. accepting a dict that we pass on.

pyproject.toml
```diff
@@ -33,15 +33,14 @@ classifiers = [
...
-    "Programming Language :: Python :: 3.11",
     "Topic :: Software Development",
     "Topic :: Software Development :: Libraries",
     "Topic :: Software Development :: Libraries :: Python Modules",
     "Typing :: Typed",
 ]
 
 [tool.poetry.dependencies]
-python = ">= 3.8"
+python = ">= 3.8 < 3.11"
```
RobbeSneyders, line 43: We shouldn't set an upper version for Python. We've had issues with it before on Connexion. See:
- https://github.com/python-poetry/poetry/issues/3747
- https://stackoverflow.com/questions/73116647/why-cant-i-install-a-python-package-with-the-python-requirement-3-8-3-11-i

pyproject.toml
```diff
@@ -51,14 +50,16 @@ fsspec = { version = ">= 2023.4.0", optional = true}
...
+kfp = { version = "2.0.1", optional = true, extras =["kubernetes"] }
 pandas = { version = ">= 1.3.5", optional = true }
+google-cloud-aiplatform =  { version = "1.32.0", optional = true}
 
 [tool.poetry.extras]
 aws = ["fsspec", "s3fs"]
 azure = ["fsspec", "adlfs"]
 gcp = ["fsspec", "gcsfs"]
 kfp = ["kfp"]
+vertex = ["kfp", "google-cloud-aiplatform"]
```
RobbeSneyders, line 62: Should we include the `gcp` dependencies as well?

.github/workflows/pipeline.yaml
```diff
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.8', '3.9', '3.10', '3.11']
+        python-version: ['3.8', '3.9', '3.10']
```
RobbeSneyders, line 14: Why do we need to drop 3.11?

src/fondant/executor.py
```diff
@@ -188,22 +188,29 @@ def _add_and_parse_args(cls, spec: ComponentSpec):
             if arg.name in cls.optional_fondant_arguments():
                 input_required = False
                 default = None
-            elif arg.default is not None:
+            elif arg.default is not None and arg.optional is False:
                 input_required = False
                 default = arg.default
+            elif arg.default is not None and arg.optional is True:
+                input_required = False
+                default = None
```
RobbeSneyders, line 196: Can you explain this logic?

src/fondant/compiler.py
```diff
@@ -270,67 +307,189 @@ def kfp_pipeline():
...
+            task = self.kfp_kubernetes.add_node_selector(
+                task,
+                node_pool_label,
+                node_pool_name,
             )
-            task.apply(self.kfp_gcp.use_preemptible_nodepool())
+        return task
+
+
+class VertexCompiler(Compiler):
```
RobbeSneyders, line 381: As discussed, would check what we can merge with the `KubeFlowCompiler`

src/fondant/component_spec.py
```diff
@@ -191,21 +190,83 @@ def outputs_additional_subsets(self) -> bool:
...
+        args = self.default_arguments
+
+        args.update(
+            {
+                name: Argument(
+                    name=name,
+                    description=arg_info["description"],
+                    type=arg_info["type"],
+                    default=arg_info["default"] if "default" in arg_info else None,
+                    optional=_is_optional(arg_info),
```
RobbeSneyders, line 208:
```suggestion
                    optional=arg_info.get("default") == "None",
```

src/fondant/component_spec.py
```diff
@@ -230,101 +291,106 @@ class KubeflowComponentSpec:
...
+        )
+
     @classmethod
-    def from_fondant_component_spec(
-        cls,
-        fondant_component: ComponentSpec,
-    ) -> "KubeflowComponentSpec":
-        """Create a Kubeflow component spec from a Fondant component spec."""
+    def from_fondant_component_spec(cls, fondant_component: ComponentSpec):
+        """Generate a Kubeflow component spec from a ComponentOp."""
```
RobbeSneyders, line 328:
```suggestion
        """Generate a Kubeflow component spec from a Fondant component spec."""
```

src/fondant/compiler.py
RobbeSneyders: Let's pick this up when finishing Vertex.

components/load_from_hf_hub/fondant_component.yaml
```diff
@@ -19,11 +19,11 @@ args:
...
-    default: None
+    default: []
   n_rows_to_load:
     description: Optional argument that defines the number of rows to load. Useful for testing pipeline runs on a small scale
     type: int
-    default: None
+    default: -1
   index_column:
     description: Column to set index to in the load component, if not specified a default globally unique index will be set
     type: str
```
PhilippeMoussalli, line 29: Does it make sense to later on revert back to None and handle the conversion internally? It could be something like setting an arbitrary value if it's None in the fondant component and parsing it back to None again when the component is run

components/load_from_hf_hub/src/main.py
```diff
@@ -20,9 +19,9 @@ def __init__(self,
                  *_,
                  dataset_name: str,
                  column_name_mapping: dict,
-                 image_column_names: t.Optional[list],
-                 n_rows_to_load: t.Optional[int],
-                 index_column: t.Optional[str],
+                 image_column_names: list,
```
PhilippeMoussalli, line 22: Can we keep them as optional since they have a default or does `optional` only apply when the default is `None` by convention?

docs/components/component_spec.md
```diff
@@ -127,9 +127,6 @@ The `args` section describes which arguments the component takes. Each argument
 `description` and a `type`, which should be one of the builtin Python types. Additionally, you can
 set an optional `default` value for each argument.
 
-_Note:_ default iterable arguments such as `dict` and `list` have to be passed as a string
```
PhilippeMoussalli, line 130: Nice

examples/pipelines/datacomp/pipeline.py
```diff
@@ -41,70 +40,70 @@
...
+)
+download_images_op = ComponentOp.from_registry(
+    name="download_images",
+    arguments={
+        "retries": 2,
+        "min_image_size": 0,
+    },
+    node_pool_label="node_pool",
+    node_pool_name="n2-standard-64-pool",
+    input_partition_rows=-1,
```
PhilippeMoussalli, line 56: is this needed since it's the default?

examples/pipelines/datacomp/pipeline.py
```diff
@@ -41,70 +40,70 @@
...
+    cache=False,
+)
+detect_text_op = ComponentOp(
+    component_dir="components/detect_text",
+    arguments={
+        "batch_size": 2,
+    },
+    node_pool_label="node_pool",
+    node_pool_name="model-inference-mega-pool",
```
PhilippeMoussalli, line 65:
```suggestion
    node_pool_name="model-inference-pool",
```
A100s are not really available atm

examples/pipelines/datacomp/pipeline.py
PhilippeMoussalli, line 1: Can we re-enable the cache for all components?

PhilippeMoussalli (commented):
Thanks Robbe! Left a few comments. Can we maybe merge this one into another branch and then tackle all other issues on other branches or would you prefer to merge directly to main?

components/load_from_hf_hub/fondant_component.yaml
RobbeSneyders: I'm not sure if it's such a bad thing that the default needs to match the argument type. We would also need an arbitrary value for each type to make this work.

components/load_from_hf_hub/src/main.py
RobbeSneyders: Optional means that a `None` value can be passed in or set as a default. This is no longer the case now.

examples/pipelines/datacomp/pipeline.py
RobbeSneyders: No, it's not, I'll remove it.

.github/workflows/pipeline.yaml
RobbeSneyders: Dependency resolving of 3.11 is really slow. Seems to be related to both `kfp` and `google-cloud-aiplatform`

components/load_from_hf_hub/fondant_component.yaml
PhilippeMoussalli: Before we would assign a value as an optional argument if the default is set to `None`. Now the notion of optional arguments is missing and optional arguments are checked implicitly against arbitrary values defined in the default which can be a bit confusing. 

Was more thinking from the perspective of the end user. So if they would assign a default to `None`, we would handle it as optional and return it as `None` to the end user. Implementing this is still not very clear to me but we could just assign arbitrary known values. For example, if the users sets an list argument to `None`, we would pass it as `[]` and the convert it back to `None` during parsing

components/load_from_hf_hub/src/main.py
PhilippeMoussalli: I still think we should have optional argument (see comment above)

examples/pipelines/datacomp/pipeline.py
```diff
@@ -41,70 +40,64 @@
...
-#         "pct_threshold": 0.5,
-#     },
-#     node_pool_label="node_pool",
-#     node_pool_name="n2-standard-64-pool",
-# )
-
 
 # add ops to pipeline
 pipeline.add_op(load_from_hub_op)
-# pipeline.add_op(download_images_op, dependencies=load_from_hub_op)
```
PhilippeMoussalli, line 105: Can you uncomment here?

PhilippeMoussalli (approved):
Thanks Robbe, left one comment regarding arguments but I think we can discuss it separately

RobbeSneyders (approved):
Thanks, let's merge!

PR #478, opened 2023-10-02
Bugfix partitioning

PhilippeMoussalli:
The number of partitions set based on the number of specified rows per partition (optional argument) did not take into account the number of available cpu cores which can be suboptimal if the # cores is larger than the inferred number of estimated partitions. 

This PR addresses this bug by taking into account the number of cores before repartitioning

### DIFF
--- a/src/fondant/data_io.py
+++ b/src/fondant/data_io.py
@@ -36,27 +36,60 @@ class DaskDataLoader(DataIO):
         Returns:
             The partitioned dataframe.
         """
+        n_workers: int = os.cpu_count()  # type: ignore
+        original_n_partitions = dataframe.npartitions
+
         if self.input_partition_rows != "disable":
             if isinstance(self.input_partition_rows, int):
                 # Only load the index column to trigger a faster compute of the rows
                 total_rows = len(dataframe.index)
                 # +1 to handle any remainder rows
-                n_partitions = (total_rows // self.input_partition_rows) + 1
-                dataframe = dataframe.repartition(npartitions=n_partitions)
+                n_partitions_from_row = (total_rows // self.input_partition_rows) + 1
                 logger.info(
-                    f"Total number of rows is {total_rows}.\n"
-                    f"Repartitioning the data from {dataframe.partitions} partitions to have"
-                    f" {n_partitions} such that the number of partitions per row is approximately"
+                    f"Number of specified rows per partition is "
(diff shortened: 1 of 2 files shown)


RobbeSneyders:
I'm not sure if we need to fix this @PhilippeMoussalli.

It depends on which meaning we want the `rows_per_partition` argument to have. Before, we were using it as a target: The user specifies the number of rows they want per partition, fondant tries to achieve that number as closely as possible.

With your change, we handle it as `max_rows_per_partition` instead. If we want this, we might want to change the name or at least the description of the parameter. Caveat here is that we approximate the number of rows per partition, so it might still be slightly higher than the value set by the argument.

If we do want to go this way, I think we can simplify the logic a bit to:
```
if max(n_paritions_from_row, n_workers) > original_partitions:
    if n_paritions_from_row > n_workers:
        # Repartition based on n_partitions_from_row
    else:
        # Repartition based on n_workers
```

PhilippeMoussalli:
> I'm not sure if we need to fix this @PhilippeMoussalli.
> 
> It depends on which meaning we want the `rows_per_partition` argument to have. Before, we were using it as a target: The user specifies the number of rows they want per partition, fondant tries to achieve that number as closely as possible.
> 
> With your change, we handle it as `max_rows_per_partition` instead. If we want this, we might want to change the name or at least the description of the parameter. Caveat here is that we approximate the number of rows per partition, so it might still be slightly higher than the value set by the argument.
> 
> If we do want to go this way, I think we can simplify the logic a bit to:
> 
> ```python
> if max(n_paritions_from_row, n_workers) > original_partitions:
>     if n_paritions_from_row > n_workers:
>         # Repartition based on n_partitions_from_row
>     else:
>         # Repartition based on n_workers
> ```

Come to think of it, I indeed don't know if we need to include it. I initially wanted to change it since it wasn't allowing me to properly benchmark but that's just because the specified number of rows resulted in partitions that are less than the amount of workers. It might be that indeed for some reason you want to enforce a certain number of partitions per row. 

Perhaps we can keep it as it originally was and just return a warning if the resulting number of partitions are lower than the workers? Wdyt?

RobbeSneyders:
> Perhaps we can keep it as it originally was and just return a warning if the resulting number of partitions are lower than the workers? Wdyt?

Sounds good :+1:

RobbeSneyders (approved):

PR #479, opened 2023-10-02
Add caching dependency

PhilippeMoussalli:
Current implemented caching does not take into account the previous component cache key when estimating the cache key. This can be an issue in some scenarios, for example: 

 **1st run:** load_from_hub (input_dataset=x) -> download_images
 **2nd run:** load_from_hub (input_dataset=y) -> download_images
 
 In the second run, the download images component won't execute and will be cached since it might have the same input arguments as the first run. However, both runs start from different input datasets and thus the second run would actually end up loading in the dataset cached from the 1st run. 

This PR addresses this by including the hash key of the dependent component in the component cache key estimation.

### DIFF
--- a/src/fondant/compiler.py
+++ b/src/fondant/compiler.py
@@ -104,7 +104,7 @@ class DockerCompiler(Compiler):
             path = base_path
         return path, volume
 
-    def _generate_spec(  # noqa: PLR0912
+    def _generate_spec(
         self,
         pipeline: Pipeline,
         *,
@@ -126,12 +126,9 @@ class DockerCompiler(Compiler):
         for component_name, component in pipeline._graph.items():
             component_op = component["fondant_component_op"]
 
-            if component_cache_key is None:
-                component_cache_key = component_op.get_component_cache_key()
-            else:
-                component_cache_key = component_op.get_component_cache_key(
-                    previous_component_cache=component_cache_key,
-                )
+            component_cache_key = component_op.get_component_cache_key(
+                previous_component_cache=component_cache_key,
+            )
 
             metadata = Metadata(
                 pipeline_name=pipeline.name,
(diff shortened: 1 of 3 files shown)


RobbeSneyders:
Are we using the `input_manifest_path` to calculate the hash key as well? I assumed we were, and I would expect it to change whenever the hash key of the previous component changes.

PhilippeMoussalli:
> Are we using the `input_manifest_path` to calculate the hash key as well? I assumed we were, and I would expect it to change whenever the hash key of the previous component changes.

We're not using it. The path of the manifest does not change depending on the hash key and is always resolved to a path that contains the run id the current running pipeline. 

The save path looks like this at the moment

```
save_path_base_path = (
    f"{manifest.base_path}/{manifest.pipeline_name}/{manifest.run_id}/"
    f"{manifest.component_id}/manifest.json"
```

src/fondant/compiler.py
```diff
@@ -121,15 +121,24 @@ def _generate_spec(
 
         pipeline.validate(run_id=run_id)
 
+        component_cache_key = None
+
         for component_name, component in pipeline._graph.items():
             component_op = component["fondant_component_op"]
 
+            if component_cache_key is None:
```
RobbeSneyders, line 129: I think we can remove the `if/else` check here, since we check it inside `get_component_cache_key()`

RobbeSneyders (commented):
Ok, thanks for the explanation. I had some misassumptions about how the caching works, but this seems logical. Small comment below.

src/fondant/compiler.py
PhilippeMoussalli: ahh yes, good catch

RobbeSneyders (approved):

PR #484, opened 2023-10-04
Generate READMEs for all components using a script

RobbeSneyders:
This PR adds a script to generate a component README from its `fondant_component.yaml`.

As a next step, I'd like to add these to the documentation.


components/download_images/README.md
```diff
@@ -1,15 +1,67 @@
...
+        # Add arguments
+        "timeout": 10,
+        "n_connections": 100,
+        "image_size": 256,
+        "resize_mode": border,
+        "resize_only_if_bigger": False,
+        "max_aspect_ratio": inf,
+    }
+)
+pipeline.add_op(Download images_op, dependencies=[...])  #Add previous component as dependency
```
PhilippeMoussalli, line 63:
```suggestion
pipeline.add_op(download_images_op, dependencies=[...])  #Add previous component as dependency
```

PhilippeMoussalli (commented):
Nice :) This will save us a lot of time 

Few questions: 
1) The defaults are not show in the argument table. Is it worth adding them (perhaps explicitly mention that those are optional) and then not showing them in the **Usage** section? I feel like this will show better that default arguments don't need to be passed but might be less clear 
2) We mention now testing but we don't have tests for every component. Can we dynamically create this segment only if we have a test for the component? 
3) How is this script run? is it going to be part of CI/CD or should we do it manually every time we create a component?

components/filter_comments/README.md
```diff
@@ -0,0 +1,47 @@
...
+Component that filters code based on the code to comment ratio
+
+### Inputs/Outputs
+
+**The component comsumes:**
+- code
+  - content: string
+
+**The component produces:**
+
```
PhilippeMoussalli, line 13: Probably annoying to implement but can we mention that the component does not produce any subsets instead of keeping it blank?

mrchtr (approved):
Thanks @RobbeSneyders! I'm already a big fan of it.

PhilippeMoussalli (approved):
Thanks Robbe!

PR #489, opened 2023-10-05
Optimize GPU components

PhilippeMoussalli:
PR that modifies all current GPU components by:
* Batching both the preprocessing and inference to avoid OOM issues
* Disabling a pytorch API that caused **illegal memory access issues**. More on this issue [here](https://github.com/pytorch/pytorch/issues/99372)

### DIFF
--- a/components/caption_images/src/main.py
+++ b/components/caption_images/src/main.py
@@ -13,7 +13,6 @@ from transformers import BatchEncoding, BlipForConditionalGeneration, BlipProces
 
 logger = logging.getLogger(__name__)
 
-os.environ['CUDA_LAUNCH_BLOCKING'] = "1"
 os.environ['TORCH_CUDNN_V8_API_DISABLED'] = "1"
 
 
--- a/components/caption_images/src/main.py
+++ b/components/caption_images/src/main.py
@@ -1,6 +1,7 @@
 """This component that captions images using a model from the Hugging Face hub."""
 import io
 import logging
+import os
 import typing as t
 
 import numpy as np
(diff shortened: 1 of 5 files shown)


examples/pipelines/datacomp/components/mask_images/src/main.py
```diff
@@ -21,11 +21,13 @@ def mask_image(image_data, boxes):
             x1, y1, x2, y2 = tuple(box)
             # crop image
             cropped_image = np.array(image)[y1:y2, x1:x2]
-            # compute average color within cropped image
-            avg_color_per_row = np.average(cropped_image, axis=0)
-            avg_color = np.average(avg_color_per_row, axis=0)
-            r, g, b = tuple(avg_color)
-            draw.rectangle(((x1, y1), (x2, y2)), fill=(int(r), int(g), int(b)))
+
+            if cropped_image.any():
```
PhilippeMoussalli, line 25: This is not really related to this PR but is needed to tackle edge cases

components/caption_images/src/main.py
```diff
@@ -12,75 +13,97 @@
 
 logger = logging.getLogger(__name__)
 
+os.environ['CUDA_LAUNCH_BLOCKING'] = "1"
```
RobbeSneyders, line 16: We no longer need this, right? It was just for debugging I think.
Same in the other components.

RobbeSneyders (commented):
Thanks @PhilippeMoussalli! We might want to create a model inference component in the future which packages the batching functionality, so users only need to implement their code per batch.

components/caption_images/src/main.py
PhilippeMoussalli: Kept it in case we run into other issues later on and we need further debugging. We can remove it later on once we've tested enough GPU components at scale and are sure that everything runs fine

RobbeSneyders: I don't think it helped us since are already running single-threaded. It also didn't change the stacktrace, the first one was already correct. And it seems like it [really should not be used when not debugging](https://forums.developer.nvidia.com/t/cuda-graph-api-fail-when-cuda-launch-blocking-1-is-set/265619/2).

PhilippeMoussalli: oh ok, I'll remove it then

RobbeSneyders (approved):
Thanks!

PR #508, opened 2023-10-10
Update commoncrawl pipeline

mrchtr:
Update the commoncrawl pipeline to be compatible with kfp v2.


RobbeSneyders (commented):
Thanks @mrchtr!

RobbeSneyders:
I updated all components as part of #477, including these ones.

PR #510, opened 2023-10-11
Remove abstract component test

### LINKED ISSUE #367: Remove Abstract test class and update tests
In https://github.com/ml6team/fondant/pull/335 we decided to no longer use any abstraction for testing as it's straightforward to explicitly call the `transform` method of a component. We should remove the abstract test class and update the tests that use it.

@mrchtr agree?

mrchtr:
Removing abstract component test and update image caption component test.

Resolves #367


RobbeSneyders (approved):
Thanks @mrchtr!

PR #515, opened 2023-10-12
Bugfix input partition rows

mrchtr:
Currently, pipeline executions without defined input partition rows will fail. Due to the migration to KfP v2, the default value of `input_partition_rows` wasn't correctly overwritten. We need to set the default to -1 in the executor as well.


RobbeSneyders:
Thanks @mrchtr! But I think this will be fixed in #512

mrchtr:
> Thanks @mrchtr! But I think this will be fixed in #512

You are right! Going to close this one again.

PR #519, opened 2023-10-13
Vertex cli

PhilippeMoussalli:
PR that adds functionalities for compiling and running Vertex Pipelines with the CLI. 

After this PR is merged, i'd like to simplify a bit the compilation for both vertex and compile since there is a lot of repetition #517

### DIFF
--- a/src/fondant/cli.py
+++ b/src/fondant/cli.py
@@ -25,14 +25,14 @@ import typing as t
 from collections import defaultdict
 from types import ModuleType
 
-from fondant.compiler import DockerCompiler, KubeFlowCompiler
+from fondant.compiler import DockerCompiler, KubeFlowCompiler, VertexCompiler
 from fondant.component import BaseComponent, Component
 from fondant.executor import ExecutorFactory
 from fondant.explorer import (
     run_explorer_app,
 )
 from fondant.pipeline import Pipeline
-from fondant.runner import DockerRunner, KubeflowRunner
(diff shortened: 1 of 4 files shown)


src/fondant/cli.py
```diff
@@ -335,6 +365,38 @@ def register_run(parent_parser):
...
+            """,
+        action="store",
+    )
+    vertex_parser.add_argument(
+        "--project-id",
+        help="""The project id of the GCP project used to submit the pipeline""",
+    )
+    vertex_parser.add_argument(
+        "--project-region",
+        help="The project region of the GCP project used to submit the pipeline ",
```
RobbeSneyders, line 382: This is on a per pipeline basis right? I tested it and it works. So can we update the help message to reflect hits? Now it seems like the region is tied to the project.

RobbeSneyders (commented):
Thanks @PhilippeMoussalli!

One small comment.

src/fondant/cli.py
PhilippeMoussalli: Indeed this is the region where the pipeline is run and can be different from the project region. Updated the description

RobbeSneyders:
Thanks! You need to update this line from `project_region` to `region` to fix the test:
https://github.com/ml6team/fondant/blob/1c385056efb8e83a192b485e56785d1baff88340/tests/test_runner.py#L89

RobbeSneyders (approved):

PR #520, opened 2023-10-13
Component/generate embeddings

tillwenke:


docs/components/custom_component.md
```diff
@@ -1,4 +1,4 @@
-# Creating custom components
+ git config pull.rebase false# Creating custom components
```
mrchtr, line 1: I think this shouldn't be part of this PR.

.pre-commit-config.yaml
```diff
@@ -56,13 +56,14 @@ repos:
           - types-jsonschema
           - types-PyYAML
           - types-requests
+          - types-retry
         pass_filenames: false
 
-#  - repo: local
```
mrchtr, line 61: Could you exclude this changes from your PR as well?

components/generate_embeddings/src/main.py
```diff
@@ -0,0 +1,53 @@
...
+        self,
+        *_,
+        model: str,
+        model_provider: str,
+    ):
+        self.model = model
+        self.model_provider = model_provider
+
+        # to pass embedding model API keys via env vars
+        secrets_to_env_vars()
```
mrchtr, line 28: For now I think the easiest way would be to pass the secret as component argument to the component. 
The goal of the reusable components is that you can use it out of the box. This wouldn't work here, if a user has to adapt the code inside of the component.

mrchtr (commented):
Thanks @tillwenke. I've left a few comments.

Would you up for adding a simple test case to your component? 

Based on your component code I can not see why your pipeline is failing. I think we have to look on your whole pipeline to see what is going wrong.

tillwenke:
> Thanks @tillwenke. I've left a few comments.
> 
> Would you up for adding a simple test case to your component?
> 
> Based on your component code I can not see why your pipeline is failing. I think we have to look on your whole pipeline to see what is going wrong.

done

components/generate_embeddings/src/main.py
```diff
@@ -0,0 +1,53 @@
...
+        self,
+        *_,
+        model_provider: str,
+        model: str,
+        api_keys: dict,
+    ):
+        self.model_provider = model_provider
+        self.model = model
+
+        to_env_vars(api_keys)
```
mrchtr, line 28: Could you extend the readme on how to use the different model providers? E.g. in how the api key environment variable should be named. I can assume that this differs from provider to provider.

components/generate_embeddings/src/main.py
```diff
@@ -0,0 +1,53 @@
...
+        if model_provider == "cohere":
+            return CohereEmbeddings(model=model)
+        if model_provider == "huggingface":
+            return HuggingFaceEmbeddings(model_name=model)
+        if model_provider == "openai":
+            return OpenAIEmbeddings(model=model)
+        msg = f"Unknown provider {model_provider}"
+        raise ValueError(msg)
+
+    @retry()  # make sure to keep trying even when api call limit is reached
```
PhilippeMoussalli, line 43: Does this keep retrying infinitely? Would be nice if this can be somehow be modified to log when a connection limit is reached.

components/generate_embeddings/tests/generate_embeddings_test.py
PhilippeMoussalli, line 1: Nice to see some tests :)

PR #521, opened 2023-10-13
Add IndexWeaviate component

tillwenke:
(could you add my gh user to the ml6 org so I can add my example pipeline using this component?)


.pre-commit-config.yaml
```diff
@@ -56,13 +56,14 @@ repos:
           - types-jsonschema
           - types-PyYAML
           - types-requests
+          - types-retry
         pass_filenames: false
 
-#  - repo: local
```
mrchtr, line 61: Could you exclude this file from this PR?

mrchtr (commented):
Thanks @tillwenke! I've added a few comments.

components/index_embeddings/fondant_component.yaml
```diff
@@ -0,0 +1,43 @@
...
+      corresponding to the model used to generate the embeddings. 
+      You can find the available modules in the weaviate documentation: https://weaviate.io/developers/weaviate/modules/retriever-vectorizer-modules. 
+      This can be used for smoother interaction with the weaviate instance after indexing e.g. using langchain similarity search. 
+      The module used also has to be enabled in your weaviate instance deployment. 
+      Can also be {"vectorizer": "none"} if you do not want to make use of this feature.
+    type: dict
+  overwrite:
+    description: Whether to overwrite/ re-create the existing weaviate class and its embeddings.
+    type: bool
+    default: 0
```
PhilippeMoussalli, line 43: I think it's better to set the default as a Boolean

components/index_embeddings/src/main.py
```diff
@@ -0,0 +1,69 @@
...
+        if self.overwrite:
+            self.client.schema.delete_class(self.class_name)
+            self.client.schema.create_class(self.class_obj)
+        elif not self.client.schema.exists(self.class_name):
+            self.client.schema.create_class(self.class_obj)
+        else:
+            # do not index
+            return dataframe
+
+        self.client.batch.configure(batch_size=100, dynamic=True, num_workers=2)
```
PhilippeMoussalli, line 54: Would be nice if we can also include some of those parameters in the spec as opposed to hardcoding them. You can set the current values as default

components/index_embeddings/src/main.py
```diff
@@ -0,0 +1,69 @@
...
+        with self.client.batch as batch:
+            for part in dataframe.partitions:
```
PhilippeMoussalli, line 56: instead of looping through each individual partition, can we use Dask's [`map_partitions()`](https://docs.dask.org/en/stable/generated/dask.dataframe.DataFrame.map_partitions.html) to write partitions in parallel. This would increase the speed quite significantly for large datasets. One caveat is to check if the weaviate DB can handle concurrent writes.

components/index_embeddings/src/main.py
```diff
@@ -0,0 +1,69 @@
...
+                        "passage": data["text_data"],
+                        "identifier": identifier,
+                    }
+                    batch.add_data_object(
+                        data_object=properties,
+                        class_name=self.class_name,
+                        vector=data["text_embedding"],
+                    )
+
+        return dataframe
```
PhilippeMoussalli, line 69: you don't need to return a dataframe for a `WriteComponent`

components/index_embeddings/fondant_component.yaml
tillwenke: can I just write: False? if I recall correctly this wasnt possible when i tried it as it reads it can only read it as a str or int

tillwenke: that is why I also cound not set a default for the vectorizer arg

PhilippeMoussalli: I think if you're on the latest branch it should work, we recently fixed something similar related to arguments. Could you try it and let me know if you run into other issues?

components/index_embeddings/src/main.py
RobbeSneyders: This seems to introduce problems due to both `Dask` and `Weaviate` trying to spawn processes / threads.

RobbeSneyders:
Thanks @tillwenke, great work!

I pushed a commit to address my own feedback directly as there is some time pressure on getting this merged :)

Still need to test in the end to end pipeline once the other components are merged, but otherwise LGTM.

RobbeSneyders:
I was able to run the pipeline end-to-end both with the local and Vertex runner. I pushed the necessary changes. The pipeline will fail due to the component READMEs, so let's ignore that. I'll fix them in a separate PR.

RobbeSneyders (approved):

PR #522, opened 2023-10-16
Refactor compile method for kfp and vertex

PhilippeMoussalli:
PR that refactors the compiler module:
* `set_configuration()` is added as an abstract method since it's implemented in all of the compilers
* Since kfp and vertex have the same compile method, the `compile()` method is now defined in the kfp compiler and the vertex compiler is now it's child class. Other methods like `resolve_imports()` and `set_configuration` are overridden

### DIFF
--- a/src/fondant/compiler.py
+++ b/src/fondant/compiler.py
@@ -28,7 +28,7 @@ class Compiler(ABC):
         """Abstract method to invoke compilation."""
 
     @abstractmethod
-    def set_configuration(self, *args, **kwargs) -> None:
+    def _set_configuration(self, *args, **kwargs) -> None:
         """Abstract method to set pipeline configuration."""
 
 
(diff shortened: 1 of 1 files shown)


src/fondant/compiler.py
```diff
@@ -27,6 +27,10 @@ class Compiler(ABC):
     def compile(self, *args, **kwargs) -> None:
         """Abstract method to invoke compilation."""
 
+    @abstractmethod
+    def set_configuration(self, *args, **kwargs) -> None:
```
RobbeSneyders, line 31: Any reason to make this a public method?

RobbeSneyders (commented):
Thanks @PhilippeMoussalli!

src/fondant/compiler.py
PhilippeMoussalli: Not needed indeed, will revert it

RobbeSneyders (approved):

PR #523, opened 2023-10-16
Revert fsspec changes

mrchtr:
Due to a bug in fsspec, we added `auto_mkdir=True` to the fsspec calls. This is no longer necessary. This PR reverts the changes and avoids using the version with the bug.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,7 +47,7 @@ importlib-resources = { version = ">= 1.3", python = "<3.9" }
 jsonschema = ">= 4.18"
 pyarrow = ">= 11.0.0"
 
-fsspec = { version = ">= 2023.4.0", optional = true}
+fsspec = { version = ">= 2023.4.0, !=2023.9.1", optional = true}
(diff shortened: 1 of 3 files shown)


RobbeSneyders:
Agree with @PhilippeMoussalli, but I do think we should remove it for reading. Doesn't really make sense to create the directories there.

mrchtr:
I've updated the PR. Now the `auto_mkdir` has only been removed from the reading operations.

PhilippeMoussalli (approved):
Thanks Matthias!

pyproject.toml
```diff
@@ -47,7 +47,7 @@ importlib-resources = { version = ">= 1.3", python = "<3.9" }
 jsonschema = ">= 4.18"
 pyarrow = ">= 11.0.0"
 
-fsspec = { version = ">= 2023.4.0", optional = true}
+fsspec = { version = ">= 2023.4.0, !=2023.9.1", optional = true}
```
RobbeSneyders, line 50: I think we no longer need this?

PR #524, opened 2023-10-16
Modify arg default

PhilippeMoussalli:


src/fondant/component_spec.py
```diff
@@ -191,14 +192,22 @@ def outputs_additional_subsets(self) -> bool:
 
     @property
     def args(self) -> t.Mapping[str, Argument]:
+        def _get_default(argument_info):
+            if "default" in argument_info:
+                if argument_info["default"] == "None":
```
PhilippeMoussalli, line 197: This is needed for arguments that are taken from the componentSpec which always have a string type

src/fondant/compiler.py
```diff
@@ -289,6 +290,30 @@ def compile(
         pipeline.validate(run_id=run_id)
         logger.info(f"Compiling {pipeline.name} to {output_path}")
 
+        def set_component_exec_args(
```
PhilippeMoussalli, line 293: This was previously defined in the ComponentSpec but in fact should be evaluated at runtime. Optional arguments with defaults should not be included in the arg section of the componentOp otherwise some errors occur 

```python
I1016 10:42:19.355692      27 cache.go:116] Connecting to cache endpoint 10.0.12.128:8887
[2023-10-16 10:42:21,330 | fondant.cli | INFO] Component `LoadFromHubComponent` found in module main
usage: fondant [-h] [--component_spec COMPONENT_SPEC] [--cache CACHE]
               [--input_partition_rows INPUT_PARTITION_ROWS]
               [--cluster_type CLUSTER_TYPE] [--client_kwargs CLIENT_KWARGS]
fondant: error: argument --input_partition_rows: invalid int value: "{{$.inputs.parameters['input_partition_rows']}}"
```

src/fondant/component_spec.py
```diff
@@ -34,7 +34,7 @@ class Argument:
     name: str
     description: str
     type: str
-    default: t.Any = None
+    default: t.Optional[t.Any] = None
```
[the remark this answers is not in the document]
PhilippeMoussalli: ahh good to know

PhilippeMoussalli:
> Thanks @PhilippeMoussalli! The pipeline that was failing before now succeeds.

Glad it works :) I hope everything is fine now with the argument but some things might still pop up

RobbeSneyders (approved):

PR #526, opened 2023-10-17
Update vertex and general docs

PhilippeMoussalli:
PR to add documentation Vertex + slight modification on removing outdated docs and some restructuring


README.md
```diff
@@ -308,7 +308,7 @@ speed up your data preparation work.
 - Data lineage and experiment tracking
 - Distributed execution, both on and off cluster
 - Support other dataframe libraries such as HF Datasets, Polars, Spark
-- Move reusable components into a decentralized component registry
```
RobbeSneyders, line 311: why was this removed? This is still on the roadmap :)

docs/pipeline.md
```diff
@@ -32,115 +33,171 @@ def build_pipeline():
...
 
+!!! note "IMPORTANT"
+  When using other runners you will need to make sure that your new environment has access to:
+  - The base path of your pipeline (as mentioned above)
+  - The images used in your pipeline (make sure you have access to the registries where the images are stored)
 
+### Local Runner
 
-## Setting Custom partitioning parameters
+This local runner is mainly aimed at local development and quick iterations, there is no scaling so using small slices of your data is advised.
```
RobbeSneyders, line 69: It does scale, right. Especially on a larger VM. I would mention that it only scales to the machine you're running on. And that switching to Vertex / KfP has advantages that you can choose hardware per component, get better monitoring, reproducibility, etc.

docs/pipeline.md
```diff
@@ -32,115 +33,171 @@ def build_pipeline():
...
 
-**1) Repartitioning the Loaded DataFrame:** This step is optional and comes into play if the number
-of partitions is fewer than the available workers on the data processing instance.
-By repartitioning, the maximum number of workers can be efficiently utilized, leading to faster
-and parallel processing.
+Note that for components that do not come from the registry (local custom components) the compiler will add a build subsection (see component_3 in the example above) instead of referring to the image specified in the `component_spec.yaml`. This allows docker-compose to build and rebuild the container used in that component allowing for quicker iteration.
 
-**2) Repartitioning the Written DataFrame:** The written dataframe is also repartitioned into
-smaller sizes (default 250MB) to enable the next component to load these partitions into memory.
+In order to compile your pipeline to a `docker-compose` spec you need to import the `DockerCompiler`
```
RobbeSneyders, line 95: Should we document the Python API here? I would focus on using the CLI, and including it here might make it unclear if they need to add this to their code somewhere.

RobbeSneyders: I would even just document the `run` command, not the `compile` command.

docs/infrastructure.md
RobbeSneyders, line 1: Can we update this page to mention Vertex and the LocalRunner at the top? And remove that Fondant is built on KfP :)

docs/pipeline.md
RobbeSneyders: Same for the other runners below.

RobbeSneyders (commented):
Thanks @PhilippeMoussalli, nice improvements!

README.md
PhilippeMoussalli: Wasn't this the docker hub?

docs/infrastructure.md
PhilippeMoussalli: Maybe this page should just be specific on setting up the infra required for KFP. Since Vertex and docker don't really require an infrastructure, I don't see what we can add here. All the info on setting them up is in the pipeline page

RobbeSneyders (approved):
Thanks!

PR #527, opened 2023-10-17
Add fondant build command

### LINKED ISSUE #249: Add a `fondant build` command
We currently have a `build_images.sh` script for each example pipeline. If we implement https://github.com/ml6team/fondant/issues/245, we no longer need this for fondant development, but users might still need this to build their images before running on kubeflow.

It might make sense to implement a `fondant build` command that builds the images instead. For the local runner this is quite straightforward. The biggest challenge will probably be pushing to remote repositories and handling the accompanying credentials.

RobbeSneyders:
Fixes #249 

Please test this on your local machine when reviewing :)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,6 +51,7 @@ fsspec = { version = ">= 2023.4.0", optional = true}
 gcsfs = { version = ">= 2023.4.0", optional = true }
 s3fs = { version = ">= 2023.4.0", optional = true }
 adlfs = { version = ">= 2023.4.0", optional = true }
+docker = {version = ">= 6.1.3", optional = true }
 kfp = { version = "2.3.0", optional = true, extras =["kubernetes"] }
 pandas = { version = ">= 1.3.5", optional = true }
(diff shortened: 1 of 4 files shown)


PhilippeMoussalli (approved):
Thanks Robbe :) Tested it out and it seems to work. 

I am not 100% about this but I did notice that it takes slightly longer to build with the docker python SDK compared to the CLI (I might be wrong though). Do you know if there are differences between the two? 

Can we also add some tests and docs? can be later to prioritize the CC work but maybe let's add a ticket for it

RobbeSneyders:
> Thanks Robbe :) Tested it out and it seems to work.

Great! I'll ask Jan if he can test on Windows as well :)

> I am not 100% about this but I did notice that it takes slightly longer to build with the docker python SDK compared to the CLI (I might be wrong though). Do you know if there are differences between the two?

There should not really be a difference, as they both connect to the docker engine which does the actual building.

> Can we also add some tests and docs? can be later to prioritize the CC work but maybe let's add a ticket for it

Sounds good :+1:

PR #528, opened 2023-10-17
Chunker component

PhilippeMoussalli:
PR that adds a basic chunking component that chunks text based on length and overlap as well as the example rag pipeline

Taken from https://www.pinecone.io/learn/chunking-strategies/
Different chunking strategies are referenced there as well, might be interesting to reference them in the CC demo for people to try to implement 

The starting dataset is arbitrary and can be changed later on, I just needed something for testing


components/chunk_text/fondant_component.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+image: fndnt/chunk_text:dev
+
+consumes:
+  text:
+    fields:
+      data:
+        type: string
+
+produces:
```
mrchtr, line 16: What do you think about keeping a reference to the original document? E.g. doc_1 is chunked to doc_1_1, doc_1_2, ...
I can assume in some use case setup it is useful to find the original document of the chunk. 

We could also keep the text as it is and add an additional column `chunks` which will be a list of chunks.

PhilippeMoussalli: > What do you think about keeping a reference to the original document? E.g. doc_1 is chunked to doc_1_1, doc_1_2, ... I can assume in some use case setup it is useful to find the original document of the chunk.
> 
That's the currently the case since I am taking the original id of the document and appending the chunk number to it [here](https://github.com/ml6team/fondant/blob/d6b5480aba0489a160e35d7c71164d5cd29da645/components/chunk_text/src/main.py#L25)

> We could also keep the text as it is and add an additional column `chunks` which will be a list of chunks.

I think that's a good alternative, although just thinking about the next component (embedding) i'd rather have an embedding per row rather than a list of embeddings. I think it better matches also how things will be stored in the vector database.

RobbeSneyders: Agree with having one row per chunk. Adding `original_document_id` as a column makes sens I think though.

components/chunk_text/src/main.py
```diff
@@ -0,0 +1,68 @@
...
+import typing as t
+
+import pandas as pd
+from fondant.component import PandasTransformComponent
+from langchain.text_splitter import RecursiveCharacterTextSplitter
+
+logger = logging.getLogger(__name__)
+
+
+def chunk_text(row, text_splitter: RecursiveCharacterTextSplitter) -> t.List[t.Tuple]:
```
RobbeSneyders, line 19: I would move this function into the `Component`, so we can just access the text splitter as `self.text_splitter` instead of passing it around.

RobbeSneyders (commented):
Thanks @PhilippeMoussalli! Looks good in general, left 2 comments.

For the testing, I think we need to get more used to writing tests for the `transform` method so we can do unit testing instead of integration testing.

components/chunk_text/fondant_component.yaml
PhilippeMoussalli: yes might be better to link back to the original document, added!

PhilippeMoussalli:
> Thanks @PhilippeMoussalli! Looks good in general, left 2 comments.
> 
> For the testing, I think we need to get more used to writing tests for the `transform` method so we can do unit testing instead of integration testing.

I agree :) added some tests

RobbeSneyders (approved):
Thanks! Looks great!

PR #529, opened 2023-10-17
Add resource limits for Vertex

RobbeSneyders:
This was necessary to be able to allocate enough memory. Added the missing ones to the KfP compiler as well.

Proof that it works: https://console.cloud.google.com/vertex-ai/locations/europe-west1/pipelines/runs/datacomp-filtering-pipeline-20231017201538?project=soy-audio-379412

### DIFF
--- a/src/fondant/compiler.py
+++ b/src/fondant/compiler.py
@@ -407,10 +407,16 @@ class KubeFlowCompiler(Compiler):
         accelerator_name = fondant_component_operation.accelerator_name
         node_pool_label = fondant_component_operation.node_pool_label
         node_pool_name = fondant_component_operation.node_pool_name
+        cpu_request = fondant_component_operation.cpu_request
+        cpu_limit = fondant_component_operation.cpu_limit
         memory_request = fondant_component_operation.memory_request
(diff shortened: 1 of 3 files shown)


src/fondant/compiler.py
```diff
@@ -459,10 +465,16 @@ def resolve_imports(self):
     @staticmethod
     def _set_configuration(task, fondant_component_operation):
         # Unpack optional specifications
+        cpu_limit = fondant_component_operation.cpu_limit
+        memory_limit = fondant_component_operation.memory_limit
         number_of_accelerators = fondant_component_operation.number_of_accelerators
         accelerator_name = fondant_component_operation.accelerator_name
 
         # Assign optional specification
+        if cpu_limit is not None:
```
PhilippeMoussalli, line 474: should we also add the cpu/memory request for Vertex?

PhilippeMoussalli (approved):

src/fondant/compiler.py
RobbeSneyders: According to the [docs](https://cloud.google.com/vertex-ai/docs/pipelines/machine-types), it only supports limits.

PR #531, opened 2023-10-18
Fix explorer build script for DockerHub

RobbeSneyders:
The release pipeline fails on building the explorer script. It was not properly updated for our switch to DockerHub.

Works now: https://hub.docker.com/r/fndnt/data_explorer

### DIFF
--- a/scripts/build_explorer.sh
+++ b/scripts/build_explorer.sh
@@ -23,8 +23,8 @@ if [ -z "${tag}" ]; then
 fi
 
 # Set github repo information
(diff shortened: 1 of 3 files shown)


PhilippeMoussalli (approved):
Thanks Robbe, I see that the default namespace in `tag_components.sh` and `tag_explorer.sh` is still set to `ml6team`. Can we also change them to `fndnt`?

RobbeSneyders:
Good catch, updated and tested (the explorer one).

PR #537, opened 2023-10-18
Add network argument to vertex runner

RobbeSneyders:
This is needed to access private IPs from Vertex pipelines (eg. a private Weaviate endpoint).


PhilippeMoussalli (approved):
Thanks Robbe! :) 

There is one CLI test that is failing though (`test_vertex_run()`). I think you need to ad 
`network=None` at the argparse and namespace level

PR #538, opened 2023-10-18
Update component readmes

RobbeSneyders:
This should fix the pre-commit pipeline.


PhilippeMoussalli (approved):
Thanks Robbe! I think later on we can think about removing some of the internal arguments that are either defined in the component spec or used internally [#539](https://github.com/ml6team/fondant/issues/539)

RobbeSneyders:
I didn't even notice that they were added :sleeping: 

Already fixed.
Fixes #539

Issue #518, opened 2023-10-13
Add componentOp warning for unused set configurations


PhilippeMoussalli:

RobbeSneyders:
Can you explain what you mean with this @PhilippeMoussalli? Agree with taking any runner specific code out of the `ComponentOp` class, but I wouldn't want the user to have to change which `ComponentOp` class they use when submitting to a different runner.

PhilippeMoussalli:
Perhaps we can return proper warnings if a certain field in the `componentOp` is set but will not be used in the specified runner (e.g. `node_pool_name` for local runner)

PR #551, opened 2023-10-24
Add componentOp warning for unused configuration

PhilippeMoussalli:
Addresses #518 

Separating the ComponentOp for different runners is not ideal since it would require changing the class names when switching runners. Instead, it's best to return warning logs if a certain configuration is set but not valid for a certain runner (e.g. nodepool for docker runner). This PR modifies the `set_configuration` method of the Compiler to handle this.


RobbeSneyders (commented):
Thanks @PhilippeMoussalli!

Currently this is a bit hard-coded though. I would propose to group the infrastructure configuration in a map in the `ComponentOp` class. That way we can just iterate over it dynamically.

PhilippeMoussalli:
> Thanks @PhilippeMoussalli!
> 
> Currently this is a bit hard-coded though. I would propose to group the infrastructure configuration in a map in the `ComponentOp` class. That way we can just iterate over it dynamically.

Can you elaborate a bit more? we still need to know if a certain config applies to a certain compiler/runner. Would we then have separate configuration maps in the `ComponentOp` for each runner?

RobbeSneyders:
The issue is that we currently would have to update the `DockerCompiler` if we add any Vertex config options. 

I meant something like this:

```python
component_op = ComponentOp.from_registry(
    name=...,
    args={...},
    resources={
        "memory_request": "16G",
        "cpu_request": 8, 
    }
)
```

We can also make it a class, so users don't have to use python strings and we can validate the values easily:

```python
    resources=Resources(
        memory_request="16G",
        cpu_request=8,
    )
```

Separate resource maps per runner would only make sense if we accept multiple, so the user doesn't need to change the code if they want to move from local to a remote runner. Eg:

```python
    resources=[
        DockerResources(
            memory_request="16G",
            cpu_request=8,
        ),
        KubeflowResources(
            node_pool="my-nodepool",
        ),
    ]
```

Not sure if that's a good idea.

PhilippeMoussalli:
> The issue is that we currently would have to update the `DockerCompiler` if we add any Vertex config options.
> 
> I meant something like this:
> 
> ```python
> component_op = ComponentOp.from_registry(
>     name=...,
>     args={...},
>     resources={
>         "memory_request": "16G",
>         "cpu_request": 8, 
>     }
> )
> ```
> 
> We can also make it a class, so users don't have to use python strings and we can validate the values easily:
> 
> ```python
>     resources=Resources(
>         memory_request="16G",
>         cpu_request=8,
>     )
> ```
> 
> Separate resource maps per runner would only make sense if we accept multiple, so the user doesn't need to change the code if they want to move from local to a remote runner. Eg:
> 
> ```python
>     resources=[
>         DockerResources(
>             memory_request="16G",
>             cpu_request=8,
>         ),
>         KubeflowResources(
>             node_pool="my-nodepool",
>         ),
>     ]
> ```
> 
> Not sure if that's a good idea.

I actually like that idea, make it a bit more explicit which resources belong to which runner (typehinting) and it doesn't require changing the resources if you switch runner (provided that you've already added the corresponding resource). I will implement it

RobbeSneyders:
Let me just give one counter argument to my own proposal before we accept it :stuck_out_tongue: 

I think `cpu_request` and `memory_request` are generic for each runner. Even when running on Kubeflow, I can imagine a lot of users will just define the resources they need instead of a specific node pool. So if my assumption is correct, a single resource map will look like this:

```python
    resources=Resources(
        memory_request="16G",
        cpu_request=8,
    )
```

While a resource map per runner will look like this:

```python
    resources=[
        DockerResources(
            memory_request="16G",
            cpu_request=8,
        ),
        KubeflowResources(
            memory_request="16G",
            cpu_request=8,
        ),
    ]
```

So this could be nicer for the simple option, while less nice for the specialized option (eg. when defining `node_pool`).

I don't have a strong preference, just wanted to highlight all the options I see with their advantages and downsides.

PhilippeMoussalli:
> ```python
>     resources=Resources(
>         memory_request="16G",
>         cpu_request=8,
>     )
> ```

Can you explain again how the first option can help avoiding the problem you mentioned:

`The issue is that we currently would have to update the DockerCompiler if we add any Vertex config options.`

if I decide to add a `custom_vertex_resource` at some point 

  resources=Resources(
      memory_request="16G",
      cpu_request=8,
      custom_vertex_resource: "resource"
  )

I would still have to check if that value is set during compilation for the other compilers and return a corresponding error message

RobbeSneyders:
Good point. With a `dict` resource map, we could pop the items we use and check for any remaining issues. But I guess this becomes more complex again with a single class.

PhilippeMoussalli:
> Good point. With a `dict` resource map, we could pop the items we use and check for any remaining issues. But I guess this becomes more complex again with a single class.

Thinking about it again I think I'd opt for the single class approach, we're able to catch false settings (gpu mainly) and other things at compile time if they are set incorrectly. Let me know what you think, then I can update the docs and pipelines

RobbeSneyders (approved):
Thanks @PhilippeMoussalli, this looks good to me.

PhilippeMoussalli:
Thanks! Update the docs, I won't update the pipelines here since we plan on removing them. We should remember to change them in the examples repos if we bump the fondant version to a new release that includes this change

PR #552, opened 2023-10-25
Remove Starcoder pipeline

PhilippeMoussalli:
PR that removes the starcoder pipelines and it's dependent components


RobbeSneyders (approved):
Thanks @PhilippeMoussalli!

I think this makes sense since it's not up to date and the components are too specific to provide a lot of value.

PR #555, opened 2023-10-26
Add screenshots data explorer

### LINKED ISSUE #496: Add screenshots to the data explorer documentation
It would be great to visually show what users can expect.

mrchtr:
fix #496


PhilippeMoussalli (approved):
Nice! Thanks Matthias

docs/data_explorer.md
```diff
@@ -24,13 +24,20 @@ The user can specify a pipeline and a specific pipeline run and component to exp
 
 The chosen subset (and the columns within the subset) can be explored in 3 tabs.
 
+![data explorer](https://github.com/ml6team/fondant/blob/main/docs/art/data_explorer/data_explorers.png?raw=true)
```
RobbeSneyders, line 27:
```suggestion
![data explorer](https://github.com/ml6team/fondant/blob/main/docs/art/data_explorer/data_explorer.png?raw=true)
```

RobbeSneyders (changes requested):
Thanks @mrchtr!

One small typo in one of the urls. Also, I think we can use relative urls here, which should be a bit more robust.

Can you also move the `Data Explorer UI` section to the top and the `How to use?` to the bottom?

mrchtr:
> Also, I think we can use relative urls here, which should be a bit more robust.

Not sure if this will work. The images where broken some weeks ago. Then we started to replace the relative links with the urls to the images.

RobbeSneyders:
> Not sure if this will work. The images where broken some weeks ago. Then we started to replace the relative links with the urls to the images.

This should only be the case on the main page, since we insert the README there and the urls need to be relevant to different directories on the site vs Github.

RobbeSneyders (approved):

PR #586, opened 2023-10-31
Datacomp submission improvement

PhilippeMoussalli:
PR that updates the datacomp pipeline setting to be reusable. After this PR is merged, I will move the datacomp to a separate repo


examples/pipelines/datacomp/components/mask_images/Dockerfile
```diff
@@ -11,7 +11,7 @@ RUN pip3 install --no-cache-dir -r requirements.txt
 
 # Install Fondant
 # This is split from other requirements to leverage caching
-ARG FONDANT_VERSION=main
+ARG FONDANT_VERSION=0.6.2
```
mrchtr, line 14: Should we really change the Dockerfile? I think we can fix the version by using the fondant build command, right?

PhilippeMoussalli: right I just saw the build args command, I will revert it back

examples/pipelines/datacomp/components/filter_clip_score/fondant_component.yaml
```diff
@@ -1,6 +1,6 @@
 name: Filter CLIP score
 description: Component that filters out bad captions (Empty captions, Captions with weird characters, Captions that are dates)
-image: fndnt/filter_clip_score:dev
+image: fndnt/filter_clip_score:0.6.2
```
RobbeSneyders, line 3: We'll have to check if we still want to use the `fndnt` registry for the custom components in the examples.

RobbeSneyders (approved):

PR #587, opened 2023-10-31
Split component and pipeline SDKs

### LINKED ISSUE #554: Split component and pipeline SDK
Fondant currently implicitly provides two separate SDKs:

- A component SDK which helps to implement components and executes them:
  - Base component classes
  - Executors
  - Data IO

- A pipeline SDK which helps to implement pipelines and compiles and runs them:
  - Pipeline & ComponentOp
  - Compilers
  - Runners

And some core functionality which is used for both:
- ComponentSpec
- Manifest

Note that most of our dependencies are only used in one of these parts.

I believe we would benefit from making these more explicit. This could make fondant more understandable for users, and could make it easier for us to maintain independence between these SDKs in the code.

My proposal would be to split the code base into 3 [namespace packages](https://packaging.python.org/en/latest/guides/packaging-namespace-packages/):
- component
- pipeline
- core

Which we still distribute as one, but imported separately. We could still provide different extras to install only the dependencies needed for a specific SDK (eg. `fondant[pipeline]`)

RobbeSneyders:
Fixes #554 

This PR splits the component and pipeline SDKs. Some things I want to add in follow-up PRs:
- Revisit the install extras. Eg. Dask is not needed in the pipeline
- Automatically check that the pipeline and component SDKs don't import from each other


data_explorer/app/widgets.py
```diff
@@ -27,7 +28,7 @@ def build_sidebar(base_path: str) -> Tuple[Manifest, str, Dict[str, str]]:
         Tuple[Optional[str], Optional[str], Optional[Dict]: Tuple with manifest path,
         subset name and fields
     """
-    fs = get_filesystem(base_path)
+    fs, _ = fsspec.core.url_to_fs(base_path)
```
PhilippeMoussalli, line 31: nice find

src/fondant/cli.py
```diff
@@ -28,15 +28,9 @@
...
-from fondant.build import build_component
-from fondant.compiler import DockerCompiler, KubeFlowCompiler, VertexCompiler
-from fondant.component import BaseComponent, Component
-from fondant.executor import ExecutorFactory
-from fondant.explorer import (
-    run_explorer_app,
-)
-from fondant.pipeline import Pipeline
-from fondant.runner import DockerRunner, KubeflowRunner, VertexRunner
+if t.TYPE_CHECKING:
```
PhilippeMoussalli, line 31: what does this do?

PhilippeMoussalli (approved):
Thanks Robbe!

src/fondant/cli.py
RobbeSneyders: It only imports the guarded imports during type checking, not at runtime. See [docs](https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING).

GeorgesLorre (approved):
nice! Going to be clean

PR #590, opened 2023-11-03
Add generic filtering component

PhilippeMoussalli:
Draft PR for generic filtering component

Notes:
* Implemented as a `DaskTransform` Component, no need to implement another transform class called `filter` since there aren't any needed additional functionalities

Things that still need to be figured out: 
* Proper way of defining the `filtering_conditions`. Those are currently implemented as a list of tuples. KFP only excepts standard types -> need to figure out a proper way of serializing it. We might also want to properly present it to the user in a way that's easy for them to figure out what filtering function are available
* Passing custom filtering functions? is this needed or should we let the user implement their own functions

Out of the box ideas: 
* Do we need a separate component for this, can we have something define at the componentOp level to apply the filtering before/after loading/transforming the dataset? 

```python
component= ComponentOp(...).filter_before(<filtering_conditions>)
component= ComponentOp(...).filter_after(<filtering_conditions>)
```
After this is merged, we can replace all currently existing filtering components with that one.


components/filter_dataset/fondant_component.yaml
```diff
@@ -0,0 +1,15 @@
...
+  dummy_variable: #TODO: fill in here
+    fields:
+      data:
+        type: binary
+
+
+args:
+  filtering_conditions:
+    description: The filtering conditions to apply to the dataset
+    type: dict
```
PhilippeMoussalli, line 15: This is not final, might change after figuring out how to serialize it

RobbeSneyders:
Closing since this is outdated. Let's pick this up again in the future.

PR #592, opened 2023-11-06
Add component install extra and update others

RobbeSneyders:
Follow-up on #587

### DIFF
--- a/components/caption_images/Dockerfile
+++ b/components/caption_images/Dockerfile
@@ -12,7 +12,7 @@ RUN pip3 install --no-cache-dir -r requirements.txt
 # Install Fondant
 # This is split from other requirements to leverage caching
 ARG FONDANT_VERSION=main
-RUN pip3 install fondant[aws,azure,gcp]@git+https://github.com/ml6team/fondant@${FONDANT_VERSION}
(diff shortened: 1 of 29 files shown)


PhilippeMoussalli (commented):
Thanks Robbe. 

General question related to frameworks, I feel like for most frameworks out there the base installation offers the core functionalities of the framework and the extras enable additional functionalities. 

For us, it seem like the user cannot do anything with Fondant without installing any extras. They can now neither build/push components nor submit pipelines. Shouldn't we define some essential functionalities that are included in the base installation?

PhilippeMoussalli:
> You can run a pipeline on the local runner with the base installation.
> 
> The downside to including more dependencies in the base installation, is that they will be installed inside each component as well.
> 
> We could still simplify things by removing the `docker` extra and including it in both the `vertex` and `kfp` extras.

I see, I forgot that `docker` is only needed for pushing images:

1) In this case we probably don't need to include them in the `requirements.txt` of the example notebooks 
2) I would maybe still include then `docker` with all the runners since it can be used to publish images on all of them
3) Can we also raise errors to instruct the user to install the `component` extra if it's not installed (similar to docker and kfp/vertex)

PhilippeMoussalli (approved):
Thanks! :)

PR #594, opened 2023-11-07
Add simple pipeline and integration test for the LocalRunner

mrchtr:
This PR adds a simple sample pipeline to the tests and a integration test for the LocalRunner.
The pipeline consists of three components. I've used lightweight components without a GPU to speedup the execution of the integration.

Components of the pipeline:
- Load from file
- Custom dummy component (testing build of local components - returning the dataframe without any modifications)
- Chunk text component

After the pipeline ran, the test checks that the folders of the pipelines steps were created correctly. 

## Open question

- Actually, the custom component uses fondant from the main branch as dependency. If we a code change breaks something regarding to the build/usage of custom components, the integration test will not cover these. I think that should be fine under the assumption such breaking changes would be covered by existing unit tests.


tests/sample_pipeline_test/test_sample_pipeline.py
```diff
@@ -0,0 +1,88 @@
+# This file contains a sample pipeline. Loading data from a parquet file, using  load_from_parquet
```
GeorgesLorre, line 1:
```suggestion
# This file contains a sample pipeline. Loading data from a parquet file, using the load_from_parquet
```

tests/sample_pipeline_test/test_sample_pipeline.py
```diff
@@ -0,0 +1,88 @@
...
+    return pipeline
+
+
+def test_local_runner(sample_pipeline):
+    DockerCompiler().compile(
+        sample_pipeline,
+        output_path="docker-compose.yaml",
+        extra_volumes=[str(DATA_DIR.resolve()) + ":/data"],
+    )
+    DockerRunner().run("docker-compose.yaml")
```
GeorgesLorre, line 68: This works in a CI/CD pipeline ?? Cool

GeorgesLorre:
How long does it take to run this test ?

mrchtr:
> How long does it take to run this test ?

The bottleneck is the image pulling process. If the images are pulled it took around 30s. In the ci/cd where we always pulling the images it are ~90s. Should be fine for a single integration test. If we implement more tests this might be not fast enough. 
Not sure if we have to trigger the integrations test on every push to the branch.

Alternatively, if we refactor the test structure a bit (#595), we could execute the integration test in an extra step. Execute the unit test before we execute the integration test. This would make the pipeline failing fast and we will not enter the integrations tests every time.

tests/sample_pipeline_test/test_sample_pipeline.py
PhilippeMoussalli, line 1: Is it possible to move this file two levels above under 'test' so that it's at the same level as the other test files?

tests/sample_pipeline_test/test_sample_pipeline.py
```diff
@@ -0,0 +1,88 @@
...
+        str(DATA_DIR / "dummy-pipeline" / "dummy-pipeline-*" / "*"),
+    )
+
+    assert len(pipeline_dirs) == NUMBER_OF_COMPONENTS
+    for dir in pipeline_dirs:
+        assert os.path.exists(Path(dir) / "index")
+        assert os.path.exists(Path(dir) / "text")
+        assert os.path.exists(Path(dir) / "manifest.json")
+
+    try:
```
GeorgesLorre, line 82: you could work with temporary directories like we do in other tests:

https://github.com/ml6team/fondant/blob/main/tests/test_compiler.py#L131

tests/sample_pipeline_test/test_sample_pipeline.py
mrchtr: Indeed for now the better approach. I've updated the structure.

tests/sample_pipeline_test/test_sample_pipeline.py
mrchtr: Very useful. Thanks!

PhilippeMoussalli (approved):
Thanks Matthias! Nice addition

PR #596, opened 2023-11-08
Fixing pre-commit building component READMEs

mrchtr:
The pre-commit was failing due to missing dependencies.


RobbeSneyders:
Thanks @mrchtr, I think we should add PyYAML as a core dependency of Fondant though, since we use it explicitly in `fondant.core`

RobbeSneyders (approved):

PR #606, opened 2023-11-08
Fix pipeline label

PhilippeMoussalli:
Assign pipeline description label at the service level since labels can't be assigned at the root level


mrchtr (approved):
Thanks @PhilippeMoussalli

PR #607, opened 2023-11-08
Fix naming error

PhilippeMoussalli:
Bugfix that fixes the component name based on the expected format in different places (kfp, component folder, original name, ...)


mrchtr (approved):
Thanks @PhilippeMoussalli. Works as expected again.

PR #608, opened 2023-11-08
First POC of sagemaker compiler/runner

GeorgesLorre:
First draft of a SageMaker compiler/runner

I included a pipeline I used for testing as an example and it also has the spec.json which is generated by the SageMakerCompiler. (this example will be removed in the end result)

The main parts are in the compiler.py and runner.py


src/fondant/pipeline/runner.py
```diff
@@ -126,3 +126,32 @@ def get_name_from_spec(self, input_spec: str):
...
+        with open(input_spec) as f:
+            pipeline = f.read()
+            response = client.create_pipeline(
+                PipelineName="my-pipeline4",
+                PipelineDefinition=pipeline,
+                RoleArn="TODO",
+            )
+            print(response)
+
+    def get_name_from_spec(self, input_spec: str):
```
GeorgesLorre, line 153: needs to be implemented

src/fondant/pipeline/runner.py
```diff
@@ -126,3 +126,32 @@ def get_name_from_spec(self, input_spec: str):
...
+        self.__resolve_imports()
+
+    def __resolve_imports(self):
+        import boto3
+
+        self.boto3 = boto3
+
+    def run(self, input_spec: str, *args, **kwargs):
+        """Run a sagemaker pipeline."""
+        client = self.boto3.client("sagemaker")
```
GeorgesLorre, line 142: We need to use the lower level SDK to start a pipeline from a spec. There is no support for this in the python SDK

src/fondant/pipeline/runner.py
```diff
@@ -126,3 +126,32 @@ def get_name_from_spec(self, input_spec: str):
...
+
+        with open(input_spec) as f:
+            pipeline = f.read()
+            response = client.create_pipeline(
```
GeorgesLorre, line 146: will need to add logic to update/delete pipelines since names are unique

src/fondant/pipeline/compiler.py
```diff
@@ -487,3 +487,133 @@ def _set_configuration(task, fondant_component_operation):
...
+                            f"{path}/{metadata.pipeline_name}/{metadata.run_id}/"
+                            f"{dependency}/manifest.json",
+                        ],
+                    )
+
+                    depends_on = [steps[-1]]
+            else:
+                depends_on = []
+
+            script = self.generate_component_script(component_name, command)
```
GeorgesLorre, line 583: These files are not deleted but could be after compiling

src/fondant/pipeline/runner.py
```diff
@@ -126,3 +126,32 @@ def get_name_from_spec(self, input_spec: str):
...
             spec = yaml.safe_load(f)
             return spec["pipelineInfo"]["name"]
+
+
+class SagemakerRunner(Runner):
+    def __init__(self):
+        self.__resolve_imports()
+
+    def __resolve_imports(self):
+        import boto3
```
PhilippeMoussalli, line 136: does this need to be added to `pyproject.toml`? Can we do a try/except with a valid message on which extras to install?

src/fondant/pipeline/runner.py
PhilippeMoussalli, line 142: Maybe we can move the client to the `__init__` similar to the kfp runner

src/fondant/pipeline/runner.py
PhilippeMoussalli: I don't think this is an abstract method, you can omit this method and just read it directly from the spec in the `run` method

src/fondant/pipeline/runner.py
```diff
@@ -126,3 +126,32 @@ def get_name_from_spec(self, input_spec: str):
...
+        client = self.boto3.client("sagemaker")
+
+        with open(input_spec) as f:
+            pipeline = f.read()
+            response = client.create_pipeline(
+                PipelineName="my-pipeline4",
+                PipelineDefinition=pipeline,
+                RoleArn="TODO",
+            )
+            print(response)
```
PhilippeMoussalli, line 151: Probably for debugging, let's switch it to `logging` (debug or info depending on what you think is more suitable)

src/fondant/pipeline/compiler.py
```diff
@@ -487,3 +487,133 @@ def _set_configuration(task, fondant_component_operation):
...
+
+
+class SagemakerCompiler(Compiler):
+    def __init__(self):
+        self._resolve_imports()
+
+    def _resolve_imports(self):
+        try:
+            import sagemaker
+            import sagemaker.processing
```
PhilippeMoussalli, line 499: do we need all those extra imports for checking? I think `import sagemaker` might suffice

PhilippeMoussalli (commented):
Thanks Georges :) 

Looks really good, I thought it might deviate a lot from the current runners but seems clean!

src/fondant/pipeline/compiler.py
```diff
@@ -487,3 +487,133 @@ def _set_configuration(task, fondant_component_operation):
...
+                instance_type="ml.t3.medium",
+                instance_count=1,
+                base_job_name=component_name,
+                role=self.sagemaker.get_execution_role(),
+            )
+            step = self.sagemaker.workflow.steps.ProcessingStep(
+                name=component_name,
+                processor=processor,
+                depends_on=depends_on,
+                code=script,
```
PhilippeMoussalli, line 597: Just to better understand. Can you explain again how you resolved the issue of the long command? is it simply by passing the command to execute to `code` which has no limit? 

The code here is then the command to execute and the `command` in the `processor` is the executing command?

src/fondant/pipeline/compiler.py
PhilippeMoussalli: maybe we can create temp files in this function instead

src/fondant/pipeline/compiler.py
GeorgesLorre: The script is a an actual file we create per component, it contains the execute command and all arguments (which were the main offenders for the length limit).

I have included the script generated by this in the example pipeline in this PR (see `playground/some-cool-pipeline/artifacts/`)

we then point the ProcessingStep to this file and SageMaker will upload it to S3 and set it as input automatically you can see this in the resulting `playground/some-cool-pipeline/spec.json` 

This avoids the length issues but requires the right credentials during compile to upload these files

src/fondant/pipeline/runner.py
GeorgesLorre: Yes ! Will add it 

I guess its installed along with the sagemaker package but I will add it explicitly

src/fondant/pipeline/runner.py
GeorgesLorre: Just some copy paste leftovers from the other compilers

src/fondant/pipeline/compiler.py
PhilippeMoussalli: oh ok good to know

src/fondant/pipeline/compiler.py
PhilippeMoussalli: Thanks all clear now! Good that we don't need to add any custom logic at the executor

pyproject.toml
```diff
@@ -55,13 +55,15 @@ docker = {version = ">= 6.1.3", optional = true }
 kfp = { version = "2.3.0", optional = true, extras =["kubernetes"] }
 pandas = { version = ">= 1.3.5", optional = true }
 google-cloud-aiplatform =  { version = "1.34.0", optional = true}
+sagemaker = "^2.197.0"
```
RobbeSneyders, line 58: This should be marked as optional.

src/fondant/pipeline/compiler.py
RobbeSneyders: Can you test it to be sure? :)

src/fondant/pipeline/compiler.py
```diff
@@ -487,3 +487,133 @@ def _set_configuration(task, fondant_component_operation):
...
+                        [
+                            "--input_manifest_path",
+                            f"{path}/{metadata.pipeline_name}/{metadata.run_id}/"
+                            f"{dependency}/manifest.json",
+                        ],
+                    )
+
+                    depends_on = [steps[-1]]
+            else:
+                depends_on = []
```
RobbeSneyders, line 581: This looks very similar to the other compilers. Do you think it makes sense to combine this?

RobbeSneyders (commented):
Thanks @GeorgesLorre!

Looks to be going in the right direction!

src/fondant/pipeline/compiler.py
GeorgesLorre:
```
  File "/Users/georgeslorre/ML6/internal/express/src/fondant/pipeline/compiler.py", line 594, in compile
    step = self.sagemaker.workflow.steps.ProcessingStep(
AttributeError: module 'sagemaker.workflow' has no attribute 'steps'
```

when I comment out the 3 extra imports

GeorgesLorre:
Closing, will be introduced in more focused PR's

PR #609, opened 2023-11-08
Update hub

PhilippeMoussalli:
PR that updates the hub: 

* Added the `resize_images` component 
* Restructured based on the new component categorization in the readme of this PR https://github.com/ml6team/fondant/tree/user-journey-documentation


RobbeSneyders:
This file is automatically generated @PhilippeMoussalli. You can see that the changes are not reflected in the built site, because they get overwritten.

I think the only way to achieve this, is by marking each component with a tag. And then grouping them by tag in the generation script. This is something we'll need for the next version of the hub anyway.

PhilippeMoussalli:
> This file is automatically generated @PhilippeMoussalli. You can see that the changes are not reflected in the built site, because they get overwritten.
> 
> I think the only way to achieve this, is by marking each component with a tag. And then grouping them by tag in the generation script. This is something we'll need for the next version of the hub anyway.

Missed the fact that we had a script to generate this, I added an optional tag to the spec and modified the generation script.

RobbeSneyders:
Nice, this looks great.

If we're gonna store it in the `fondant_component.yaml`, we need to properly think about how we want to include this information, since it's a user facing interface (and one of our main ones).

I think the current one might be a bit rigid. Eg. where do you see a clip scoring component fit? I see a couple of options. In the hub mockups, most components also have multiple tags.

Some options I see:
- A `tags` field, which is a list of strings
- A general `annotations` field with arbitrary key/value pairs

PhilippeMoussalli:
> Nice, this looks great.
> 
> If we're gonna store it in the `fondant_component.yaml`, we need to properly think about how we want to include this information, since it's a user facing interface (and one of our main ones).
> 
> I think the current one might be a bit rigid. Eg. where do you see a clip scoring component fit? I see a couple of options. In the hub mockups, most components also have multiple tags.
> 
> Some options I see:
> 
> * A `tags` field, which is a list of strings
> * A general `annotations` field with arbitrary key/value pairs

Good point, I turned the tags into an array of strings and moved component type checks from the component spec schema to the generate hub script. This way, users building custom component won't face any issues and can insert any tag or omit them and the component type tags are only mandatory for components in the hub.

RobbeSneyders (approved):
Thanks!

Issue #599, opened 2023-11-08
Insert extensive guide on Docker installation


PhilippeMoussalli:

RobbeSneyders:
Documentation from Airflow: https://github.com/apache/airflow/blob/main/CONTRIBUTORS_QUICK_START.rst#docker-community-edition

PR #619, opened 2023-11-13
Expand docker installation guide

PhilippeMoussalli:
Addresses #599 
I think referencing the docker installation docs rather than pasting few snippets is better in our case since we don't require anything custom compared to airflow


docs/guides/installation.md
```diff
@@ -58,12 +58,25 @@ Check out the [guide](../components/publishing_components.md) on publishing comp
 
 To execute pipelines locally, you must
 have [Docker compose](https://docs.docker.com/compose/install/) and Python >=3.8
-installed on your system.
+installed on your system. We only support Docker compose V2 and above. If you have an older version of
+Docker compose, please upgrade to the latest version as described in the [Docker compose migration doc](https://docs.docker.com/compose/migrate/).
 
 #### TODO: Modify/extend considerations for Docker-compose/desktop
```
RobbeSneyders, line 64: Can we remove this TODO now?

PhilippeMoussalli: done!

PR #626, opened 2023-11-14
Use names instead of directory names in hub

RobbeSneyders:
Looks a bit nicer :)


PhilippeMoussalli (commented):
I agree :) might need some adjustments though in the template. Currently the name serves both as a display but also to define the README path in the `hub_template`

RobbeSneyders:
Fixed :)

PR #651, opened 2023-11-20
Augment DockerRunner to support running from a fondant Pipeline

GeorgesLorre:


GeorgesLorre:
```
  E   TypeError: unsupported operand type(s) for |: 'type' and 'type'
 ```
 Thanks python 3.8...

File: src/fondant/pipeline/runner.py [hunk 1]
```diff
@@ -33,6 +36,39 @@ def run(self, input_spec: str, *args, **kwargs):
...
+                "Found reference to un-compiled pipeline... compiling to {output_path}",
+            )
+            compiler = DockerCompiler()
+            compiler.compile(
+                input,
+                output_path=output_path,
+                extra_volumes=extra_volumes,
+                build_args=build_args,
+            )
+            self.run(output_path)
```
RobbeSneyders, line 68: Shouldn't this be `_run`?

src/fondant/pipeline/runner.py
```diff
@@ -33,6 +36,39 @@ def run(self, input_spec: str, *args, **kwargs):
 
         subprocess.call(cmd)  # nosec
 
+    def run(
+        self,
+        input: t.Union[Pipeline, str],
+        *,
+        output_path: str = "docker-compose.yml",
```
RobbeSneyders, line 43: Do we need to expose this? I would move this to a temp file or hidden directory.

RobbeSneyders (commented):
Thanks @GeorgesLorre! The general approach looks good!

src/fondant/pipeline/runner.py
mrchtr, line 43: I know super picky, but can we use a common yaml file extension within Fondant. Either `yml` or `yaml`?
btw: accordingly [to the docker compose docs](https://docs.docker.com/compose/compose-file/03-compose-file/#:~:text=The%20default%20path%20for%20a,prefers%20the%20canonical%20compose.yaml%20.) the new preferred default file for docker compose is called `compose.yaml`

PhilippeMoussalli: Wouldn't this make it more difficult to run the pipeline again from the same component spec to follow the "compile once, run many" approach (well in this case it would be  "run once, run many")?

Unless this is temporary until we introduce a general compiler that we can start from as proposed here https://github.com/orgs/ml6team/projects/1/views/8?filterQuery=&pane=issue&itemId=44594106

GeorgesLorre: Fair point ( can we add this to precommit ?). I will make the changes

GeorgesLorre:
> > _ No description provided. _
> 
> Are we planning on doing the same for kfp and Vertex?

Yes the idea is to make the Runner classes prime user interfaces (next to the cli). Now you need to import a compiler and runner and and compile first before running. The idea here is that the user would only need 1 class that is easy to use.

src/fondant/pipeline/runner.py
RobbeSneyders: I would only hide this in the Runner. Users can still explicitly compile using the `compile` command or `Compiler` class.

GeorgesLorre: With expose you mean the file and not the argument ?

RobbeSneyders: Both

GeorgesLorre: updated

PR #653, opened 2023-11-20
Update core package

mrchtr:
First PR related to the data structure redesign. 

Implements the following: 
- New manifest structure (including validation, and evolution)
- New ComponentSpec structure (including validation)
- Removes `Subsets` and `Index`

Not all tests are running successfully. But this are already quite a few changes. Therefore, I've created PR on feature branch `feature/redesign-dataset-format-and-interface`, to have quicker feedback loops.


src/fondant/core/component_spec.py
```diff
@@ -191,27 +163,28 @@ def tags(self) -> t.List[str]:
 
     @property
     def index(self):
-        return ComponentSubset({"fields": {}})
+        # TODO: check usage
+        return {"fields": {}}
```
RobbeSneyders, line 167: I think the index should be implemented as a `Field`.

src/fondant/core/component_spec.py
```diff
@@ -191,27 +163,28 @@ def tags(self) -> t.List[str]:
...
+    def consumes(self) -> t.Mapping[str, Field]:
         """The subsets consumed by the component as an immutable mapping."""
         return types.MappingProxyType(
             {
-                name: ComponentSubset(subset)
-                for name, subset in self._specification.get("consumes", {}).items()
-                if name != "additionalSubsets"
+                name: Field(name=name, type=Type.from_json(field))
+                for name, field in self._specification["consumes"].items()
+                if name != "additionalFields"
```
RobbeSneyders, line 176: I don't think we still need this `if name != "additionalFields"`. I would implement a first version without it and then add it in a separate PR.

src/fondant/core/manifest.py
```diff
@@ -15,60 +15,7 @@
...
-class Index(Subset):
-    """Special case of a subset for the index, which has fixed fields."""
-
-    @property
-    def fields(self) -> t.Dict[str, Field]:
-        return {
-            "id": Field(name="id", type=Type("string")),
-            "source": Field(name="source", type=Type("string")),
-        }
+from fondant.core.schema import Field, ManifestField, Type
```
RobbeSneyders, line 18: If we have a separate `ManifestField`, we probably don't want to import `Field` here to make sure it is not used.

src/fondant/core/manifest.py
```diff
@@ -196,13 +143,41 @@ def copy(self) -> "Manifest":
...
+        return self._specification["index"]
+
     def update_metadata(self, key: str, value: t.Any) -> None:
         self.metadata[key] = value
 
     @property
     def base_path(self) -> str:
         return self.metadata["base_path"]
 
+    def get_dataset_location(self, spec: ComponentSpec) -> str:
```
RobbeSneyders, line 157: Can you make this a property?

src/fondant/core/manifest.py
```diff
@@ -196,13 +143,41 @@ def copy(self) -> "Manifest":
...
+        """Determines dataset location using the base_path and component spec."""
+        return self.base_path + "/" + spec.component_folder_name
+
+    def retrieve_field_mapping(self):
```
RobbeSneyders, line 161: Can you make this a property?

src/fondant/core/schema.py
```diff
@@ -161,11 +162,32 @@ def __eq__(self, other):
...
+        self,
+        name: str,
+        type: Type,
+    ) -> None:
+        self._name = name
+        self._type = type
+
+    @property
+    def name(self) -> str:
+        """The absolute location of the field."""
```
RobbeSneyders, line 179: These docstrings should be updated.

src/fondant/core/schema.py
```diff
@@ -161,11 +162,32 @@ def __eq__(self, other):
...
+        return self._name
+
+    @property
+    def type(self) -> Type:
+        """The absolute location of the field."""
+        return self._type
+
+    @property
+    def type_name(self) -> Type:
```
RobbeSneyders, line 188: Is this used, and if so, what is the difference with `type` or `name`?

src/fondant/core/schema.py
```diff
@@ -181,3 +203,16 @@ def validate_partition_size(arg_value):
         )
         raise InvalidTypeSchema(msg)
     return arg_value
+
+
+@dataclass
+class ManifestField(Field):
```
RobbeSneyders, line 209: Do we need to have a separate `ManifestField`, or should we update `Field` to optionally accept a location? I would not give it a `base_path` property. Either let it only hold a relative location, and require the manifest to resolve the absolute path. Or just let the field hold the absolute path.

src/fondant/core/schemas/common.json
```diff
@@ -57,6 +57,11 @@
     "fields": {
       "type": "object",
       "minProperties": 1,
+      "properties": {
```
RobbeSneyders, line 60: Can you explain how you want to implement this now? It's not clear which the final design is to me. As mentioned above, it's fine for me to leave this out for now and add it as a separate PR.

src/fondant/core/schemas/manifest.json
```diff
@@ -60,13 +60,13 @@
       },
```
RobbeSneyders, line 60: The `field` definition here has a `fields` attribute, which I don't think is correct.

src/fondant/core/component_spec.py
RobbeSneyders, line 167: Would return a `Field` instance representing the index here.

RobbeSneyders (commented):
Thanks @mrchtr! I did a first quick review. 

I would focus on getting the `core` module ported first and make sure the tests for those are passing. So everything related to the component spec and manifest.

Could you remove all the other changes from this branch and open those as PRs to this one? That would make it a lot easier to review :)

src/fondant/core/manifest.py
mrchtr: Not really. It is used later on in `data_io.py`

```python
location = self.manifest.get_dataset_location(self.component_spec)
```

We use the component_spec to retrieve the dataset location. `base_path + spec.component_folder_name`
I thought it would be nice to encapsulate this logic into a method, to give this a name. 
But maybe the manifest isn't the right place for it.

src/fondant/core/manifest.py
mrchtr: I've removed the `ManifestField` again.

src/fondant/core/component_spec.py
```diff
@@ -191,38 +163,28 @@ def tags(self) -> t.List[str]:
 
     @property
     def index(self):
-        return ComponentSubset({"fields": {}})
+        return Field(name="index", location=self._specification["index"].location)
 
     @property
-    def consumes(self) -> t.Mapping[str, ComponentSubset]:
+    def consumes(self) -> t.Mapping[str, Field]:
         """The subsets consumed by the component as an immutable mapping."""
```
RobbeSneyders, line 170:
```suggestion
        """The fields consumed by the component as an immutable mapping."""
```

src/fondant/core/component_spec.py
```diff
@@ -191,38 +163,28 @@ def tags(self) -> t.List[str]:
...
-                if name != "additionalSubsets"
+                name: Field(name=name, type=Type.from_json(field))
+                for name, field in self._specification["consumes"].items()
             },
         )
 
     @property
-    def produces(self) -> t.Mapping[str, ComponentSubset]:
+    def produces(self) -> t.Mapping[str, Field]:
         """The subsets produced by the component as an immutable mapping."""
```
RobbeSneyders, line 180:
```suggestion
        """The fields produced by the component as an immutable mapping."""
```

src/fondant/core/manifest.py
```diff
@@ -220,41 +195,76 @@ def cache_key(self) -> str:
         return self.metadata["cache_key"]
 
     @property
-    def index(self) -> Index:
-        return Index(self._specification["index"], base_path=self.base_path)
-
-    @property
-    def subsets(self) -> t.Mapping[str, Subset]:
+    def fields(self) -> t.Mapping[str, Field]:
         """The subsets of the manifest as an immutable mapping."""
```
RobbeSneyders, line 199:
```suggestion
        """The fields of the manifest as an immutable mapping."""
```

src/fondant/core/manifest.py
```diff
@@ -220,41 +195,76 @@ def cache_key(self) -> str:
...
+        # e.g. ('images', {'location': '/component1', 'type': 'binary'})
```
RobbeSneyders, line 200: This comment is confusing to me. It shows a tuple, while we're returning a mapping.

src/fondant/core/manifest.py
```diff
@@ -220,41 +195,76 @@ def cache_key(self) -> str:
...
+                    name=name,
+                    type=Type(field["type"]),
+                    location=field["location"],
+                )
+                for name, field in self._specification["fields"].items()
             },
         )
 
-    def add_subset(
+    def add_fields(
```
RobbeSneyders, line 212: This method is only used in tests as far as I can see, and `add_subset` is no longer used. So I think we can remove it.

src/fondant/core/manifest.py
```diff
@@ -220,41 +195,76 @@ def cache_key(self) -> str:
...
+    def remove_field(self, name: str) -> None:
+        if name not in self._specification["fields"]:
+            msg = f"Field {name} not found in specification"
             raise ValueError(msg)
 
-        del self._specification["subsets"][name]
+        del self._specification["fields"][name]
 
-    def evolve(  # noqa : PLR0912 (too many branches)
+    def evolve(  # : PLR0912 (too many branches)
```
RobbeSneyders, line 267: This comment might have been changed by pre-commit

src/fondant/core/manifest.py
```diff
@@ -274,68 +284,24 @@ def evolve(  # noqa : PLR0912 (too many branches)
         # Update `component_id` of the metadata
         component_id = component_spec.component_folder_name
         evolved_manifest.update_metadata(key="component_id", value=component_id)
+
         if run_id is not None:
             evolved_manifest.update_metadata(key="run_id", value=run_id)
 
         # Update index location as this is currently always rewritten
```
RobbeSneyders, line 291:
```suggestion
        # Update index location as this is always rewritten
```

src/fondant/core/manifest.py
```diff
@@ -274,68 +284,24 @@ def evolve(  # noqa : PLR0912 (too many branches)
...
-        # If additionalFields is False for a consumed subset,
-        # Remove all fields from that subset that are not listed
-        for subset_name, subset in component_spec.consumes.items():
-            if subset_name in evolved_manifest.subsets and not subset.additional_fields:
-                for field_name in evolved_manifest.subsets[subset_name].fields:
-                    if field_name not in subset.fields:
-                        evolved_manifest.subsets[subset_name].remove_field(
-                            field_name,
-                        )
+        evolved_manifest.add_or_update_field(Field(name="index"))
```
RobbeSneyders, line 292: The new location should be added here.

src/fondant/core/manifest.py
```diff
@@ -274,68 +284,24 @@ def evolve(  # noqa : PLR0912 (too many branches)
...
+        # evolved_manifest._specification["index"][
+        #    "location"
+        # ] = f"/{self.pipeline_name}/{evolved_manifest.run_id}/{component_id}"
+
+        # TODO handle additionalFields
 
         # For each output subset defined in the component, add or update it
```
RobbeSneyders, line 299:
```suggestion
        # Add or update all produced fields defined in the component spec
```

src/fondant/core/manifest.py
```diff
@@ -196,13 +143,41 @@ def copy(self) -> "Manifest":
...
+        Example:
+           {
+               "/base_path/component_1": ["Name", "HP"],
+               "/base_path/component_2": ["Type 1", "Type 2"],
+           }
+        """
+        field_mapping = {}
+        for field_name, field in self.fields.items():
+            location = (
+                f"{self.base_path}/{self.pipeline_name}/{self.run_id}{field.location}"
```
RobbeSneyders, line 173:
```suggestion
                f"{self.base_path}/{self.pipeline_name}/{self.run_id}/{field.location}"
```

Is this slash not needed?

src/fondant/core/schema.py
```diff
@@ -161,11 +162,34 @@ def __eq__(self, other):
...
+        return self._name
+
+    @property
+    def type(self) -> Type:
+        """The absolute location of the field."""
+        return self._type
+
+    @property
+    def location(self) -> str:
+        """The absolute location of the field."""
```
RobbeSneyders, line 191:
```suggestion
        """The relative location of the field."""
```
You seem to be using this as a relative location in the other classes.

src/fondant/core/schemas/component_spec.json
```diff
@@ -28,43 +28,18 @@
       }
     },
     "consumes": {
-      "$ref": "#/definitions/subsets"
+      "$ref": "#/definitions/field"
```
RobbeSneyders, line 31:
```suggestion
      "$ref": "common.json#/definitions/fields"
```

src/fondant/core/schemas/component_spec.json
```diff
@@ -28,43 +28,18 @@
...
     },
     "produces": {
-      "$ref": "#/definitions/subsets"
+      "$ref": "#/definitions/field"
```
RobbeSneyders, line 34:
```suggestion
      "$ref": "common.json#/definitions/fields"
```

src/fondant/core/schemas/component_spec.json
```diff
@@ -28,43 +28,18 @@
...
-          "default": true
-        }
-      },
-      "minProperties": 1,
-      "additionalProperties": {
-        "$ref": "#/definitions/subset"
-      }
+    "field": {
+      "$ref": "common.json#/definitions/fields"
     },
```
RobbeSneyders, line 43:
```suggestion
```

RobbeSneyders (commented):
Thanks @mrchtr, clean work!

I still need to review the tests. I would split the examples per module as well though.

src/fondant/core/manifest.py
mrchtr: The schema requires that the location starts with a `/`. What means `field.location` starts always with a `/`.
Look a bit unusual here but it is indeed not needed.
We could think about defining the location without the leading `/`. Not sure if there was a reason to have it.

RobbeSneyders: Ok then it's fine. I think it looks clean starting with a `/` indeed.

mrchtr:
> I still need to review the tests. I would split the test examples per module as well though.

I've cleaned the test a bit more. I moved the relevant sample files into `tests/core/examples`.
Removed some of the test cases for manifest evolution. Mainly the ones related to `additionalFields`. We have to keep this in mind when we are implementing #656. We should add new test (re-add relevant) cases after the implementation.

src/fondant/core/manifest.py
```diff
@@ -274,68 +271,23 @@ def evolve(  # noqa : PLR0912 (too many branches)
...
+        # Update index location as this is always rewritten
+        evolved_manifest.add_or_update_field(
+            Field(name="index", location=component_spec.component_folder_name),
+        )
+
+        # TODO handle additionalFields
+
+        # Add or update all produced fields defined in the component spec
+        for name, field in component_spec.produces.items():
+            # If field was part not part of the input manifest, add field to output manifest.
```
PhilippeMoussalli, line 287:
```suggestion
            # If field was not part of the input manifest, add field to output manifest.
```

src/fondant/core/component_spec.py
```diff
@@ -191,38 +163,28 @@ def tags(self) -> t.List[str]:
 
     @property
     def index(self):
```
RobbeSneyders, line 165: Can you add a return type?

src/fondant/core/component_spec.py
```diff
@@ -191,38 +163,28 @@ def tags(self) -> t.List[str]:
...
-        return ComponentSubset({"fields": {}})
+        return Field(name="index", location=self._specification["index"].location)
```
RobbeSneyders, line 166: I don't think the index has a location in the component spec?

src/fondant/core/manifest.py
```diff
@@ -196,13 +143,41 @@ def copy(self) -> "Manifest":
     def metadata(self) -> t.Dict[str, t.Any]:
         return self._specification["metadata"]
 
+    @property
+    def index(self) -> t.Dict[str, t.Any]:
+        return self._specification["index"]
```
RobbeSneyders, line 148: Can we return a `Field` here to be consistent?

tests/core/examples/evolution_examples/1/output_manifest.json
```diff
@@ -0,0 +1,33 @@
...
+      "images_data": {
+         "type": "binary",
+         "location":"/example_component"
+      },
+      "captions_data": {
+         "type": "binary",
+         "location":"/example_component"
+      },
+      "embeddings_data": {
+         "type": "list<item: float>",
```
PhilippeMoussalli, line 29: I don't think that's the intended way of storing the types in the output manifest. They should match the format defined in the component spec: 
```bash
 "type": array
  "items":
    "type": float32
```

src/fondant/core/manifest.py
```diff
@@ -220,39 +195,61 @@ def cache_key(self) -> str:
...
+        elif overwrite is False and field.name in self._specification["fields"]:
+            msg = (
+                f"A field with name {field.name} already exists. Set overwrite to true, "
+                f"if you want to update the field."
+            )
+            raise ValueError(msg)
+        else:
+            self._specification["fields"][field.name] = {
+                "location": f"/{self.component_id}",
+                "type": field.type.name,
```
PhilippeMoussalli, line 224: See my comment below. We should store them as json mainly to match the format of the component spec 

```suggestion
                "type": field.type.to_json(),
```

src/fondant/core/manifest.py
```diff
@@ -196,13 +143,41 @@ def copy(self) -> "Manifest":
...
+
     def update_metadata(self, key: str, value: t.Any) -> None:
         self.metadata[key] = value
 
     @property
     def base_path(self) -> str:
         return self.metadata["base_path"]
 
+    @property
+    def field_mapping(self):
```
RobbeSneyders, line 158: I was adding some comments to this method, but see you updated it in #654. Do you think we can move the changes here? It should be possible with some git magic.

My comments:
- [x] Add a return type
- [x] Sorted with index first
- [ ] Include index 

Most of them seem to be resolved, except for the last one. I think we need to explicitly include the index, since only the index might be written by a filter component.

src/fondant/core/component_spec.py
mrchtr: I think we don't need the index in the ComponentSpec anymore.

src/fondant/core/manifest.py
mrchtr: Thanks! Wasn't aware of this, but it makes totally sense!

src/fondant/core/component_spec.py
RobbeSneyders: I think you're right.

src/fondant/core/manifest.py
RobbeSneyders: Thanks!

src/fondant/core/manifest.py
```diff
@@ -196,13 +144,52 @@ def copy(self) -> "Manifest":
...
+        location of the dataframe with the index. This allows an efficient left join operation.
+
+        Example:
+           {
+               "/base_path/component_1": ["Name", "HP"],
+               "/base_path/component_2": ["Type 1", "Type 2"],
+           }
+        """
+        field_mapping = {}
+        for field_name, field in {"Index": self.index, **self.fields}.items():
```
RobbeSneyders, line 175: Shouldn't we use "id" here instead of "Index"?

tests/core/test_component_specs.py
```diff
@@ -49,12 +48,19 @@ def test_component_spec_pkgutil_error(mock_get_data):
 
 
 def test_component_spec_validation(valid_fondant_schema, invalid_fondant_schema):
-    """Test that the manifest is validated correctly on instantiation."""
+    """Test that the comp is validated correctly on instantiation."""
```
RobbeSneyders, line 51:
```suggestion
    """Test that the component spec is validated correctly on instantiation."""
```

tests/core/test_manifest_evolution.py
```diff
@@ -35,22 +35,3 @@ def test_evolution(input_manifest, component_spec, output_manifest):
     )
 
     assert evolved_manifest._specification == output_manifest
-
-
-def test_component_spec_location_update():
```
RobbeSneyders, line 40: We could keep this test and just let it check the index path.

tests/examples/example_specs/components/component.yaml
```diff
@@ -3,19 +3,15 @@ description: This is an example component
...
-      data:
-        type: array
-        items:
-          type: float32
-    additionalFields: false
+  images_data:
+    type: array
+    items:
+      type: float32
+additionalFields: false
```
RobbeSneyders, line 14: This was removed for now, right?

tests/core/examples/evolution_examples/1/output_manifest.json
```diff
@@ -0,0 +1,38 @@
...
+         "type": "binary",
+         "location":"/example_component"
+      },
+      "embeddings_data": {
+         "type": {
+            "type": "array",
+            "items": {
+               "type": "float32"
+            }
+         },
```
RobbeSneyders, line 34:
```suggestion
         "type": "array",
         "items": {
            "type": "float32"
         },
```

src/fondant/core/manifest.py
RobbeSneyders: This broke some of the tests though.

```python
>>> type.to_json()

{
  "type": type
}
```

So now we get:
```python
{
  "location": ...,
  "type": {
    "type": type
  }
}
```

I think we can solve it like this:

```python
self._specification["fields"][field.name] = {
  "location": f"/{self.component_id}",
  **field.type.name,
}
```

RobbeSneyders (commented):
Thanks @mrchtr, I think these are the last ones. I think the tests should pass with these changes.

RobbeSneyders:
FYI, as merge strategy, let's squash merge your separate PRs into your feature branch, and then rebase and merge the feature branch when it's ready. Then we keep the PRs as separate commits on main.

tests/examples/example_specs/components/component.yaml
mrchtr: Would leave it here for now. The file is moved to a different folder anyway in #655.

RobbeSneyders (approved):
Great work @mrchtr!

Issue #643, opened 2023-11-15
Implement new pipeline interface


mrchtr:
Changes will effect the `pipeline` package. 

Implementation based on the discussion in #567 

Summarisation of 567, we will implement the following dataset interface: 

```
pipeline = Pipeline(
    pipeline_name="my_pipeline",
    pipeline_description="description of my pipeline",
    base_path="/foo/bar",
)

dataset = pipeline.read(
    name="load_images",
    schema={
        "image": type.binary  # or pa.binary()
    }
)

dataset = dataset.apply(
    name="caption_images", 
    consumes={
        "images_data": "image"
    }
    produces={
        "captions": "text"
    }


dataset = dataset.apply(
    component_dir="embed_text", 
    consumes={
        "text_data":"text"
    }
)

dataset.write(
    name="write_data", 
    schema={
        "image": "image", 
        "caption": "text" 
    }
)
```

```[tasklist]
### Tasks
- [ ] https://github.com/ml6team/fondant/issues/567
- [ ] https://github.com/ml6team/fondant/issues/688
- [ ] https://github.com/ml6team/fondant/issues/689
- [ ] https://github.com/ml6team/fondant/issues/638
- [ ] https://github.com/ml6team/fondant/issues/670
- [ ] https://github.com/ml6team/fondant-usecase-RAG/issues/12
- [ ] https://github.com/ml6team/fondant-usecase-controlnet/issues/4
- [ ] https://github.com/ml6team/fondant-usecase-filter-creative-commons/issues/5
```

PR #654, opened 2023-11-21
Refactor component package

mrchtr:
Refactor component package as part of #643


src/fondant/component/data_io.py
```diff
@@ -82,35 +82,7 @@ def partition_loaded_dataframe(self, dataframe: dd.DataFrame) -> dd.DataFrame:
...
-
-        # add subset prefix to columns
-        subset_df = subset_df.rename(
-            columns={col: subset_name + "_" + col for col in subset_df.columns},
-        )
-
-        return subset_df
-
+    # TODO: probably not needed anymore!
     def _load_index(self) -> dd.DataFrame:
```
RobbeSneyders, line 86: This can be removed.

src/fondant/component/data_io.py
```diff
@@ -135,20 +108,34 @@ def load_dataframe(self) -> dd.DataFrame:
...
-            # left joins -> filter on index
-            dataframe = dd.merge(
-                dataframe,
-                subset_df,
-                left_index=True,
-                right_index=True,
-                how="left",
+        dataframe = None
+        field_mapping = self.manifest.field_mapping
+        for location, fields in field_mapping.items():
```
RobbeSneyders, line 113: We need to make sure that we read the location containing the latest index first, so the left join below is correct.

src/fondant/component/data_io.py
```diff
@@ -135,20 +108,34 @@ def load_dataframe(self) -> dd.DataFrame:
...
+                index="id",
+                calculate_divisions=True,
             )
 
+            if dataframe is None:
+                # ensure that the index is set correctly and divisions are known.
+                dataframe = partial_df
+            else:
+                dask_divisions = partial_df.set_index("id").divisions
+                unique_divisions = list(dict.fromkeys(list(dask_divisions)))
```
RobbeSneyders, line 126: Why is this needed? And why are you using `dict.fromkeys`? I would assume we can use `set`?

src/fondant/component/data_io.py
```diff
@@ -135,20 +108,34 @@ def load_dataframe(self) -> dd.DataFrame:
...
+                columns=fields,
+                index="id",
+                calculate_divisions=True,
             )
 
+            if dataframe is None:
+                # ensure that the index is set correctly and divisions are known.
+                dataframe = partial_df
+            else:
+                dask_divisions = partial_df.set_index("id").divisions
```
RobbeSneyders, line 125: I would expect us to use the divisions from the already loaded dataframe, not from the newly loaded one.

src/fondant/component/data_io.py
```diff
@@ -170,79 +157,46 @@ def write_dataframe(
         dataframe: dd.DataFrame,
         dask_client: t.Optional[Client] = None,
     ) -> None:
-        write_tasks = []
+        columns_to_produce = [
+            column_name for column_name, field in self.component_spec.produces.items()
+        ]
 
-        dataframe.index = dataframe.index.rename("id")
```
RobbeSneyders, line 175: We are assuming that the index is called "id" in the loader, so I think we still need this?

src/fondant/component/executor.py
```diff
@@ -569,12 +549,16 @@ def _execute_component(
 
         # Clear divisions if component spec indicates that the index is changed
         if self._infer_index_change():
+            # TODO: might causing issues for merging components
```
RobbeSneyders, line 552: We still need divisions, but we need to make sure to reset and recalculate them if the index changes. See the [Dask documentation](https://docs.dask.org/en/stable/generated/dask.dataframe.DataFrame.map_partitions.html#:~:text=Also%20note%20that%20the%20index%20and%20divisions%20are%20assumed%20to%20remain%20unchanged.%20If%20the%20function%20you%E2%80%99re%20mapping%20changes%20the%20index/divisions%2C%20you%E2%80%99ll%20need%20to%20clear%20them%20afterwards%3A).

RobbeSneyders (commented):
Just a quick partial review. I'll need to take my time to dive a bit deeper to review with Dask performance in mind.

src/fondant/component/data_io.py
mrchtr: I've updated the `field_mapping` method. The method returns an immutable OrderedDict now. The first entry is the location of the dataset of the latest index.

src/fondant/component/data_io.py
mrchtr: To summarise theory to double check that I didn't missed a detail:
- we are saving parquet files with index and devision information
- start reading the parquet file with the most recent index (index.name = id)
- loading iterative the next parquet file (containing the same divisions and index.name = id) and apply a left join

If so, we can trust the division and index information of the parquet files and don't need to set it here.

src/fondant/component/data_io.py
PhilippeMoussalli: I might have missed something but it seems to be that we're currently loading/merging all the fields defined in the manifest since that's what `field_mapping` contains. Shouldn't we only base ourselves on the fields defined in the `consumes` section of the component spec?

RobbeSneyders: Good catch :sweat_smile:

I think this should work:
```python
fields = [field for field in fields if 
          (field in self.component_spec.consumes or field == DEFAULT_INDEX_NAME)]
```

src/fondant/component/data_io.py
RobbeSneyders: Yes, that's my understanding as well.

src/fondant/component/data_io.py
mrchtr: I've started the implementation of the consumes and produces in #665.
Maybe it is easier to include the changes of the component package there here as well! I'll take a look. 

Basically, the idea was the following:

- Load the dataframe based on the manifest, later on renaming the columns based on the defined consumes before the dataset is passed to the transform method. 
- When produce is defined renaming the columns of the output dataframe accordingly to the produce mapping, write it to the bucket and use the written column names for the output manifest. 

Am I overlooking something?

PhilippeMoussalli: > Good catch 😅
> 
> I think this should work:
> 
> ```python
> fields = [field for field in fields if 
>           (field in self.component_spec.consumes or field == DEFAULT_INDEX_NAME)]
> ```

I don't see why we need the manifest field mapping in the first place. 

Can't we do something like this (a bit similar to how it was before):

```
for field_name in self.component_spec.consumes.items():
             field_location = self.manifest.fields[field_name].location
```

RobbeSneyders:
I merged the base branch into this one so we can review @mrchtr. Please double check that I did it correctly :)

src/fondant/component/component.py
```diff
@@ -16,7 +16,9 @@ class BaseComponent:
         **kwargs: The provided user arguments are passed in as keyword arguments
     """
 
-    def __init__(self, spec: ComponentSpec, **kwargs):
+    def __init__(
+        self, spec: ComponentSpec, schema: t.Optional[t.Dict[str, str]] = None, **kwargs
```
RobbeSneyders, line 20: Where is this used? If we do need it, can you add it to the docstring?

src/fondant/component/data_io.py
```diff
@@ -25,9 +27,11 @@ def __init__(
         manifest: Manifest,
         component_spec: ComponentSpec,
         input_partition_rows: t.Optional[int] = None,
+        consumes: t.Optional[t.Dict[str, str]] = None,
```
RobbeSneyders, line 30: I assume these are changes related to the pipeline interface change? Can we keep those out of this PR for now?

src/fondant/component/data_io.py
```diff
@@ -82,77 +86,58 @@ def partition_loaded_dataframe(self, dataframe: dd.DataFrame) -> dd.DataFrame:
...
         dataframe = self.partition_loaded_dataframe(dataframe)
 
         logging.info(f"Columns of dataframe: {list(dataframe.columns)}")
 
+        # rename columns accordingly to the consumes
+        if self.consumes:
+            reverted_consumes = {v: k for k, v in self.consumes.items()}
+            dataframe = dataframe.rename(columns=reverted_consumes)
+
+        logging.info(f"Columns of components dataframe: {list(dataframe.columns)}")
```
RobbeSneyders, line 140: Same here, can we keep these changes out for now?

src/fondant/component/data_io.py
```diff
@@ -162,87 +147,64 @@ def __init__(
         *,
         manifest: Manifest,
         component_spec: ComponentSpec,
+        produces: t.Optional[t.Dict[str, str]] = None,
```
RobbeSneyders, line 150: Same here, can we keep these changes out for now?

src/fondant/component/data_io.py
```diff
@@ -162,87 +147,64 @@ def __init__(
...
-                subset_name=subset_name,
-                subset_spec=subset_spec,
-            )
-            write_tasks.append(write_subset_task)
+        dataframe = dataframe[columns_to_produce]
+
+        # Rename produces accordingly
+        if self.produces:
+            reverted_produces = {v: k for k, v in self.produces.items()}
+            dataframe = dataframe.rename(columns=reverted_produces)
```
RobbeSneyders, line 174: Same here, can we keep these changes out for now?

src/fondant/component/executor.py
RobbeSneyders, line 1: I think most changes in this file are related to the changed interface. Can we keep this out of this PR for now?

RobbeSneyders (commented):
Can we keep any changes related to the changed interface out of this PR and in #665? It's a bit hard to keep track otherwise.

RobbeSneyders (approved):
Thanks @mrchtr! Good to go!

Issue #661, opened 2023-11-22
Create Sagemaker compiler that loops over fondant pipeline and creates a Sagemaker pipeline spec


GeorgesLorre:

GeorgesLorre:
[sagemaker pipeline spec](https://aws-sagemaker-mlops.github.io/sagemaker-model-building-pipeline-definition-JSON-schema/)

PR #662, opened 2023-11-22
Feature/sagemaker compiler

GeorgesLorre:
resolves: https://github.com/ml6team/fondant/issues/661


src/fondant/pipeline/compiler.py
```diff
@@ -515,3 +517,167 @@ def _set_configuration(self, task, fondant_component_operation):
...
+            self.sagemaker = sagemaker
+
+        except ImportError:
+            msg = """You need to install the sagemaker extras to use the sagemaker compiler,\n
+                     you can install it with `pip install fondant[sagemaker]`"""
+            raise ImportError(
+                msg,
+            )
+
+    def build_command(
```
GeorgesLorre, line 542: So this could be reused by other compilers and be abstracted more

src/fondant/pipeline/compiler.py
```diff
@@ -515,3 +517,167 @@ def _set_configuration(self, task, fondant_component_operation):
...
+        """
+        run_id = pipeline.get_run_id()
+        path = pipeline.base_path
+        pipeline.validate(run_id=run_id)
+
+        component_cache_key = None
+
+        steps: t.List[t.Any] = []
+
+        with tempfile.TemporaryDirectory(dir=os.getcwd()) as tmpdirname:
```
GeorgesLorre, line 607: the scripts are uploaded to s3 by sagemaker once we compile the pipeline that is why we save the files to a temp folder that only gets deleted at the end.

GeorgesLorre:
So the coverage dropping is quite painful but the coverage overall is pretty bad for `compiler.py`. Once we have all the pieces in place (updated runners, abstracted command building) we can take another good look on how to test the compilers.

src/fondant/pipeline/compiler.py
```diff
@@ -515,3 +517,167 @@ def _set_configuration(self, task, fondant_component_operation):
...
+    ) -> None:
+        """Compile a fondant pipeline to sagemaker pipeline spec and save it
+        to a specified output path.
+
+        Args:
+            pipeline: the pipeline to compile
+            output_path: the path where to save the Kubeflow pipeline spec.
+            instance_type: the instance type to use for the processing steps
+            (see: https://aws.amazon.com/ec2/instance-types/ for options).
+            role_arn: the role arn to use for the processing steps,
```
PhilippeMoussalli, line 596:
```suggestion
            role_arn: the Amazon Resource Name role to use for the processing steps,
```

PhilippeMoussalli: wasn't very clear what "arn" is

src/fondant/pipeline/compiler.py
```diff
@@ -515,3 +517,167 @@ def _set_configuration(self, task, fondant_component_operation):
...
+                )
+
+        return command
+
+    def compile(
+        self,
+        pipeline: Pipeline,
+        output_path: str,
+        instance_type: str = "ml.t3.medium",
+        role_arn: t.Optional[str] = None,
```
PhilippeMoussalli, line 586:
```suggestion
    def compile(
        self,
        pipeline: Pipeline,
        *,
        output_path: str,
        instance_type: str = "ml.t3.medium",
        role_arn: t.Optional[str] = None,
```

src/fondant/pipeline/compiler.py
```diff
@@ -515,3 +517,167 @@ def _set_configuration(self, task, fondant_component_operation):
...
+                depends_on = [steps[-1]] if component["dependencies"] else []
+
+                script = self.generate_component_script(
+                    component_name,
+                    command,
+                    tmpdirname,
+                )
+
+                if not role_arn:
+                    role_arn = self.sagemaker.get_execution_role()
```
PhilippeMoussalli, line 638: Can you add a comment on what this does exactly?

tests/test_compiler.py
```diff
@@ -605,3 +607,52 @@ def test_caching_dependency_kfp(tmp_path_factory):
...
+        "--output_manifest_path",
+        "/foo/bar/example_pipeline/example_pipeline_2024/component_2/manifest.json",
+        "--foo",
+        "'bar'",
+        "--baz",
+        "'qux'",
+    ]
+
+    # with dependencies
+    dependencies = ["component_2"]
```
PhilippeMoussalli, line 638: small nitpick but would name it `component_1`. Right now it looks like the component is dependent on itself

src/fondant/pipeline/compiler.py
```diff
@@ -515,3 +517,167 @@ def _set_configuration(self, task, fondant_component_operation):
...
+
+    def _set_configuration(self, *args, **kwargs) -> None:
+        raise NotImplementedError
+
+    def generate_component_script(
+        self,
+        component_name: str,
+        command: t.List[str],
+        directory: str,
+    ):
```
PhilippeMoussalli, line 675: Could you add a return type?

src/fondant/pipeline/compiler.py
```diff
@@ -515,3 +517,167 @@ def _set_configuration(self, task, fondant_component_operation):
...
+                logger.info(f"Compiling service for {component_name}")
+
+                command = self.build_command(
+                    metadata,
+                    component_op.arguments,
+                    component["dependencies"],
+                )
+                depends_on = [steps[-1]] if component["dependencies"] else []
+
+                script = self.generate_component_script(
```
PhilippeMoussalli, line 631: Could you rename `script` to `script_path`. At first glance it seems like the function is returning the script string

src/fondant/pipeline/compiler.py
PhilippeMoussalli: would make it an internal function and rename it to `get_build_command()`

src/fondant/pipeline/runner.py
```diff
@@ -65,7 +65,7 @@ def run(
             experiment = self.client.get_experiment(experiment_name=experiment_name)
         except ValueError:
             logger.info(
-                f"Defined experiment '{experiment_name}' not found. Creating new experiment"
+                f"defined experiment '{experiment_name}' not found. creating new experiment"
```
PhilippeMoussalli, line 68: I think the previous formatting was ok

tests/test_compiler.py
```diff
@@ -605,3 +607,52 @@ def test_caching_dependency_kfp(tmp_path_factory):
...
+        "--input_manifest_path",
+        "/foo/bar/example_pipeline/example_pipeline_2024/component_2/manifest.json",
+    ]
+
+
+def test_sagemaker_generate_script(tmp_path_factory):
+    compiler = SagemakerCompiler()
+    command = ["echo", "hello world"]
+    with tmp_path_factory.mktemp("temp") as fn:
+        script = compiler.generate_component_script("component_1", command, fn)
```
PhilippeMoussalli, line 653: Same here with the path

```suggestion
        script_path = compiler.generate_component_script("component_1", command, fn)
```

PhilippeMoussalli (commented):
Thanks Georges! Left a few comments.

src/fondant/pipeline/compiler.py
```diff
@@ -515,3 +517,167 @@ def _set_configuration(self, task, fondant_component_operation):
...
+                    ],
+                )
+
+        return command
+
+    def compile(
+        self,
+        pipeline: Pipeline,
+        output_path: str,
+        instance_type: str = "ml.t3.medium",
```
mrchtr, line 585: Could we use a different machine type? I think you had to wait a while until the provisioning was done?

src/fondant/pipeline/compiler.py
```diff
@@ -515,3 +517,167 @@ def _set_configuration(self, task, fondant_component_operation):
...
+                script = self.generate_component_script(
+                    component_name,
+                    command,
+                    tmpdirname,
+                )
+
+                if not role_arn:
+                    role_arn = self.sagemaker.get_execution_role()
+
+                processor = self.sagemaker.processing.ScriptProcessor(
```
mrchtr, line 640: Not sure if it happens already in a different place. Could we add a clear exception message if a user isn't logged in to an AWS account? Maybe also check if the user has permissions to create sagemaker jobs.

src/fondant/pipeline/compiler.py
GeorgesLorre: I still need to do some experimentation here, but I will update it once I have a better view.

src/fondant/pipeline/compiler.py
GeorgesLorre: The Sagemaker SDK will produce errors.

```
botocore.exceptions.NoCredentialsError: Unable to locate credentials
```

PhilippeMoussalli (approved):
Thanks Georges!

RobbeSneyders (approved):
Thanks @GeorgesLorre!

mrchtr (approved):

PR #665, opened 2023-11-22
Implementation of new pipeline interface

mrchtr:
Draft implementation of the new pipeline interface:
- modified `pipeline.py` (introduce new interface itself)
- renaming dataset accordingly to the defined `consumes` and `produces`
- introduce schema which can be used inside of generic components

Note: Didn't fixed the test yet. Added a dummy test just to visualise how the pipeline could look like. First wanted to check if this goes into the right direction.


components/embed_text/fondant_component.yaml
```diff
@@ -40,12 +36,12 @@ args:
       Pass only the keys required by the model provider or conveniently pass all keys you will ever need. 
       Pay attention how to name the dictionary keys so that they can be used by the model provider.
     type: dict
-    default: {}
+    default: { }
```
PhilippeMoussalli, line 39: any reason for this?

src/fondant/pipeline/pipeline.py
```diff
@@ -319,11 +328,136 @@ def __init__(
...
+                                                arguments=arguments,
+                                                input_partition_rows=input_partition_rows,
+                                                cache=cache,
+                                                cluster_type=cluster_type,
+                                                client_kwargs=client_kwargs,
+                                                resources=resources,
+                                                schema=schema)
+
+        self.add_op(component_op)
+        return self
```
PhilippeMoussalli, line 352: General question 

Do we need to return the class instance. So instead of this:

```python
pipeline = Pipeline(
    pipeline_name="my_pipeline",
    pipeline_description="description of my pipeline",
    base_path="/foo/bar",
)

dataset = pipeline.read(
    name="load_images",
    schema={
        "image": type.binary  # or pa.binary()
    }
)
```

We would have: 

```python
pipeline = Pipeline(
    pipeline_name="my_pipeline",
    pipeline_description="description of my pipeline",
    base_path="/foo/bar",
)

pipeline.read(
    name="load_images",
    schema={
        "image": type.binary  # or pa.binary()
    }
)
```

A bit similar to the the way we used to add operations to the pipeline. I think both are valid but not sure which one is more intuitive

src/fondant/pipeline/pipeline.py
```diff
@@ -319,11 +328,136 @@ def __init__(
...
+        # Get previous component
+        previous_component = list(self._graph.items())[-1]
+        if previous_component is None:
+            msg = f"No previous component found."
+            raise ValueError(msg)
+
+        self.add_op(component_op, dependencies=previous_component)
+        return self
+
+    def _build_component_op(self, name, *, arguments: t.Optional[t.Dict[str, t.Any]] = None,
```
PhilippeMoussalli, line 417: Can you move this to the top?

src/fondant/pipeline/pipeline.py
```diff
@@ -361,6 +496,8 @@ def add_op(
             "dependencies": dependencies_names,
         }
 
+        return self
```
PhilippeMoussalli, line 499: Don't think this is needed since you're not returning the attribute in your read/apply/write functions

src/fondant/pipeline/pipeline.py
```diff
@@ -319,11 +328,136 @@ def __init__(
...
+    def write(self,
+              name,
+              *,
+              arguments: t.Optional[t.Dict[str, t.Any]] = None,
+              input_partition_rows: t.Optional[t.Union[str, int]] = None,
+              cache: t.Optional[bool] = True,
+              cluster_type: t.Optional[str] = "default",
+              client_kwargs: t.Optional[dict] = None,
+              resources: t.Optional[Resources] = None,
+              consumes: t.Optional[t.Dict[str, str]] = None,
```
PhilippeMoussalli, line 395: Does the write function need a consumes?

src/fondant/pipeline/pipeline.py
```diff
@@ -319,11 +328,136 @@ def __init__(
...
+    def apply(self,
+              name,
+              *,
+              arguments: t.Optional[t.Dict[str, t.Any]] = None,
+              input_partition_rows: t.Optional[t.Union[str, int]] = None,
+              cache: t.Optional[bool] = True,
+              cluster_type: t.Optional[str] = "default",
+              client_kwargs: t.Optional[dict] = None,
+              resources: t.Optional[Resources] = None,
+              consumes: t.Optional[t.Dict[str, str]] = None,
```
PhilippeMoussalli, line 364: I would add return an error if both consumes and produces are None

src/fondant/pipeline/pipeline.py
```diff
@@ -319,11 +328,136 @@ def __init__(
...
+    def read(self,
+             name,
+             *,
+             arguments: t.Optional[t.Dict[str, t.Any]] = None,
+             input_partition_rows: t.Optional[t.Union[str, int]] = None,
+             cache: t.Optional[bool] = True,
+             cluster_type: t.Optional[str] = "default",
+             client_kwargs: t.Optional[dict] = None,
+             resources: t.Optional[Resources] = None,
+             schema: t.Optional[t.Dict[str, str]] = None) -> "Pipeline":
```
PhilippeMoussalli, line 340: The schema here is mandatory I suppose

src/fondant/pipeline/pipeline.py
```diff
@@ -319,11 +328,136 @@ def __init__(
...
+              name,
+              *,
+              arguments: t.Optional[t.Dict[str, t.Any]] = None,
+              input_partition_rows: t.Optional[t.Union[str, int]] = None,
+              cache: t.Optional[bool] = True,
+              cluster_type: t.Optional[str] = "default",
+              client_kwargs: t.Optional[dict] = None,
+              resources: t.Optional[Resources] = None,
+              consumes: t.Optional[t.Dict[str, str]] = None,
+              schema: t.Optional[t.Dict[str, str]] = None) -> "Pipeline":
```
PhilippeMoussalli, line 396: Same comment as above, is the schema here just needed for renaming or does it also specify which fields to write?

src/fondant/pipeline/pipeline.py
PhilippeMoussalli, line 1: Would docstring all the read/write/apply functions since they're user facing

tests/pipeline/examples/pipelines/valid_pipeline/example_1/fourth_component/fondant_component.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+args:
+  storage_args:
+    description: Storage arguments
+    type: str
+  some_list:
+    description: Some list
+    type: list
+    items:
+      type: int
```
PhilippeMoussalli, line 28: I don't think we can define arguments like this. This is only the case for the fields

PhilippeMoussalli (commented):
Thanks Matthias! I think this is headed in the right direction :) 
Left some comments

src/fondant/pipeline/pipeline.py
mrchtr: The idea behind returning the class instance was that to call apply on a `dataset` afterwards.
Returning the class instance gives the user the most flexibility. The user can decide which variant he wants to implement. 

Either:
```python
dataset = pipeline.read(...)
dataset = dataset.apply(...)
```

or 

```python
pipeline.read(...)
dataset.apply(...)
```

or even: 

```python
pipeline.read(...).apply(...).apply(...)
```

src/fondant/pipeline/pipeline.py
mrchtr: The schema should specify which fields and belonging types will be written.

src/fondant/pipeline/pipeline.py
PhilippeMoussalli: ok that makes sense :)

src/fondant/component/executor.py
```diff
@@ -75,6 +78,15 @@ def __init__(
         self.metadata = Metadata.from_dict(metadata)
         self.user_arguments = user_arguments
         self.input_partition_rows = input_partition_rows
+        self.schema = schema
+
+        # TODO: either overwrite the component spec
```
mrchtr, line 83: Looking for some input on this todo. Either we are overwrite the component spec, or we passing the consumes and produces further to data_io. Where could use the attribute values of consumes and produces if available instead of the component spec. 
Not sure which one is the better approach.

src/fondant/pipeline/pipeline.py
```diff
@@ -319,11 +333,193 @@ def __init__(
...
+        self,
+        name,
+        *,
+        arguments: t.Optional[t.Dict[str, t.Any]] = None,
+        input_partition_rows: t.Optional[t.Union[str, int]] = None,
+        cache: t.Optional[bool] = True,
+        cluster_type: t.Optional[str] = "default",
+        client_kwargs: t.Optional[dict] = None,
+        resources: t.Optional[Resources] = None,
+        schema: t.Dict[str, str],
```
PhilippeMoussalli, line 376: Shouldn't this schema be `Dict[str, Type]`? I would add some example on how this would look like in the docstring so the users know what to input and mainly because it's different from the `write` schema (this one would also benefit from having examples in the description)

src/fondant/pipeline/pipeline.py
```diff
@@ -319,11 +333,193 @@ def __init__(
...
+        self,
+        name,
+        *,
+        arguments: t.Optional[t.Dict[str, t.Any]] = None,
+        input_partition_rows: t.Optional[t.Union[str, int]] = None,
+        cache: t.Optional[bool] = True,
+        cluster_type: t.Optional[str] = "default",
+        client_kwargs: t.Optional[dict] = None,
+        resources: t.Optional[Resources] = None,
+        consumes: t.Optional[t.Dict[str, str]] = None,
```
PhilippeMoussalli, line 460: does the `write` requires a `consumes`?

src/fondant/pipeline/pipeline.py
RobbeSneyders: Also, the `Dataset` class can provide an interface for more interactive data exploration.

RobbeSneyders:
Closing in favor of #685

PR #666, opened 2023-11-22
Add document viewer to dataset explorer

PhilippeMoussalli:
PR that adds document viewer to data explorer. The viewer attempts to automatically detect the string type and render it to its most appropriate format. Note that the normal string value does not get rendered to a pdf value but rather a normal text/markdown.

I don't think it makes a lot of sense to render normal unformated text to pdf. We can still detect whether a strings is a base64 pdf encoded string and render that to it's original pdf format. 

Screenshot from 2023-11-22 17-32-32


RobbeSneyders (commented):
Thanks @PhilippeMoussalli! Will this be the only way to view text that is too long to display, or will you still add a tooltip as well?

PhilippeMoussalli:
> Thanks @PhilippeMoussalli! Will this be the only way to view text that is too long to display, or will you still add a tooltip as well?

I was trying to include it but did not manage to get it working (requires some custom java script). Someone posted a solution recently but It did not work on my end. I asked them for more info 

https://discuss.streamlit.io/t/tooltip-with-aggrid/55396

RobbeSneyders:
> I was trying to include it but did not manage to get it working (requires some custom java script). Someone posted a solution recently but It did not work on my end. I asked them for more info
> 
> https://discuss.streamlit.io/t/tooltip-with-aggrid/55396

Just did a quick test, and this seems to work:
```python
for field in fields:
    options_builder.configure_column(field=field, tooltipField=field)
```

It's quite slow to show up the first time though.

RobbeSneyders:
Seems like the delay might be configurable:
https://stackoverflow.com/questions/55805496/ag-grid-tooltip-takes-long-time-to-render

data_explorer/app/pages/dataset.py
```diff
@@ -1,15 +1,41 @@
...
 from df_helpers.image_render import configure_image_builder, convert_image_column
 from interfaces.dataset_interface import DatasetLoaderApp
-from st_aggrid import AgGrid, ColumnsAutoSizeMode, GridOptionsBuilder
+from st_aggrid import AgGrid, ColumnsAutoSizeMode, GridOptionsBuilder, JsCode
 
 
 class DatasetExplorerApp(DatasetLoaderApp):
     @staticmethod
-    def setup_app_page(dataframe, fields):
+    def _get_hover_vis_script(field: str) -> JsCode:
```
RobbeSneyders, line 13: Is there a reason to add this custom javascript? I think the original style fits better with the style of the data explorer.

RobbeSneyders (commented):
I think you accidentally removed the document viewer :sweat_smile:

data_explorer/app/pages/dataset.py
PhilippeMoussalli: Removed it, I thought having some contrast might be a bit nicer but we can keep it simple

PhilippeMoussalli:
> I think you accidentally removed the document viewer 😅

woops, readded it!

RobbeSneyders (approved):
Thanks @PhilippeMoussalli!

PR #667, opened 2023-11-23
Use new data format

mrchtr:
This PR applies the usage of the new data format:

- fixes all tests
- update component specifications and component code
- remove subset field usage in `pipeline.py`


components/chunk_text/tests/chunk_text_test.py
```diff
@@ -3,11 +3,11 @@
 from src.main import ChunkTextComponent
 
 
-def test_transform():
+def test_transform__():
```
RobbeSneyders, line 6: Why the `__` suffix?

components/embed_text/fondant_component.yaml
```diff
@@ -40,12 +36,12 @@ args:
       Pass only the keys required by the model provider or conveniently pass all keys you will ever need. 
       Pay attention how to name the dictionary keys so that they can be used by the model provider.
     type: dict
-    default: {}
+    default: { }
```
RobbeSneyders, line 39: Can you turn this off in your formatter?

components/embedding_based_laion_retrieval/fondant_component.yaml
```diff
@@ -7,18 +7,14 @@ tags:
...
+      type: float32
 
 produces:
-  images:
-    fields:
-      url:
-        type: string
+  images_url:
+    type: string
   additionalSubsets: false
```
RobbeSneyders: You might want to comment this one out as well.

components/image_cropping/src/main.py
```diff
@@ -46,12 +46,12 @@ def __init__(
...
     def transform(self, dataframe: pd.DataFrame) -> pd.DataFrame:
         # crop images
-        dataframe["images"]["data"] = dataframe["images"]["data"].apply(
+        dataframe["images_data"] = dataframe["images_data"].apply(
             lambda image: remove_borders(image, self.cropping_threshold, self.padding),
         )
 
         # extract width and height
-        dataframe["images"][["width", "height"]] = dataframe["images"]["data"].apply(
+        dataframe["images_width", "images_width"] = dataframe["images_data"].apply(
```
RobbeSneyders, line 54:
```suggestion
        dataframe["images_width", "images_height"] = dataframe["images_data"].apply(
```

src/fondant/core/manifest.py
```diff
@@ -238,7 +238,7 @@ def remove_field(self, name: str) -> None:
 
         del self._specification["fields"][name]
 
-    def evolve(  # noqa : PLR0912 (too many branches)
+    def evolve(  # : PLR0912 (too many branches)
```
RobbeSneyders, line 241: Something happened to the comment here.

src/fondant/pipeline/pipeline.py
```diff
@@ -443,50 +443,36 @@ def _validate_pipeline_definition(self, run_id: str):
...
-                                f"'{component_spec.name}' component does not match  the "
-                                f"previously created subset definition.\n The '{field_name}' "
-                                f"field is currently defined with the following schema:\n"
-                                f"{manifest_fields[field_name]}\nThe current component to "
-                                f"trying to invoke it with this schema:\n{subset_field}"
-                            )
-                            raise InvalidPipelineDefinition(
-                                msg,
-                            )
+                        # Get the corresponding manifest fields
```
RobbeSneyders, line 459:
```suggestion
                    # Get the corresponding manifest fields
```

RobbeSneyders (commented):
Thanks @mrchtr!

All the test files in `tests/pipeline/examples are shown as newly added. I think this is because you didn't remove the old ones (or at least didn't include this in git).

I would be in favor of choosing better column names. Eg.
- `images_data` -> `image`
- `captions_data` -> `caption`

But for me it's fine to include it like this for now. We can change them later.

components/chunk_text/tests/chunk_text_test.py
mrchtr: PyCharm related bug. For some reasons I can't execute different tests with the same name in different folders.

Issue #656, opened 2023-11-22
Implement previous_index


mrchtr:
Based on #567 `additionalFields` can be defined inside the `consumes` and `produces` section. 
Additional we need a separat keyword to mark index changes (e.g. `previous_index`)

#653 removed some manifest evolution test cases. We should add new test cases there to test the new logic of additionalFields.

PR #668, opened 2023-11-24
Implement `previous_index` field

RobbeSneyders:
#656 

We might want to validate this by checking that the field mentioned in `previous_index` is also defined in the `consumes` section.


mrchtr (commented):
`tests/example_specs/evolution_examples` (on main) contains 8 examples, also some related to the additionalFields. I've deleted some of them in my previous PR cause we have disabled the additionalFields.
Could you have a look if all test cases (from the main branch) are still covered?

mrchtr (approved):
Thanks @RobbeSneyders. Beside my comment above it looks good so far.

RobbeSneyders:
I checked the tests. We had multiple to test the different combinations of `additionalFields` and `additionalSubsets`. We no longer need those, now we just need one to test the `previous_index`.

PR #669, opened 2023-11-24
Fix cli creds

PhilippeMoussalli:
PR that unifies the way extra volumes are passed across the compiler/runner/explorer. `--credentials` argument was removed in favor of `--extra-volumes` to make it more uniform.


GeorgesLorre (approved):
Cool! Thx.

PR #671, opened 2023-11-24
Explorer front page

PhilippeMoussalli:
PR that adds the explorer front page: 

Screenshot from 2023-11-24 16-30-23


RobbeSneyders:
Thanks @PhilippeMoussalli!

Nice to see that it links to the data explorer. By combining this with [query parameters](https://discuss.streamlit.io/t/how-to-use-dropdown-with-url-query-parameter/13337), we could probably make this even pre-select the relevant data. I don't think this is a priority though.

RobbeSneyders (approved):

PR #675, opened 2023-11-27
Fix output dataframe path

RobbeSneyders:
The output dataframe was not written to the correct path:

image

It might make sense to centralize this functionality somewhere, but not sure where.


PhilippeMoussalli (approved):
Thanks!

PR #677, opened 2023-11-27
Hide partitions from users

PhilippeMoussalli:
PR that hides the partitions from the users. Users can now scroll through the dataset through a next/previous buttons. 

In addition, a small changes was made to the statistics pages. Previously it had a mix of both local and global statistics which made it confusing. Now it's only restricted to global statistics. 

https://github.com/ml6team/fondant/assets/47530815/05a32c54-7fbe-45c3-b7ba-d13e976674e2


RobbeSneyders:
Thanks @PhilippeMoussalli!

I think it's a bit confusing what the difference is between the `previous` / `next` buttons you added and the pagination of the dataframe view.

PhilippeMoussalli:
> Thanks @PhilippeMoussalli!
> 
> I think it's a bit confusing what the difference is between the `previous` / `next` buttons you added and the pagination of the dataframe view.

I agree, I removed the pagination completely so it should be good if we keep the default rows to display at 20
image

PR #678, opened 2023-11-27
Augment SagemakerRunner to support running from pipeline objects

GeorgesLorre:
Same as for the docker runner except its uglier

### DIFF
--- a/tests/pipeline/test_runner.py
+++ b/tests/pipeline/test_runner.py
@@ -183,27 +183,29 @@ def test_sagemaker_runner(tmp_path_factory):
         ]
 
 
-class MockSagemakerCompiler:
-    def compile(
-        self,
-        pipeline,
-        output_path,
-        *,
-        instance_type,
-        role_arn,
-    ) -> None:
-        with open(output_path, "w") as f:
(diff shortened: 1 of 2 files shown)


RobbeSneyders (approved):
Thanks @GeorgesLorre. One minor comment on the test.

tests/pipeline/test_runner.py
```diff
@@ -181,3 +181,29 @@ def test_sagemaker_runner(tmp_path_factory):
...
+    with mock.patch(
+        "fondant.pipeline.runner.SagemakerCompiler",
+        new=MockSagemakerCompiler,
+    ), mock.patch("boto3.client", spec=True):
+        runner = SagemakerRunner()
+        runner.run(
+            input=PIPELINE,
+            pipeline_name=PIPELINE.name,
+            role_arn="arn:something",
+        )
```
RobbeSneyders, line 209: Probably makes sense here to test that the compiler is actually called?

GeorgesLorre: After a nights sleep I figured out a way

PR #679, opened 2023-11-27
Use cleaner field names in reusable components

RobbeSneyders:
This PR cleans up the field names in the reusable components. They were just concatenated when first migrating away from the subsets.

I tested all components with tests, fixed the outdated tests, and standardized the test directory structure. Each `tests` directory now has a `pytest.ini` so the `PYTHONPATH` is set  correctly both inside and outside of docker, the `test_requirements.txt` was moved into the `tests` directory, and the `Dockerfile` was updated accordingly.


components/language_filter/Dockerfile
PhilippeMoussalli, line 1: language_filter -> filter_text_language

components/image_cropping/README.md
PhilippeMoussalli, line 1: crop_images

components/image_resolution_extraction/README.md
PhilippeMoussalli, line 1: extract_image_resolution

PhilippeMoussalli (commented):
Thanks Robbe! 

Left some suggestion on changing the name for a few components. I think we should generally stick to the convention <action>_<modality> for consistency

PR #680, opened 2023-11-27
Add cli commands for sagemaker

GeorgesLorre:


src/fondant/cli.py
```diff
@@ -390,9 +394,37 @@ def register_compile(parent_parser):
         default="vertex_pipeline.yml",
     )
 
+    # sagemaker parser
+    sagemaker_parser.add_argument(
+        "ref",
+        help="""Reference to the pipeline to run, can be a path to a spec file or
```
RobbeSneyders, line 400: I see that this is the same for the other frameworks, but this is probably the help text that should be on the runner. For the compiler, this always needs to be a module containing the pipeline instance I guess?

RobbeSneyders (approved):
Thanks @GeorgesLorre! One minor comment :)

src/fondant/cli.py
```diff
@@ -390,9 +391,36 @@ def register_compile(parent_parser):
...
+        "ref",
+        help="""Reference to the pipeline to run, a path to a to a module containing
+        the pipeline instance that will be compiled (e.g. my-project/pipeline.py)""",
+        action="store",
+    )
+    sagemaker_parser.add_argument(
+        "--output-path",
+        "-o",
+        help="Output path of compiled pipeline",
+        default="sagemaker_pipeline.json",
```
PhilippeMoussalli, line 405: aren't we know hiding the compiled pipeline files in a `.pipeline` folder?

PR #682, opened 2023-11-27
Explorer new dataset format

PhilippeMoussalli:
PR that changes the explorer to match the new dataset format. 

image

There is some weird refersh  error that is occurring which might get resolved/ will be easier to debug after removing the partitions. 
Let's focus on first getting this [PR](https://github.com/ml6team/fondant/pull/677) merged since it might fix the issue


PhilippeMoussalli:
> Thanks @PhilippeMoussalli!
> 
> Ran into some issues when testing:

Yes I ran into a few of those as well, I introduced new updates to fix them: 

**- Flickering**
Was due to the fact that the pandas dataframe kept reloading everytime and merging, fixed this by caching the function that loads it. We're using the mapping fields dict as the cache key in this case (not possible to cache dask dataframe object)

The rest of the small issues were also handled (pipeline visualization, image gallery, numerical analysis, fixed screen size). I fixed other issues as well to better navigate the datasets. Let me know if you run into any more issues

PR #686, opened 2023-11-28
Feature/build 2 ecr

GeorgesLorre:
I :
- created a public repo called `fndnt` ont the correct AWS account (the name reservation is pending)
- altered the `build_components.sh` script. So now It will create repo's if they do not exist yet and add a tag with the public ECR uri. (if you want to run it locally you will need to be logged in in Docker with AWS credentials)
- I ran the script locally for one component and without building to docker hub (only ECR) succesfully
- setup OIDC on the Public AWS account for github actions (using [1](https://www.automat-it.com/post/using-github-actions-with-aws-iam-roles)  [2](https://aws.amazon.com/blogs/security/use-iam-roles-to-connect-github-actions-to-actions-in-aws/))
- updated the github actions to include fetching credentials via OIDC and authenticating to the public ecr. (every AWS account can only have 1 public ECR so all aws commands using `aws ecr-public ...` will be ran against it.

I still need to test everything together in github itself.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -7,12 +7,11 @@ on:
   workflow_dispatch:
 
 env:
-  
   AWS_REGION : us-east-1
 
 permissions:
-      id-token: write
-      contents: read
+  id-token: write
+  contents: read
 
 jobs:
   docker:
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -8,7 +8,7 @@ on:
 
 env:
   
-  AWS_REGION : <"us-east-1">
+  AWS_REGION : us-east-1
 
 permissions:
       id-token: write   # This is required for requesting the JWT
@@ -45,7 +45,7 @@ jobs:
         with:
           role-to-assume: arn:aws:iam::904708148714:role/fondant-github-actions
           role-session-name: GitHub_to_AWS_via_FederatedOIDC
-          aws-region: ${{env.AWS_DEFAULT_REGION}}
+          aws-region: ${{env.AWS_REGION}}
 
       - name: Login to Amazon ECR Public
         id: login-ecr-public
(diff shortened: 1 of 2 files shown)


scripts/build_components.sh
```diff
@@ -73,12 +74,17 @@ for dir in "${components_to_build[@]}"; do
...
+    aws_full_image_names+=("$aws_full_image_name")
   done
 
   echo "Updating the image version in the fondant_component.yaml with:"
   echo "${full_image_names[0]}"
   sed -i -e "s|^image: .*|image: ${full_image_names[0]}|" fondant_component.yaml
 
+  # create repo if not exists
+  aws ecr-public describe-repositories --region us-east-1 --repository-names ${BASENAME} || aws ecr-public create-repository --region us-east-1 --repository-name ${BASENAME}
+  full_image_names+=("public.ecr.aws/m4f5a0g0/${BASENAME}:${tag}")
```
GeorgesLorre, line 86: This is temporary until we get the `fndnt` name reserved.

GeorgesLorre:
I will need to alter the other github action scripts aswel but want to test this first

.github/workflows/build.yaml
```diff
@@ -6,6 +6,14 @@ on:
       - main
   workflow_dispatch:
 
+env:
+
```
RobbeSneyders, line 10:
```suggestion
```

.github/workflows/build.yaml
```diff
@@ -6,6 +6,14 @@ on:
       - main
   workflow_dispatch:
 
+env:
+  
+  AWS_REGION : us-east-1
+
+permissions:
+      id-token: write
+      contents: read
```
RobbeSneyders, line 15:
```suggestion
  id-token: write
  contents: read
```

scripts/build_components.sh
RobbeSneyders: Can you update this now?

RobbeSneyders (approved):
Thanks @GeorgesLorre!

Feel free to merge after addressing the comments.

PR #691, opened 2023-11-29
Explorer search

PhilippeMoussalli:
PR that adds search functionality to the explorer 

image

other small improvements:
* Small bugfix regarding the page navigation logic
* Small changes to layout (short names, collapse sidebar information config)


RobbeSneyders:
Thanks @PhilippeMoussalli!

Looks good. One small request. Can we make it so pressing enter in the search box executes a search?

PhilippeMoussalli:
> Thanks @PhilippeMoussalli!
> 
> Looks good. One small request. Can we make it so pressing enter in the search box executes a search?

I'v included the change. Also fixed the graphviz visualization so it looks more aligned

RobbeSneyders:
Thanks!

Two more comments:
- Can you center the exact match / search / clear vertically?
  image
- This specific order of actions leads to strange behavior:
  - Search for something which doesn't lead to any results
  - Move to a different tab
  - Move back to the dataset explorer
  
  --> The search value is reset, but the data explorer is still empty. You need to search again before you can clear it. This is only the case for searches that don't lead to any results.

PhilippeMoussalli:
> Thanks!
> 
> Two more comments:
> 
> * Can you center the exact match / search / clear vertically?

There is no perfect way to fix this but It should look better now, I opted for [this solution](https://discuss.streamlit.io/t/vertically-center-the-content-of-columns/6163). They're not perfectly aligned but better than before 

image


> * This specific order of actions leads to strange behavior:
>   
>   * Search for something which doesn't lead to any results
>   * Move to a different tab
>   * Move back to the dataset explorer
>   
>   --> The search value is reset, but the data explorer is still empty. You need to search again before you can clear it. This is only the case for searches that don't lead to any results.

This should be fixed now, one other strange behavior I found is that toggling the exact match on/off triggers the search even if enter is not hit. That is because streamlit refreshes on every widget event change so in this case it will detect that a text is present in the search field (even if "Enter" is not submit) and trigger the search. There is no neat way to separate the two events. I reverted back to only allowing the search via the submit button. There is an open issue to remove the "Enter to apply" placeholder  to avoid confusion so we probably should do this at some point but it's not there yet.

https://github.com/streamlit/streamlit/issues/7538

docs/data_explorer.md
```diff
@@ -1,17 +1,49 @@
...
-The user can specify a pipeline and a specific pipeline run and component to explore. The user will then be able to explore the different subsets produced by by Fondant components.
+### General Overview
 
-The chosen subset (and the columns within the subset) can be explored in 3 tabs.
+In the general overview, you can select the pipeline and pipeline run you want to explore. You will
+be able to see the different components that were run in the pipeline run and get an overview of
+your latest runs.
 
-![data explorer](../art/data_explorer/data_explorer.png)
+![data explorer](art/data_explorer/general_overview.png)
```
RobbeSneyders, line 18:
```suggestion
![data explorer](../art/data_explorer/general_overview.png)
```

docs/data_explorer.md
```diff
@@ -1,17 +1,49 @@
...
+The data explorer shows an interactive table of the loaded fields from a given component.
+In this you can:
+
+- Browse through different parts of the data
+- Visualize images
+- Search for specific rows using a search query
+- Visualize long documents using a document viewer
+- Compare different pipeline runs (coming soon!)
+
+![data explorer](art/data_explorer/dataset_explorer.png)
```
RobbeSneyders, line 31:
```suggestion
![data explorer](../art/data_explorer/dataset_explorer.png)
```

RobbeSneyders (commented):
Ok, LGTM.

Only the images on the docs are broken.

PR #692, opened 2023-11-30
Set default explorer version in python sdk

RobbeSneyders:
Follow-up on #681 which fixes this for the Python SDK as well.


PhilippeMoussalli (approved):
Thanks Robbe!

PR #697, opened 2023-12-04
Add functionality for pullthrough cache rule creation and URI patching

GeorgesLorre:


RobbeSneyders (commented):
Thanks @GeorgesLorre, left some comments.

In general probably good to add docstrings with some information.

RobbeSneyders (approved):

PR #699, opened 2023-12-04
Support pipeline factory functions as CLI reference

RobbeSneyders:
In our RAG use case, [we want to start using a pipeline factory function](https://github.com/ml6team/fondant-usecase-RAG/pull/34#discussion_r1414114415), but this is currently not supported by the CLI.

This PR fixes that.

I also included a change (the first commit) which removes the `finally` blocks. We were using this as follows:
```python
try:
    ref = ...
except SpecificError:
    ref = ...
finally:
    run(ref)
```

But apparently the `finally` block is also entered if any non-`SpecificError` exception is raised in the `try` block. This lead to all non-handled errors to become an `UnboundLocal` error. This is now no longer the case, and the original error is not caught.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -132,12 +132,28 @@ def test_pipeline_from_module(module_str):
         "examples.example_modules.pipeline:create_pipeline('test_pipeline')",
         # Factory does not expect an argument
         "examples.example_modules.pipeline:create_pipeline(name='test_pipeline')",
+        # Invalid argument
+        "examples.example_modules.pipeline:create_pipeline(name)",
+        # Not a variable or function
+        "examples.example_modules.pipeline:[]",
+        # Attribute doesn't exist
+        "examples.example_modules.pipeline:no_pipeline",
+        # Attribute is not a valid python name
+        "examples.example_modules.pipeline:pipe;line",
+        # Not a Pipeline
+        "examples.example_modules.pipeline:number",
     ],
 )
 def test_pipeline_from_module_error(module_str):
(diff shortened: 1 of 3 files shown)


GeorgesLorre:
> But apparently the `finally` block is also entered if any non-`SpecificError` exception is raised in the `try` block. This lead to all non-handled errors to become an `UnboundLocal` error. This is now no longer the case, and the original error is not caught.

TIL

GeorgesLorre:
I'm going to review this tomorrow after a nights sleep 😉

tests/test_cli.py
```diff
@@ -122,12 +126,34 @@ def test_pipeline_from_module(module_str):
...
+        "examples.example_modules.pipeline:no_pipeline",
+        # Attribute is not a valid python name
+        "examples.example_modules.pipeline:pipe;line",
+        # Not a Pipeline
+        "examples.example_modules.pipeline:number",
     ],
 )
 def test_pipeline_from_module_error(module_str):
     """Test different error cases for pipeline_from_string."""
     with pytest.raises(PipelineImportError):
```
GeorgesLorre, line 149: can't you match the error messages here ? with pytest match

GeorgesLorre (commented):
Cool! Going to open a lot op possibilities

tests/test_cli.py
GeorgesLorre: ok to merge this then if it blocking

GeorgesLorre (approved):

PR #700, opened 2023-12-05
Add logic to handle custom components

GeorgesLorre:
Add branches to make sure we only pullthrough and patch reusable components.


RobbeSneyders (approved):
:+1:

PR #704, opened 2023-12-06
Add compile and run interface to kfp based runners

GeorgesLorre:


src/fondant/pipeline/runner.py
```diff
@@ -92,11 +97,37 @@ def _resolve_imports(self):
...
+                "Found reference to un-compiled pipeline... compiling",
+            )
+            compiler = KubeFlowCompiler()
+            compiler.compile(
+                input,
+                output_path=output_path,
+            )
+            self._run(output_path)
+        else:
+            self._run(input)
```
RobbeSneyders, line 124: I think we should pass the `experiment_name` here as well?

src/fondant/pipeline/runner.py
```diff
@@ -92,11 +97,37 @@ def _resolve_imports(self):
             )
 
     def run(
```
RobbeSneyders, line 99: Can't help but notice that the flow of this method is the same for every runner. Might be worth it to lift it to the base `Runner` class. Up to you.

GeorgesLorre: I had the same idea when commuting home, will also make testing a lot easier

PR #710, opened 2023-12-08
Add support for resources with Sagemaker

GeorgesLorre:
You can now define resources on a component level for SageMaker:

```
images = raw_data.apply(
    "download_images",
    arguments={
        "input_partition_rows": 100,
        "resize_mode": "no",
    },
    resources=Resources(instance_type="ml.t3.xlarge"),
)
```

Only the instance_type is supported for SageMaker (for now)


RobbeSneyders (approved):
Thanks, we might want to split the Resources for the different runners in the future I think.

PR #712, opened 2023-12-10
Fix/sagemaker script generator

GeorgesLorre:
Hopefully fixed by:

https://docs.python.org/3/library/shlex.html#shlex.quote

Still need to test on aws

### DIFF
--- a/src/fondant/pipeline/compiler.py
+++ b/src/fondant/pipeline/compiler.py
@@ -1,6 +1,7 @@
 import json
 import logging
 import os
+import shlex
 import tempfile
 import typing as t
 from abc import ABC, abstractmethod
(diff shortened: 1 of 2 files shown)


RobbeSneyders:
Thanks @GeorgesLorre!

Maybe we should add `bash -n scriptname` to the test to validate the generated script?

GeorgesLorre:
> Maybe we should add `bash -n scriptname` to the test to validate the generated script?

It does call `fondant execute ...` so not sure what the easiest way is to call this in the test

RobbeSneyders:
`bash -n` just checks the script, [it doesn't execute it](https://tldp.org/LDP/abs/html/options.html#:~:text=%2Dn,them%20(syntax%20check)).

RobbeSneyders (approved):

PR #715, opened 2023-12-11
Add evaluate_ragas component

Hakimovich99:
Putting it as a draft because will probably change the base Dockerfile for one that already has torch dependencies, instead of installing them in the requirements.txt


components/retriever_eval_ragas/requirements.txt
```diff
@@ -0,0 +1,5 @@
+ragas==0.0.20
+transformers==4.35.2
+torch==2.1.1
+torchaudio==2.1.1
+torchvision==0.16.1
```
RobbeSneyders, line 5: Since we only use ragas explicitly, shouldn't these dependencies be installed by ragas?

RobbeSneyders (approved):

PR #716, opened 2023-12-11
Add agg_eval_results component

Hakimovich99:


RobbeSneyders:
Let's keep this component in the use case repo, since it's quite specific (it only calculates the mean).

PR #717, opened 2023-12-11
Move to modules instead of classes for home page and common interface

RobbeSneyders:
This PR simplifies the data explorer implementation a bit by moving the home page and common interfaces into modules instead of a long hierarchy of classes with a singleton class at the top.

I think we should do the same for the dataset views, but that would increase the size of the PR, so I suggest to do it later in a separate one.


data_explorer/app/config.py
```diff
@@ -1,12 +1,9 @@
 """General configuration for the app."""
 
 SESSION_STATE_VARIABLES = [
```
RobbeSneyders, line 3: We should aim to get rid of these completely. 

This is currently used to initialize all session state variables at the start of the application, to prevent errors when accessing non-initialized values. But this brings us in broken states anyway. We should make sure the flow is set up so values are only accessed after initialization, or non-initialization is handled in the accessing code.

data_explorer/app/interfaces/common_interface.py
RobbeSneyders, line 1: This file was split into `common.py` and `sidebar.py`

data_explorer/app/interfaces/utils.py
RobbeSneyders, line 1: This is a fix for an issue raised by @Hakimovich99, where you cannot switch pipelines because the run selection would error.

data_explorer/app/main.py
```diff
@@ -1,165 +1,43 @@
...
+from interfaces.sidebar import render_sidebar
+from pages.home import render_pipeline_overview
 from st_pages import show_pages_from_config
 
-LOGGER = logging.getLogger(__name__)
-
 dask.config.set({"dataframe.convert-string": False})
 
 
-class PipelineOverviewApp(MainInterface):
```
RobbeSneyders, line 21: Moved to home.py

src/fondant/pipeline/runner.py
RobbeSneyders, line 1: Extra logs since the docker compose logs don't show on Windows for Jan.

PR #718, opened 2023-12-11
Reactivate readme generation pre-commit

RobbeSneyders:
This was temporarily deactivated since the specification on the main branch was out of sync. We still need to update the generic components to generate a more useful README, but that's for a separate PR.

Devs will need to reset their local pre-commit environments for this to work.


mrchtr (approved):
Thanks! Re-installed and tested the pre-commit locally. Works on my end.

PR #721, opened 2023-12-12
Fix data io for generic PandasTransformComponent

RobbeSneyders:
I ran into an issue when using the generic `evaluate_ragas` component. 

```Python
component = RetrieverEval(
    module="langchain.llms",
    llm_name="OpenAI",
    llm_kwargs={"openai_api_key": os.environ["OPENAI_KEY"]},
    produces={
        "context_precision": pa.float32(),
        "context_relevancy": pa.float32(),
    },
)
```
Lead to the following error:
```
ValueError: Fields ['context_precision', 'context_relevancy'] defined in output dataset but not found in dataframe
```

Because we were not using the `operation_spec` yet for the `PandasTransformComponent`. I think we can reduce the chance of similar errors in the future by only passing the operation_spec to the component, but then we'll need to include the other information from the component spec in there as well.

### DIFF
--- a/src/fondant/component/executor.py
+++ b/src/fondant/component/executor.py
@@ -514,7 +514,11 @@ class PandasTransformExecutor(TransformExecutor[PandasTransformComponent]):
         return ["input_manifest_path", "input_partition_rows"]
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


mrchtr (approved):
Thanks @RobbeSneyders!

PR #722, opened 2023-12-12
Fix environment variable setting in embed_text component

Hakimovich99:
Set the environment key before calling the embedding model to avoid "no api key found" error.


RobbeSneyders (approved):
Thanks @Hakimovich99!

PR #731, opened 2023-12-18
Make data explorer non-blocking with docker compose

PhilippeMoussalli:
Small PR to avoid showing the localhost url before the image is build

### DIFF
--- a/src/fondant/explore.py
+++ b/src/fondant/explore.py
@@ -115,9 +115,9 @@ def run_explorer_app(  # type: ignore
             ["--base_path", f"{base_path}"],
         )
 
+    subprocess.call(cmd, stdout=subprocess.PIPE)  # nosec
+
     logging.info(
         f"Running image from registry: {container} with tag: {tag} on port: {port}",
     )
     logging.info(f"Access the explorer at http://localhost:{port}")
-
-    subprocess.call(cmd, stdout=subprocess.PIPE)  # nosec
--- a/src/fondant/explore.py
+++ b/src/fondant/explore.py
@@ -3,35 +3,34 @@ import os
 import shlex
 import subprocess  # nosec
 import typing as t
+from dataclasses import asdict
 from importlib.metadata import version
 from pathlib import Path
 
 import fsspec.core
+import yaml
 from fsspec.implementations.local import LocalFileSystem
 
+from fondant.core.schema import DockerVolume
 
-# type: ignore
-def run_explorer_app(  # type: ignore
+CONTAINER = "fndnt/data_explorer"
+PORT = 8501
+OUTPUT_PATH = ".fondant/explorer-compose.yaml"
+
+
+def _get_service_name(container: str) -> str:
+    return os.path.basename(container)
+
+
+def _generate_explorer_spec(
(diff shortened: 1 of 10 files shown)


RobbeSneyders:
The `docker run` call is blocking, so with this change the url will only be shown after you stop the explorer again. I think the only way to improve on the current situation is by making the `docker run` command non-blocking, which would probably be an improvement for usage in the notebooks anyway. We might want to use docker compose then instead of docker so the explorer is only started once and easy to bring down.

PhilippeMoussalli:
> We might want to use docker compose then instead of docker so the explorer is only started once and easy to bring down.

Makes sense indeed, it's always a bit clunky to stop it from the notebook. I'll update the PR

PhilippeMoussalli:
> > We might want to use docker compose then instead of docker so the explorer is only started once and easy to bring down.
> 
> Makes sense indeed, it's always a bit clunky to stop it from the notebook. I'll update the PR

Updated the PR, the explorer now uses docker compose to launch the service, slight change to the cli command: 

starting the service:

```
fondant explore start -b <base_path>
```

Stopping the service: 

```
fondant explore stop
```

RobbeSneyders (commented):
Thanks @PhilippeMoussalli!

I had it a bit different in my mind, thinking about a docker compose file which we include in the Fondant package. But this approach indeed makes more sense so the compose file can be generated with different options :+1: 

The approach and code looks good to me, but I still want to test it. Will probably do so tomorrow!

src/fondant/explore.py
```diff
@@ -97,27 +76,106 @@ def run_explorer_app(  # type: ignore
...
+        output_path,
+        "up",
+        "--build",
+        "--pull",
+        "always",
+        "--remove-orphans",
+        "--detach",
+    ]
+
+    subprocess.call(cmd, stdout=subprocess.PIPE)  # nosec
```
RobbeSneyders, line 151: I think we should use [`check_call`](https://docs.python.org/3/library/subprocess.html#subprocess.check_call) here so we don't print the messages below if this fails (eg. when the docker daemon is not running).

RobbeSneyders (commented):
Just tested @PhilippeMoussalli. Works well, I would just catch errors with the docker command, see my comment below.

src/fondant/explore.py
```diff
@@ -148,7 +148,10 @@ def run_explorer_app(  # type: ignore  # noqa: PLR0913
         "--detach",
     ]
 
-    subprocess.call(cmd, stdout=subprocess.PIPE)  # nosec
+    try:
+        subprocess.check_call(cmd, stdout=subprocess.PIPE)  # nosec
+    except subprocess.CalledProcessError as e:
+        raise OSError(e.returncode)
```
RobbeSneyders, line 154: Why did you choose an `OSError` here? I would probably have gone for a `SystemExit` exception myself.

PhilippeMoussalli: Might make more sense here indeed, updated

RobbeSneyders (approved):
Thanks @PhilippeMoussalli!

PR #732, opened 2023-12-18
Reduce pulling logs

PhilippeMoussalli:
PR to reduce pulling logs when running docker compose since they can clutter the notebooks where there are many images to pull

### DIFF
--- a/src/fondant/cli.py
+++ b/src/fondant/cli.py
@@ -526,6 +526,12 @@ def register_run(parent_parser):
         action="append",
         help="Build arguments for `docker build`",
     )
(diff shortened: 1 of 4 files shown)


RobbeSneyders:
Does this remove the pulling logs completely? Because I also don't want it to look like it's hanging to the user.

PhilippeMoussalli:
> Does this remove the pulling logs completely? Because I also don't want it to look like it's hanging to the user.

Just tested it and it unfortunately does, I though it would just be less verbose. Closing this ticket for now.

PR #734, opened 2023-12-19
Add logs to weaviate

PhilippeMoussalli:
Add logs to give indication on indexing progress

### DIFF
--- a/components/index_weaviate/src/main.py
+++ b/components/index_weaviate/src/main.py
@@ -4,6 +4,7 @@ import typing as t
 import dask.dataframe as dd
 import weaviate
 from fondant.component import DaskWriteComponent
+from tqdm import tqdm
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


components/index_weaviate/src/main.py
```diff
@@ -54,9 +55,13 @@ def __init__(
...
         with self.client.batch as batch:
-            for part in dataframe.partitions:
+            for part in tqdm(
+                dataframe.partitions,
+                desc="Processing partitions",
+                total=dataframe.npartitions,
+            ):
                 df = part.compute()
-                for row in df.itertuples():
+                for row in tqdm(df.itertuples(), desc="Processing rows", total=len(df)):
```
RobbeSneyders, line 64: Have you tested this?
Providing progress on the row level might be a bit much. I guess partitions would be enough?

RobbeSneyders (approved):

components/index_weaviate/src/main.py
PhilippeMoussalli: Yeah indeed this just adds in another bar but the process time per partition is generally not that long, removed it

Issue #566, opened 2023-10-29
Implement scaling across multiple GPUs


RobbeSneyders:
Conclusions from https://github.com/ml6team/fondant/pull/489

Need to find a way to scale across GPUs, possible options:

* Multiple GPUs can be loaded for inference using pytorch [Data Parallelism](https://pytorch.org/tutorials/beginner/blitz/data_parallel_tutorial.html) (this does not work on every model) in order to parallelize the batches across multiple GPUs. One important consideration there is to use either a single threaded scheduler (not recommended) or to limit the number of workers to be the same as the number of GPU cores dask.config.set(num_workers=<#GPU>) to avoid running into issues. Other alternatives could include assigning GPUs to spawned processes (not tested yet).

* Other option could be to test out the [LocalCUDACluster] which seems like the intended way to run GPU components with Dask. Requires testing it and how it interacts with pytorch

* Whether to run a model using the processes or threaded scheduler (so far, the threaded scheduler has shown to be faster). However, most resources seem to indicate to use threads ([link](https://aws.amazon.com/blogs/machine-learning/parallelizing-across-multiple-cpu-gpus-to-speed-up-deep-learning-inference-at-the-edge/)).

Open questions: 
* How to parallelize GPU and CPU tasks efficiently: limiting the number of workers can leave some workers/CPU cores idle (when #GPU in one machine is larger than the number of cores). There is some room for optimization.

RobbeSneyders:
Interesting resource: [Intro to distributed computing on GPUs with Dask in Python](https://gist.github.com/jacobtomlinson/6f16abb716f50f81a6687bd67efd2f61)

PR #735, opened 2023-12-19
Fix illegal memory access embedding

PhilippeMoussalli:
PR that fixes issue that arises with error 139 when running embedding component with large number of rows (~100k). The issue seems to stem from multiple processes trying to access the GPU at the same time. 

This should be a temporary fix until [#566](https://github.com/ml6team/fondant/issues/566) is addressed 

more info [here](https://docs.google.com/document/d/1ugiMhf2zeLQtS9Psd4m4SIlwDl4uCWba5vW-hsSO1yo/edit)

### DIFF
--- a/components/embed_text/src/main.py
+++ b/components/embed_text/src/main.py
@@ -1,6 +1,7 @@
 import logging
 import os
 
+import dask
 import google.cloud.aiplatform as aip
 import pandas as pd
 from fondant.component import PandasTransformComponent
@@ -15,6 +16,8 @@ from langchain.schema.embeddings import Embeddings
 from retry import retry
 
 logger = logging.getLogger(__name__)
+
+dask.config.set(scheduler="processes")
 
 
 def to_env_vars(api_keys: dict):


components/embed_text/src/main.py
```diff
@@ -16,6 +17,8 @@
 
 logger = logging.getLogger(__name__)
 
+dask.config.set(scheduler="processes")
```
RobbeSneyders, line 20: Can we use the `LocalCluster` instead?

I thought it was used by default, but I think there is a bug. The docstrings say the default for `cluster_type` should be "local", but it seems to be "default" instead.

mrchtr: Indeed seems to be that the wrong default was incorporated with the new pipeline interface changes.  
I think we can set the cluster type to local again. But it will probably not fix the issue here, cause the issue was occurring with the LocalCluster already, no?

RobbeSneyders: I don't think this bug was introduced with the interface changes. I think it was always there. I assume the issue triggering this PR is due to the usage of threads (since processes fixes it), while the local cluster uses processes by default as well.

PhilippeMoussalli: should we then change the default `cluster_type` from `default` to `local` in the component spec or just explicitly change the cluster type for this component?

PhilippeMoussalli: [local_cluster.txt](https://github.com/ml6team/fondant/files/13742771/local_cluster.txt)

So using the local cluster does seem to work since it uses processes. One small downside it's that it's currently very verbose and returns a lot of warnings compared to just explicitly setting the scheduler to use processes instead  (see logs), but it did manage to complete successfully. 

I'm fine with either solutions for the moment, any preferences?

PhilippeMoussalli:
Closing this PR in favor of defining cluster type in this [PR](https://github.com/ml6team/fondant-usecase-RAG/pull/44)

PhilippeMoussalli:
Closing this PR in favor of defining cluster type in this [PR](https://github.com/ml6team/fondant-usecase-RAG/pull/44)

PR #736, opened 2023-12-19
Fix retry mechanism embedding component 

PhilippeMoussalli:
Reduce from infinite retires to a fix number

### DIFF
--- a/components/embed_text/fondant_component.yaml
+++ b/components/embed_text/fondant_component.yaml
@@ -40,6 +40,10 @@ args:
       Additional keyword arguments required for api initialization/authentication.
     type: dict
     default: {}
+  tries:
+    description: Number of retries to attempt when an embedding request fails.
+    type: int
+    default: 5
 
   
   
--- a/components/embed_text/fondant_component.yaml
+++ b/components/embed_text/fondant_component.yaml
@@ -40,7 +40,7 @@ args:
       Additional keyword arguments required for api initialization/authentication.
     type: dict
     default: {}
-  tries:
+  retries:
     description: Number of retries to attempt when an embedding request fails.
     type: int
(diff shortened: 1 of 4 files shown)


components/embed_text/src/main.py
```diff
@@ -61,9 +63,13 @@ def get_embedding_model(
...
 
-    @retry()  # make sure to keep trying even when api call limit is reached
+    # make sure to keep trying even when api call limit is reached
     def get_embeddings_vectors(self, texts):
-        return self.embedding_model.embed_documents(texts.tolist())
+        @retry(tries=self.tries, logger=logger)
+        def return_embeddings():
+            return self.embedding_model.embed_documents(texts.tolist())
+
+        return return_embeddings()
```
RobbeSneyders, line 72: FYI, you can just use a decorator like a normal function.

```suggestion
        retry_wrapper = retry(tries=self.tries, logger=logger)
        return retry_wrapper(self.embedding_model.embed_documents(texts.tolist()))
```

RobbeSneyders (commented):
Thanks @PhilippeMoussalli, we should indeed not do this endlessly.

components/embed_text/src/main.py
PhilippeMoussalli: 🤯

components/embed_text/fondant_component.yaml
```diff
@@ -40,6 +40,10 @@ args:
       Additional keyword arguments required for api initialization/authentication.
     type: dict
     default: {}
+  tries:
```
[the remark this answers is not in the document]
PhilippeMoussalli: sure changed it :)

RobbeSneyders (approved):

PR #738, opened 2023-12-20
Minor adjustments pipeline docstring

mrchtr:
Some days ago I've added examples to the pipeline interface docstring. I think it might help to access the documentation right in your IDE and could help to understand the interface.


src/fondant/pipeline/pipeline.py
```diff
@@ -558,10 +558,72 @@ def apply(
...
+                defined as follows:
+
+                ```
+                consumes = {
+                    "text": "custom_text"
+                }
+                ```
+
+                In this example, the 'text' field will be sourced from 'custom_text' and 'image'
+                will be sourced from the 'image' field.
```
PhilippeMoussalli, line 585:
```suggestion
                will be sourced from the 'image' field by default sine it's not specified in the custom mapping.
```

src/fondant/pipeline/pipeline.py
```diff
@@ -558,10 +558,72 @@ def apply(
...
+                mapping can be defined as follows:
+
+                ```
+                produces = {
+                    "width": "custom_width",
+                }
+                ```
+
+                In this example, the 'text' field will retain as text, while the 'width' field will
+                be stored with the name 'custom_width' in the output dataset.
```
PhilippeMoussalli, line 613: Same here as above

src/fondant/pipeline/pipeline.py
```diff
@@ -558,10 +558,72 @@ def apply(
...
+                Alternatively, the produces defines the data type of the output data.
+
+                ```
+                produces = {
+                    "width": pa.float32(),
+                }
+                ```
+
+                In this example, the 'text' field will retain its type 'string' without specifying a
+                different source, while the 'width' field will be produced as type float in the
```
PhilippeMoussalli, line 624:
```suggestion
                different source, while the 'width' field will be produced as type `float` in the
```

PhilippeMoussalli (approved):
Nice thanks Matthias! Left a few small comments

PR #740, opened 2023-12-21
Add component to index aws opensearch

shub-kris:
This PR aims to add support for indexing to AWS OpenSearch.


components/index_aws_opensearch/src/main.py
```diff
@@ -0,0 +1,61 @@
...
+        """
+        response = self.client.indices.create(self.index_name, body=index_body)
+
+    def write(self, dataframe: dd.DataFrame):
+        """
+        Writes the data from the given Dask DataFrame to AWS OpenSearch Index.
+        Args:
+            dataframe (dd.DataFrame): The Dask DataFrame containing the data to be written.
+        """
+        if not self.client.indices.exists(index=self.index_name):
```
mrchtr, line 53: I think we could call the `create_index` here if the index doesn't exist.

mrchtr (commented):
Thanks @shub-kris! Looks already quite promising. Could you try to add a component test as well? You could have a look into the write to Qdrant component. There was something similar implemented.

components/index_aws_opensearch/src/main.py
shub-kris: Sure, we can do that

components/index_aws_opensearch/src/main.py
```diff
@@ -0,0 +1,61 @@
...
+        )
+        self.create_index(index_body)
+
+    def create_index(self, index_body: Dict[str, Any]):
+        """Creates an index in AWS OpenSearch
+
+        Args:
+            index_body (Dict[str, Any]): Parameters that specify index settings, mappings, and aliases for newly created index.
+        """
+        response = self.client.indices.create(self.index_name, body=index_body)
```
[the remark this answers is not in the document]
shub-kris: Will add it then

components/index_aws_opensearch/src/main.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+class IndexAWSOpenSearchComponent(DaskWriteComponent):
+    def __init__(
+        self,
+        host: str,
+        region: str,
+        index_name: str,
+        index_body: Dict[str, Any],
+        port: int = 443,
```
PhilippeMoussalli, line 16: I would not include the default types here for the optional argument since you already define them in the spec and it can be misleading (they are not actually used here it will always default to the ones defined in the spec). You can instead define their types as optional just so that it's clear that they do have default values and they don't need to be explicitly defined. Similar to [here](https://github.com/ml6team/fondant/blob/main/components/load_from_hf_hub/src/main.py)

components/index_aws_opensearch/src/main.py
```diff
@@ -0,0 +1,65 @@
...
+            pool_maxsize=pool_maxsize,
+            **kwargs,
+        )
+        self.create_index(index_body)
+
+    def create_index(self, index_body: Dict[str, Any]):
+        """Creates an index in AWS OpenSearch.
+
+        Args:
+            index_body (Dict[str, Any]): Parameters that specify index settings,
```
PhilippeMoussalli, line 41:
```suggestion
            index_body: Parameters that specify index settings,
```

Our convention, we only define data types in the argument and not the docstring (single source of truth)

components/index_aws_opensearch/src/main.py
```diff
@@ -0,0 +1,65 @@
...
+            mappings, and aliases for newly created index.
+        """
+        self.client.indices.create(self.index_name, body=index_body)
+
+    def write(self, dataframe: dd.DataFrame):
+        """
+        Writes the data from the given Dask DataFrame to AWS OpenSearch Index.
+
+        Args:
+            dataframe (dd.DataFrame): The Dask DataFrame containing the data to be written.
```
PhilippeMoussalli, line 51:
```suggestion
            dataframe: The Dask DataFrame containing the data to be written.
```

components/index_aws_opensearch/src/main.py
```diff
@@ -0,0 +1,65 @@
...
+        """
+        if not self.client.indices.exists(index=self.index_name):
+            msg = f"Index: {self.index_name} doesn't exist. Please Create"
+            raise ValueError(msg)
+
+        for part in dataframe.partitions:
```
PhilippeMoussalli, line 57:
```suggestion
            for part in tqdm(
                dataframe.partitions,
                desc="Processing partitions",
                total=dataframe.npartitions,
            ):
```

Useful to add logs in case the dataset is large. Similar to https://github.com/ml6team/fondant/blob/main/components/index_weaviate/src/main.py

Don't forget to add tqdm to the list of requirements

components/index_aws_opensearch/requirements.txt
```diff
@@ -0,0 +1,2 @@
+boto3==1.34.4
```
PhilippeMoussalli, line 1: Add 

tqdm==4.65.0

(see below)

components/index_aws_opensearch/src/main.py
shub-kris: It's not needed anymore as I am doing it inside the `init`.

shub-kris: Removed the Exception too

RobbeSneyders:
FYI, there's another test example in the `index_weaviate` component`.

You can install `pre-commit` to take care of all the static checks for you (which are the reason the pipeline is currently failing). Just execute the following 3 commands:
```
pip install poetry
poetry install --all-extras
pre-commit install
```

components/index_aws_opensearch/src/main.py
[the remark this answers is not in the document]
shub-kris: Added tests too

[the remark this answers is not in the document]
PhilippeMoussalli: Thanks :) 

I think you still need to modify the Docker image to be able to run the test with it similar to [this](https://github.com/ml6team/fondant/blob/main/components/retrieve_from_weaviate/Dockerfile) so that you can run tests using 
```
docker build . --target test
```

Could you please update and test?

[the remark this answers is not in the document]
shub-kris: Updated and tested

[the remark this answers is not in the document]
PhilippeMoussalli: great :) 

did you manage to fix the failing pipelines?

after installing pre-commit you should run 

```
pre-commit run --all-files
```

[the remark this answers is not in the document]
shub-kris: Yupp

PhilippeMoussalli (approved):
Thanks @shub-kris! Nice first contribution

components/index_aws_opensearch/tests/aws_opensearch_test.py
```diff
@@ -0,0 +1,45 @@
...
+    """
+    Test case for the write method of the IndexAWSOpenSearchComponent class.
+
+    This test creates an index , inserts dummy data and then deletes the index.
+    Writes data to it using the write method of the IndexAWSOpenSearchComponent.
+    Asserts that the count of entries in the index is equal to the expected number of entries.
+    """
+    index_name = "pytest-index"
+    aws_os_comp = IndexAWSOpenSearchComponent(
+        host="search-genai-vectordb-domain-f7vxqkogveaie2qdrivnkr66om.eu-west-1.es.amazonaws.com",
```
RobbeSneyders, line 21: Is there a way to test this locally instead?

RobbeSneyders: Otherwise it might be better to mock this.

shub-kris: Mocking is the only solution to test it without having access to AWS OpenSearch Cluster.

RobbeSneyders: Yes indeed, so I would prefer that. Currently I cannot test this component myself. The goal of these tests is to be unit tests, we have other mechanism for integration testing.

shub-kris: Will do that and update the PR

RobbeSneyders (approved):
Thanks @shub-kris!

PR #741, opened 2023-12-21
Unify operation and component spec

### LINKED ISSUE #728: Unify OperationSpec and ComponentSpec
We currently pass `OperationSpec`, `ComponentSpec`, `produces`, and `consumes` to the components. Each containing a piece of the puzzle to read, transform, and write the correct data. We should be able to combine them into a more unified view so it's more clear how everything fits together.

PhilippeMoussalli:
Fixes #728 

We now construct the operation spec once during the compilation process, serialize it and pass it to the component as an argument. No need to pass in the `produces` and `consumes` arguments anymore. 

Some attributes like the arguments and the previous index that are part of the component spec can still be accessed through the operation spec since the component spec is it's attribute.

### DIFF
--- a/src/fondant/core/component_spec.py
+++ b/src/fondant/core/component_spec.py
@@ -299,7 +299,7 @@ class OperationSpec:
         consumes: t.Optional[t.Dict[str, t.Union[str, pa.DataType]]] = None,
         produces: t.Optional[t.Dict[str, t.Union[str, pa.DataType]]] = None,
     ) -> None:
-        self.component_spec = component_spec
+        self._component_spec = component_spec
 
         self._mappings = {
             "consumes": consumes,
@@ -326,7 +326,7 @@ class OperationSpec:
             return serialized_mapping
 
         specification_dict = {
-            "specification": self.component_spec.specification,
+            "specification": self._component_spec.specification,
             "consumes": _dump_mapping(self._mappings["consumes"]),
             "produces": _dump_mapping(self._mappings["produces"]),
         }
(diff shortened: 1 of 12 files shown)


src/fondant/pipeline/pipeline.py
RobbeSneyders, line 1: I think we can remove the `_dump_mapping` method from this class as well since this is now part of the `OperationSpec` (and the main reason for the decreased coverage).

src/fondant/core/component_spec.py
```diff
@@ -313,12 +294,12 @@ class OperationSpec:
...
     def __init__(
         self,
-        specification: ComponentSpec,
+        component_spec: ComponentSpec,
         *,
         consumes: t.Optional[t.Dict[str, t.Union[str, pa.DataType]]] = None,
         produces: t.Optional[t.Dict[str, t.Union[str, pa.DataType]]] = None,
     ) -> None:
-        self.specification = specification
+        self.component_spec = component_spec
```
RobbeSneyders, line 302:
```suggestion
        self._component_spec = component_spec
```

Can we make this private and expose all the attributes that are used publicly as properties? That way this class is more in control of how it's used, and linters will complain if the class is used incorrectly.

RobbeSneyders (commented):
Thanks @PhilippeMoussalli!

These changes already look good to me. I would just go one step further by making the `component_spec` attribute on the `OperationSpec` private as explained in my comment below.

src/fondant/core/component_spec.py
PhilippeMoussalli: Make sense :) added the changes, now the component spec is completely hidden from the executor

src/fondant/pipeline/pipeline.py
PhilippeMoussalli: good catch, forgot about that one

RobbeSneyders (approved):

PR #742, opened 2023-12-21
Add availability check for docker and docker compose.

### LINKED ISSUE #574: Test and validate docker versions
We see users running into issues with docker when trying to run their first fondant pipeline. We should add tests for different docker versions to validate which ones work, and check for them when running Fondant, raising a clear error when working with an unsupported version.

mrchtr:
fix #574


src/fondant/pipeline/runner.py
```diff
@@ -78,6 +80,24 @@ def run(
...
+        try:
+            # Check Docker info
+            subprocess.call(["docker", "info"])  # nosec
+
+            # Check Docker Compose version
+            subprocess.call(["docker", "compose", "version"])  # nosec
+
+        except subprocess.CalledProcessError:
+            msg = (
+                "Docker or Docker Compose is not installed or not running. Please make sure "
```
PhilippeMoussalli, line 94: Can we maybe separate the errors for each to be more specific on what needs to be installed?

src/fondant/pipeline/runner.py
```diff
@@ -61,22 +61,57 @@ def run(
...
+    def docker_is_installed():
+        """Execute docker command to check if docker is available."""
+        try:
+            # Check Docker info
+            subprocess.call(["docker", "info"])  # nosec
+            return True
+
+        except subprocess.CalledProcessError:
+            msg = (
+                "Docker or Docker Compose is not installed or not running. Please make sure "
```
PhilippeMoussalli, line 92:
```suggestion
                "Docker is not installed or not running. Please make sure "
```

src/fondant/pipeline/runner.py
```diff
@@ -61,22 +61,57 @@ def run(
...
+        """Execute docker compose command to check if docker is available."""
+        try:
+            # Check Docker info
+            subprocess.call(["docker", "compose", "version"])  # nosec
+            return True
+
+        except subprocess.CalledProcessError:
+            msg = (
+                "Docker Compose is not installed or not running. Please make sure "
+                "Docker Compose is installed."
```
PhilippeMoussalli, line 110: Maybe we can the link to the installation instruction here as well

https://fondant.ai/en/latest/guides/installation/#docker-installation

PhilippeMoussalli (approved):

src/fondant/pipeline/runner.py
```diff
@@ -61,22 +61,61 @@ def run(
...
+                self._run(output_path)
+            else:
+                self._run(input)
+
+    @staticmethod
+    def docker_is_installed():
+        """Execute docker command to check if docker is available."""
+        try:
+            # Check Docker info
+            subprocess.call(["docker", "info"])  # nosec
```
RobbeSneyders, line 87: Don't we need to use `check_call` here for it to throw a `CalledProcessError`?

src/fondant/pipeline/runner.py
```diff
@@ -61,22 +61,61 @@ def run(
...
+            return True
```
RobbeSneyders, line 88: Nit, but would move this outside of the `try: except`.

src/fondant/pipeline/runner.py
```diff
@@ -61,22 +61,61 @@ def run(
...
+                "Find more details on the Docker installation here: "
+                "https://fondant.ai/en/latest/guides/installation/#docker-installation"
             )
-            compiler = DockerCompiler()
-            compiler.compile(
-                input,
-                output_path=output_path,
-                extra_volumes=extra_volumes,
-                build_args=build_args,
+            raise OSError(
```
RobbeSneyders, line 97: Does this produce a stacktrace? I think we shouldn't.

RobbeSneyders (commented):
Thanks @mrchtr!

src/fondant/pipeline/runner.py
```diff
@@ -66,22 +67,62 @@ def run(
...
             )
-            compiler = DockerCompiler()
-            compiler.compile(
-                input,
-                output_path=output_path,
-                extra_volumes=extra_volumes,
-                build_args=build_args,
+
+        except subprocess.CalledProcessError:
+            sys.exit(
```
GeorgesLorre, line 99: Is there a reason to use `sys.exit()` and not raise the error ?

tests/pipeline/test_runner.py
```diff
@@ -24,8 +24,31 @@
 
 def test_docker_runner():
```
GeorgesLorre, line 25: Can we add a separate test that checks the errors ?

https://github.com/ml6team/fondant/blob/main/tests/test_explorer.py#L164

src/fondant/pipeline/runner.py
RobbeSneyders: Haha, I proposed this in a previous review. I think it's more user friendly to show a single error message instead of a full stack trace. At least for the CLI. For the SDK I could go either way.

If we want to have a stacktrace in the SDK, we could raise the error and catch it in the CLI instead.

tests/pipeline/test_runner.py
mrchtr: I've added two separate test cases for both situations, docker and docker compose is missing.

src/fondant/pipeline/runner.py
```diff
@@ -66,22 +67,62 @@ def run(
             to mount in the docker-compose spec.
             build_args: List of build arguments to pass to docker
         """
-        if isinstance(input, Pipeline):
-            os.makedirs(".fondant", exist_ok=True)
-            output_path = ".fondant/compose.yaml"
-            logging.info(
-                "Found reference to un-compiled pipeline... compiling",
+        if self.docker_is_installed() and self.docker_compose_is_installed():
```
RobbeSneyders, line 70: This is a bit of a strange construct with an `if` and no `else`. Reading this, I would expect us to just do nothing if `docker` or `docker compose` is not installed, while I see that we actually raise an error in those methods.

So I would either move the error messages into an `else` block here, or don't use them in an `if` block (and rename them):
```python
self.check_docker_install()
self.check_docker_compose_install()

...
```

GeorgesLorre:
```
try:
   self.check_docker_install()
   self.check_docker_compose_install()
except:
    ?
```
no ?

mrchtr: I've thought it adds some readability. I removed the if block and renamed the functions.

RobbeSneyders (approved):
Thanks!

PR #745, opened 2023-12-22
Consistent component naming

PhilippeMoussalli:
Fixes https://github.com/ml6team/fondant-use-cases/issues/52

We might need to remove some of the old component names manually from the hub at some point since we'll end up with duplicate components under different names. This should only be done after updating all the new examples to the latest release


RobbeSneyders (approved):
Thanks @PhilippeMoussalli 

I think we can leave the old images for now for backwards compatibility. Would only remove them in a couple of versions time.

PR #746, opened 2023-12-22
Add named links to hub page

PhilippeMoussalli:
Fixes https://github.com/ml6team/fondant-use-cases/issues/51


RobbeSneyders:
Thanks @PhilippeMoussalli.

Unfortunately, this doesn't render nicely on Github. What do you think about adding the named link to the hub template instead? Then we only get one named link per component, but I think that should be enough.

PhilippeMoussalli:
> Thanks @PhilippeMoussalli.
> 
> Unfortunately, this doesn't render nicely on Github.  

Can you maybe elaborate more on the original issue? I think originally it was just moving from numbered references to named ones which is now currently the case or is there something else I missed? 

> What do you think about adding the named link to the hub template instead? Then we only get one named link per component, but I think that should be enough.  

I think the hub template only has admonitions and no links defined (since it has no headers and the headers are only defined in the readme).

RobbeSneyders:
> Can you maybe elaborate more on the original issue? I think originally it was just moving from numbered references to named ones which is now currently the case or is there something else I missed?

The issue is that when linking to components in the hub, we currently need to use a dynamic link ([example](https://github.com/ml6team/fondant/blob/main/docs/guides/build_a_simple_pipeline.md#:~:text=use%20the%20reusable-,load_from_hf_hub,-component.)) which breaks when adding new components.

This PR fixes that issue, but has the downside that it breaks rendering of the READMEs on Github.

> I think the hub template only has admonitions and no links defined (since it has no headers and the headers are only defined in the readme).

Yes it would be limited to one anchor per admonition instead of per header.

But I actually think there's a better solution. We can use HTML to create invisible anchors which work both on Github and the docs site. We already use this [in our README](https://github.com/ml6team/fondant/blob/main/README.md?plain=1#L1).

PhilippeMoussalli:
> The issue is that when linking to components in the hub, we currently need to use a dynamic link ([example](https://github.com/ml6team/fondant/blob/main/docs/guides/build_a_simple_pipeline.md#:~:text=use%20the%20reusable-,load_from_hf_hub,-component.)) which breaks when adding new components.

ok now it's more clear, thanks! 

> We can use HTML to create invisible anchors which work both on Github and the docs site. We already use this [in our README](https://github.com/ml6team/fondant/blob/main/README.md?plain=1#L1).

That indeed seems to be the best solution, I added it to both the admonition header and the subheaders of the readme. I think the latter will be the most used since it opens the admonition of the selected component to the target header (description, args, ...).

RobbeSneyders (commented):
Thanks @PhilippeMoussalli! I just think you missed one dynamic link [here](https://github.com/ml6team/fondant/blob/main/docs/guides/implement_custom_components.md?plain=1#L248).

RobbeSneyders (approved):

PR #759, opened 2024-01-08
Fix ragas component

PhilippeMoussalli:
Fix to issue that @mrchtr raised

* `model_name` was actually referring to the class name used to import the model which can be `OpenAI` or `ChatOpenAI`. This caused some issues because we were trying to set `model_name` to `gpt-4`. Changed the arguments and default to reflect that. 

* ChatOpenAI seems to be the new default of defining models, changed the default model class to that one 

When initializing GPT-4 with `OpenAI()`:

```
UserWarning: You are trying to use a chat model. This way of initializing it is no longer supported. Instead, please use: `from langchain_community.chat_models import ChatOpenAI`
```

Also how it's set in RAGAS [LINK](https://docs.ragas.io/en/latest/howtos/customisations/llms.html#:~:text=Copy%20code-,from%20langchain.chat_models%20import%20ChatOpenAI,-gpt4%20%3D%20ChatOpenAI)

Changes will be required to RAG pipeline after new release

### DIFF
--- a/components/evaluate_ragas/fondant_component.yaml
+++ b/components/evaluate_ragas/fondant_component.yaml
@@ -19,13 +19,15 @@ produces:
 
 
 args:
-  module:
+  llm_module_name:
     description: Module from which the LLM is imported. Defaults to langchain.llms
     type: str
-    default: "langchain.llms"
-  llm_name:
+    default: "langchain.chat_models"
+  llm_class_name:
     description: Name of the selected llm
     type: str
+    default: "ChatOpenAI"
   llm_kwargs:
     description: Arguments of the selected llm
     type: dict
+    default: {"model_name":"gpt-4"}
--- a/components/evaluate_ragas/fondant_component.yaml
+++ b/components/evaluate_ragas/fondant_component.yaml
@@ -30,4 +30,4 @@ args:
   llm_kwargs:
     description: Arguments of the selected llm
     type: dict
-    default: {"model_name":"gpt-4"}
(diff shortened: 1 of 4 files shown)


components/evaluate_ragas/fondant_component.yaml
```diff
@@ -19,13 +19,15 @@ produces:
 
 
 args:
-  module:
+  llm_module_name:
     description: Module from which the LLM is imported. Defaults to langchain.llms
     type: str
-    default: "langchain.llms"
-  llm_name:
+    default: "langchain.chat_models"
```
mrchtr, line 25: By doing this we are deciding to use gpt-4 as default. Do you know why we have used gpt-3.5 as default before?

mrchtr (commented):
Thanks @PhilippeMoussalli! Did you already test it with the RAG evaluation pipeline?

components/evaluate_ragas/fondant_component.yaml
RobbeSneyders: gpt-4 is more expensive. I would use 3.5 by default if possible.

PhilippeMoussalli: changed to 3.5, yes but I only tested the new module with 3.5/4 on the old component. It should be fine since the only thing that's changed is the naming in this PR and the defaults. Will include the new parameters when we release

RobbeSneyders:
@mrchtr I'll let you review and approve this one since you have more context.

Issue #751, opened 2024-01-02
Infer the arguments based on component `__init__` arguments


RobbeSneyders:
When the user uses Lightweight Python components (#558) we want to get any information we currently get from the component spec from the provided Python code.

For the arguments, we should be able to get them from the `__init__` arguments of the component class. These should be typed to provide the proper information, but we can fall back on string type for any untyped arguments.

PR #761, opened 2024-01-08
Accept actual type instead of string representation in Argument class

RobbeSneyders:
Needed some refactoring while working on #751.

Submitting this as a separate PR for easy reviewing.

### DIFF
--- a/src/fondant/core/component_spec.py
+++ b/src/fondant/core/component_spec.py
@@ -2,6 +2,7 @@
 import copy
 import json
 import pkgutil
+import pydoc
 import re
 import types
 import typing as t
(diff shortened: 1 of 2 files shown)


src/fondant/core/component_spec.py
```diff
@@ -208,7 +197,7 @@ def args(self) -> t.Mapping[str, Argument]:
                 name: Argument(
                     name=name,
                     description=arg_info["description"],
-                    type=arg_info["type"],
+                    type=pydoc.locate(arg_info["type"]),  # type: ignore
```
GeorgesLorre, line 200: Is it normal I find no docs on pydoc.locate ?

GeorgesLorre (approved):

GeorgesLorre:
No tests ? 🤔

src/fondant/core/component_spec.py
RobbeSneyders: Doesn't seem well documented indeed. Got it from [here](https://stackoverflow.com/a/29831586/4098821).

RobbeSneyders:
> No tests ? 🤔

I didn't change any behavior. This code is covered indirectly.

Issue #750, opened 2024-01-02
Support applying Lightweight Python components in Pipeline SDK


RobbeSneyders:
We want to be able to apply Lightweight Python components as part of a pipeline just like we do with docker components.

```python
from fondant.component import PandasTransformComponent
from fondant.pipeline import Pipeline

class MyComponent(PandasTransformComponent):

    def __init__(...):
        ...

    def transform(dataframe):
        ...

pipeline = Pipeline(...)

dataset = pipeline.read(...)

dataset.apply(
    MyComponent,
    consumes={},
    produces={},
    arguments={},
)
```

Lightweight Python components will require some additional arguments compared to docker components. I can think of two already:
- `image`: docker image to run the code in.
- `dependencies`: additional Python dependencies to install in the container before executing.

I see two options:
- We extend the `apply` (and `read` and `write`) methods on the dataset / pipeline to support these arguments. Since they are not relevant for docker components, we might then want to split this into two separate `apply` methods for clarity.
- We add a decorator on the Lightweight Python component which defines these options:
  ```python
  from fondant.pipeline import LightWeightComponent

  @LightWeightComponent(image=..., dependencies=[...])
  class MyComponent(PandasTransformComponent):
      ...
  ```

I think I would prefer the second option so we can keep a single `apply` interface.

GeorgesLorre:
I also think a decorator will be the cleanest and clearest way of achieving this. This means that an `.apply()` can take a string (reusable component), path (local custom component) or a decorated class.  

We also need to maybe think about the eager execution interface:

This is what I have now (calling `.execute()`)
```
pipeline = Pipeline(name="foo", description="bar", base_path="/foobar")

class LoadFromParquett(DaskLoadComponent):
    def __init__(self, *_, **__):
        pass
    def load(self) -> dd.DataFrame:
        dask_df = dd.read_parquet("./foobar/sample.parquet", columns=["x", "y"])
        return dask_df

dataset1 = pipeline.execute(
    component=LoadFromParquett,
    produces={"x": pa.int32(), "y": pa.int32()},
)

```

but it might be nicer to keep `.apply()` and calling `.execute()` on top of it

```
dataset1 = pipeline.apply(
    LoadFromParquett,
    produces={"x": pa.int32(), "y": pa.int32()},
)

dataset1.execute(override_df=a_df)
```

RobbeSneyders:
[...]

PR #763, opened 2024-01-08
Add component argument inference

### LINKED ISSUE #751: Infer the arguments based on component `__init__` arguments

RobbeSneyders:
Fixes #751 

This PR introduces functionality to infer the arguments from a `Component` class. The result is a dictionary with the argument names as keys, and `Argument` instances as values, which is the format of [`component_spec.args`.](https://github.com/ml6team/fondant/blob/8e828441eec8ff91074e5c8ccf16fe405b719594/src/fondant/core/component_spec.py#L193)

We can leverage this behavior for Lightweight Python components as described in #750.

Did some TDD here, let me know if I missed any cases.

### DIFF
--- a/src/fondant/pipeline/argument_inference.py
+++ b/src/fondant/pipeline/argument_inference.py
@@ -1,0 +1,103 @@
+import inspect
+import typing as t
+
+from fondant.component import Component
+from fondant.core.component_spec import Argument
+from fondant.core.exceptions import UnsupportedTypeAnnotation
+
+BUILTIN_TYPES = [str, int, float, bool, dict, list]
+
+
+def annotation_to_type(annotation: t.Any) -> t.Type:
+    """Extract the simple built-in type from an Annotation.
+
+    Examples:
+        dict[str, int] -> dict
+        t.Optional[str] -> str
+
+    Args:
+        annotation: Annotation of an argument as returned by inspect.signature
+
+    Raises:
+        UnsupportedTypeAnnotation: If the annotation is not simple or not based on a built-in type
+
+    """
+    # If no type annotation is present, default to str
+    if annotation == inspect.Parameter.empty:
+        return str
+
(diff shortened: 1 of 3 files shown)


RobbeSneyders:
Pipeline failure is unrelated. Submitted #764 to fix it.

tests/pipeline/test_argument_inference.py
```diff
@@ -0,0 +1,247 @@
...
+        "list_argument": Argument(
+            name="list_argument",
+            type=list,
+            optional=True,
+            default=None,
+        ),
+    }
+
+
+def test_parametrized_types_old():
```
PhilippeMoussalli, line 144: what's the "old" referring to? can we keep it to `test_parametrized_types` to match the above?

PhilippeMoussalli (approved):
Thanks Robbe! Nice work, the tests are quite extensive and seem to cover all scenarios

src/fondant/pipeline/argument_inference.py
```diff
@@ -0,0 +1,103 @@
...
+import typing as t
+
+from fondant.component import Component
+from fondant.core.component_spec import Argument
+from fondant.core.exceptions import UnsupportedTypeAnnotation
+
+BUILTIN_TYPES = [str, int, float, bool, dict, list]
+
+
+def annotation_to_type(annotation: t.Any) -> t.Type:
```
GeorgesLorre, line 11: Nice !

tests/pipeline/test_argument_inference.py
GeorgesLorre, line 1: 😍

GeorgesLorre (approved):

PR #764, opened 2024-01-08
Fix readme generation

RobbeSneyders:
Fixes readme generation which was broken by #761.
Due to the setup of the pre-commit hook, this was only detected after merging to main.


PhilippeMoussalli (approved):
Thanks Robbe!

PR #765, opened 2024-01-09
Add load from pdf component

PhilippeMoussalli:
Fixes https://github.com/ml6team/fondant-use-cases/issues/54

PR that adds the functionality to load pdf documents from different local and remote storage. 

The implementation differs from the suggested solution at [#54](https://github.com/ml6team/fondant-use-cases/issues/54) since:
* Accumulating different loaders and loading each document individually seems to be inefficient since it would require the initialization of a client, temp storage, ... on every invocation [link](https://github.com/langchain-ai/langchain/blob/04caf07dee2e2843ab720e5b8f0c0e83d0b86a3e/libs/community/langchain_community/document_loaders/gcs_file.py#L62)
* The langchain cloud loaders don't have a unified interface
     *  Each would requires specific arguments to be passed (in contrast fsspec is much simpler)
     *  Only the google loader enables defining a custom loader class, the rest uses the `Unstructured` loader which requires a lot of system and cuda dependencies to have it installed (a lot of overhead for just loading pdfs)

The current implementation relies on copying the pdfs to a temporary local storage and loading them using the `PyPDFDirectoryLoader`, they are then loaded lazily. The assumption for now is that the loaded docs won't exceed the storage of the device which should be valid for most use cases. Later on, we can think on how to optimize this further.


RobbeSneyders (commented):
Thanks @PhilippeMoussalli!

- Can you move the `test_file` and `test_folder` into the `tests` directory?
- As discussed, this currently doesn't work with larger-than-memory data. For that, we probably first need to fetch all the paths, partition them, and apply a transform that loads them.

PhilippeMoussalli:
> Thanks @PhilippeMoussalli!
> 
> * Can you move the `test_file` and `test_folder` into the `tests` directory?
> * As discussed, this currently doesn't work with larger-than-memory data. For that, we probably first need to fetch all the paths, partition them, and apply a transform that loads them.

Thanks for the suggestions, made the necessary changes

components/load_from_pdf/src/main.py
```diff
@@ -73,47 +70,53 @@ def _get_meta_df() -> pd.DataFrame:
...
+        file_paths = [
+            file_path for file_path in file_paths if file_path.endswith(".pdf")
+        ]
+
+        if self.n_rows_to_load is not None:
+            file_paths = file_paths[: self.n_rows_to_load]
+
+        dask_df = dd.from_pandas(
+            pd.DataFrame({"pdf_path": file_paths}),
+            npartitions=os.cpu_count(),
```
RobbeSneyders, line 106: This could probably be a parameter so it can be upped for larger datasets.

RobbeSneyders (approved):
Thanks @PhilippeMoussalli. One optional comment. Up to you if you think it's worth the effort.

components/load_from_pdf/src/main.py
PhilippeMoussalli: Added :)

PR #767, opened 2024-01-10
add teardown method

PhilippeMoussalli:
PR that adds a teardown method to every component. Useful for shutting down database connections and clients instead of it happening abruptly after the container is shutdown. 

Related to https://github.com/ml6team/fondant-use-cases/issues/59

### DIFF
--- a/src/fondant/component/component.py
+++ b/src/fondant/component/component.py
@@ -26,6 +26,11 @@ class BaseComponent:
     ):
         pass
 
+    def teardown(self) -> None:
+        """Method called after the component has been executed."""
+
+    pass
+
 
 class DaskLoadComponent(BaseComponent):
     """Component that loads data and returns a Dask DataFrame."""
--- a/tests/component/test_component.py
+++ b/tests/component/test_component.py
@@ -274,10 +274,19 @@ def test_load_component(metadata):
         "{}",
     ]
 
+    class MockClient:
+        def __init__(self):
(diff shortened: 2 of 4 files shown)


src/fondant/component/component.py
```diff
@@ -26,6 +26,11 @@ def __init__(
     ):
         pass
 
+    def teardown(self) -> None:
+        """Method called after the component has been executed."""
+
+    pass
```
mrchtr, line 32:
```suggestion
        pass
```

tests/component/test_component.py
```diff
@@ -288,14 +297,24 @@ def load(self):
             }
             return dd.DataFrame.from_dict(data, npartitions=N_PARTITIONS)
 
+        def teardown(self) -> None:
+            self.client.shutdown()
+            assert self.client.is_connected is False
```
mrchtr, line 302: Nitpick: I would call the assert outside of the MyLoadComponent.

mrchtr (approved):
Thanks @PhilippeMoussalli! Two small nitpicks.

tests/component/test_component.py
RobbeSneyders, line 1: I would propose to create a separate test function to test the teardown method instead of expanding a current one. Ideally each test tries to test a single thing in isolation as much as possible.

GeorgesLorre (approved):
Nice feature !

PR #769, opened 2024-01-10
Handle docker compose errors

PhilippeMoussalli:
PR that handles errors when docker compose faces an issue, we now a raise a valid python error with the error command that is being thrown. This make it more clear for the user that something went off with the pipeline. 

@janvanlooy


tests/pipeline/test_runner.py
```diff
@@ -58,6 +73,8 @@ def test_docker_runner_from_pipeline():
                 "--remove-orphans",
             ],
             env=dict(os.environ, DOCKER_DEFAULT_PLATFORM="linux/amd64"),
+            capture_output=True,
+            encoding="utf8",
         )
```
GeorgesLorre, line 80: A test that checks the ` PipelineRunError` error might be nice here, we do this for other parts of the codes aswel:
https://github.com/ml6team/fondant/blob/main/tests/test_explorer.py#L164

src/fondant/pipeline/runner.py
```diff
@@ -43,10 +44,17 @@ def _run(self, input_spec: str, *args, **kwargs):
...
-        subprocess.call(  # nosec
+        output = subprocess.run(  # nosec
             cmd,
             env=dict(os.environ, DOCKER_DEFAULT_PLATFORM="linux/amd64"),
+            capture_output=True,
+            encoding="utf8",
         )
+
+        if output.returncode != 0:
+            msg = f"Command failed with error: '{output.stderr}'"
```
GeorgesLorre, line 55: Does this properly show a full traceback with line numbers etc ?

GeorgesLorre (commented):
Nice!

src/fondant/pipeline/runner.py
```diff
@@ -43,10 +44,17 @@ def _run(self, input_spec: str, *args, **kwargs):
         print("Starting pipeline run...")
 
         # copy the current environment with the DOCKER_DEFAULT_PLATFORM argument
-        subprocess.call(  # nosec
+        output = subprocess.run(  # nosec
```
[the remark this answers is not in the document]
RobbeSneyders: Sounds good to me. `run` indeed seems to be recommended now. If we still need the `check` behavior, it supports a `check=true` argument.

src/fondant/pipeline/runner.py
PhilippeMoussalli: Just tested this and it does not seem to work well when working with docker-compose:

* using subprocess to call a python script with an error results in return_code 1 and proper error handling 
* using subprocess to call a docker compose that calls docker compose that launches a python script with an error results in return_code 0. However, `output.stderr` still has some content in it that is a proper traceback to the the error 

I'v added an example for testing 
https://github.com/ml6team/fondant/blob/handle-docker-compose-errors/example_subprocess/launch_subprocess.py

We could alternatively check if `stderr` is an empty string but not sure if that's the best method since I am not sure if warnings or other content gets streamed to it. 

Any suggestions?

PhilippeMoussalli: Fixed now by adding `--abort-on-container-exit` flag to docker compose

tests/pipeline/test_runner.py
PhilippeMoussalli: Added tests

src/fondant/pipeline/runner.py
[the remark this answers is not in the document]
mrchtr: I think the `check=true` we don't need if we raise the exception anyway?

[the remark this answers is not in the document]
PhilippeMoussalli: Currently we don't have `check=true` actually 
I tested with and the error is not specific when you set it to true 

without (current state)
```bash
Traceback (most recent call last):
  File "/home/philippe/Scripts/express/a/launch.py", line 13, in <module>
    raise Exception(msg)
Exception: Command failed with error: ' Container a-web-1  Created
a-web-1  | Traceback (most recent call last):
a-web-1  |   File "app.py", line 4, in <module>
a-web-1  |     print(1 / 0)
a-web-1  | ZeroDivisionError: division by zero
Aborting on container exit...
 Container a-web-1  Stopping
 Container a-web-1  Stopped
```

With 
```bash
/home/philippe/Scripts/express/venv/bin/python /home/philippe/Scripts/express/a/launch.py 
Traceback (most recent call last):
  File "/home/philippe/Scripts/express/a/launch.py", line 4, in <module>
    output = subprocess.run(  # nosec
  File "/usr/lib/python3.10/subprocess.py", line 526, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['docker', 'compose', 'up', '--abort-on-container-exit']' returned non-zero exit status 1.
```

mrchtr (approved):
Thanks @PhilippeMoussalli!

PR #770, opened 2024-01-10
Support applying Lightweight Python components in Pipeline SDK

GeorgesLorre:


tests/component/test_python_component.py
```diff
@@ -0,0 +1,32 @@
...
+
+def test_python_component():
+    pipeline = Pipeline(name="dummy-pipeline", base_path="./data")
+
+    @python_component(
+        base_image="python:3.8-slim-buster",
+        extra_requires=["pandas", "dask"],
+    )
+    class CreateData(DaskLoadComponent):
+        def __init__(self, *_, **__):
```
GeorgesLorre, line 16: This is needed to override the init of the `BaseComponent` but ideally it is not required in the future.

src/fondant/pipeline/pipeline.py
```diff
@@ -362,16 +376,41 @@ def read(
...
-            name_or_path,
-            produces=produces,
-            arguments=arguments,
-            input_partition_rows=input_partition_rows,
-            resources=resources,
-            cache=cache,
-            cluster_type=cluster_type,
-            client_kwargs=client_kwargs,
-        )
+        if isinstance(ref, (str, Path)):
```
GeorgesLorre, line 379: Ideally I can flip this and check if the ref is a component.

src/fondant/pipeline/pipeline.py
```diff
@@ -362,16 +376,41 @@ def read(
...
+            operation = ComponentOp.from_component_yaml(ref)
+        else:
+            if not ref.is_python_component():
+                err = """Only Python components are currently supported.
+                Make sure your class is decorated with the @python_component decorator."""
+                raise ValueError(err)
+            name = ref().__class__.__name__
```
GeorgesLorre, line 386: need to instantiate the class to get the name

src/fondant/pipeline/pipeline.py
```diff
@@ -362,16 +376,41 @@ def read(
...
+                "description": "This is an example component",
+                "image": "example_component:latest",
+                "produces": {
+                    "additionalProperties": True,
+                },
+                "consumes": {
+                    "additionalProperties": True,
+                },
+            }
+            component_spec = ComponentSpec(spec_dict)
```
GeorgesLorre, line 400: This could possibly be refactored by creating and operation_spec directly

src/fondant/pipeline/pipeline.py
```diff
@@ -181,6 +179,22 @@ def __init__(
 
         self.resources = resources or Resources()
 
+    @classmethod
+    def from_component_yaml(cls, path, **kwargs):
```
GeorgesLorre, line 183: this will be the entrypoint for reusable and custom components (the ones that the python_component does not cover)

src/fondant/component/image.py
```diff
@@ -0,0 +1,9 @@
+import typing as t
+from dataclasses import dataclass
+
+
+@dataclass
+class Image:
```
GeorgesLorre, line 6: This needs way more functionality to actually make the images with extra dependencies and the script available

src/fondant/component/component.py
```diff
@@ -73,3 +75,30 @@ def transform(self, dataframe: pd.DataFrame) -> pd.DataFrame:
...
+            def image(cls) -> Image:
+                return image
+
+            @classmethod
+            def is_python_component(cls) -> bool:
+                return True
+
+        return PythonComponent
+
+    return wrapper
```
GeorgesLorre, line 104: It works but I'm not completely convinced this is the way forward. Essentially we subclass the Component and add an image property with the details provided in the decorator arguments.

GeorgesLorre:
I can recommend this talk on python decorators: https://www.youtube.com/watch?v=MjHpMCIvwsY

src/fondant/pipeline/pipeline.py
```diff
@@ -142,16 +146,10 @@ def __init__(
...
         resources: t.Optional[Resources] = None,
     ) -> None:
-        if self._is_custom_component(name_or_path):
-            self.component_dir = Path(name_or_path)
-        else:
-            self.component_dir = self._get_registry_path(str(name_or_path))
-
+        self.name = name
+        self.image = image
+        self.component_spec = component_spec
```
GeorgesLorre, line 151: I need to check further but maybe we can do without the component_spec here and use the operation_spec along with the image from this point on.

tests/component/test_python_component.py
```diff
@@ -0,0 +1,32 @@
...
+import pyarrow as pa
+from fondant.component import DaskLoadComponent, python_component
+from fondant.pipeline import Pipeline
+
+
+def test_python_component():
+    pipeline = Pipeline(name="dummy-pipeline", base_path="./data")
+
+    @python_component(
+        base_image="python:3.8-slim-buster",
```
RobbeSneyders, line 12: Here we'll need to provide a base docker image with `Fondant` installed, right?

src/fondant/pipeline/pipeline.py
RobbeSneyders: You can with the current implementation, right? Since it's a subclass of a component subclass?

RobbeSneyders (commented):
Thanks @GeorgesLorre, starting to look good!

src/fondant/component/component.py
```diff
@@ -73,3 +75,30 @@ def transform(self, dataframe: pd.DataFrame) -> pd.DataFrame:
...
+):
+    """Decorator to enable a python component."""
+
+    def wrapper(cls):
+        image = Image(
+            base_image=base_image,
+            extra_requires=extra_requires,
+        )
+
+        @wraps(cls, updated=())
```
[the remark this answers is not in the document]
GeorgesLorre: `@wraps` doesn't play well with classes:

https://stackoverflow.com/questions/6394511/python-functools-wraps-equivalent-for-classes

tests/component/test_python_component.py
GeorgesLorre: yes we need a sane default which we ideally configure in 1 place --> here

src/fondant/pipeline/pipeline.py
RobbeSneyders: Isn't `ref == ref().__class__`?
I think `ref.__name__` should work as well.

RobbeSneyders: Or is this because of the decorator?

src/fondant/component/component.py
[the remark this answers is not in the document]
RobbeSneyders: Can you add a comment? :)

tests/component/test_python_component.py
RobbeSneyders, line 1: This is testing the pipeline SDK, so should probably move to `tests/pipeline`

RobbeSneyders: Actually, this is the case for the `python_component` and `Image` as well. I think these should be part of the pipeline SDK.

src/fondant/component/image.py
GeorgesLorre: Yes but I guess we'll have shared functionality between runners that could be abstracted here

src/fondant/pipeline/pipeline.py
GeorgesLorre: Resolved by instantiating (this might be more difficult later because of args)

tests/pipeline/test_python_component.py
```diff
@@ -0,0 +1,32 @@
...
+
+        def load(self) -> dd.DataFrame:
+            df = pd.DataFrame(
+                {
+                    "x": [1, 2, 3],
+                    "y": [4, 5, 6],
+                },
+                index=pd.Index(["a", "b", "c"], name="id"),
+            )
+            return dd.from_pandas(df, npartitions=1)
```
mrchtr, line 27: I would try to add a validation for the return type as well. If a user don't returns a dask or pandas dataframe in the custom components, it will not work. It's not related to your code, we have the same requirement already for the custom components. But a here we are lowering the entry barrier and I can imagine that people will not read the documentation in detail anymore. 

So I think it might be good if we could catch somehow if it is a wrong return type. I guess it should be possible within the decorator implementation as well.

mrchtr (commented):
I like it! Looks really promising and I think adds it is great for the usability and quick experiments. 

The decorator give use some additional power that we should use imo. I would try to add useful validation and hints to guarantee a low entry barrier for users.
 
We can probably tackle this in follow PRs :)

mrchtr:
> The decorator give use some additional power that we should use imo. I would try to add useful validation and hints to guarantee a low entry barrier for users.

To follow up on this - I've played a bit around with the decorator and validations. Maybe it helps.

```python
from typing import TypeVar, Callable

import pandas as pd

T = TypeVar('T')
REQUIRED_METHODS = {"load": [pd.DataFrame]}

def LightWeightComponent(cls):
    class Wrapped(cls):
        def __init__(self, *args, **kwargs):
            for method, types in REQUIRED_METHODS.items():
                attr = self._valid_method_exists(method)
                for _type in types:
                    self._validate_return_type(_type, attr)

        def _valid_method_exists(self, name):
            if not hasattr(self, name):
                raise AttributeError(
                    f"{self.__class__.__name__} Function is missing: {name}")
            else:
                return self.__getattribute__(name)

        def _validate_return_type(self, expected_type: type, func: Callable[..., T]):
            return_type = func.__annotations__["return"]
            if not return_type == expected_type:
                raise AttributeError(
                    f"{return_type} is wrong return type in: {func}. Expected {expected_type}")


    return Wrapped

@LightWeightComponent
class BaseComponent:
    def load(self) -> str:
        pass

@LightWeightComponent
class SubClass(BaseComponent):
    def load(self) -> pd.DataFrame:
        pass

@LightWeightComponent
class SecondSubClass(BaseComponent):
    def load(self) -> str: # Through error during initialisation
        return "0"

if __name__ == '__main__':
    SubClass()
    SecondSubClass()
```

src/fondant/pipeline/lightweight_component.py
```diff
@@ -0,0 +1,38 @@
...
+def lightweight_component(
+    extra_requires: t.Optional[t.List[str]] = None,
+    base_image: t.Optional[str] = None,
+):
+    """Decorator to enable a python component."""
+
+    def wrapper(cls):
+        image = Image(
+            base_image=base_image,
+            extra_requires=extra_requires,
```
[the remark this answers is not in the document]
RobbeSneyders: I added it on my branch:
https://github.com/ml6team/fondant/compare/feature/python-component...feature/python-in-docker

src/fondant/pipeline/lightweight_component.py
```diff
@@ -0,0 +1,38 @@
+import typing as t
+from dataclasses import dataclass
+from functools import wraps
+
+
+@dataclass
+class Image:
+    base_image: t.Optional[str] = "fondant:latest"
+    extra_requires: t.Optional[t.List[str]] = None
+    script: t.Optional[str] = None
```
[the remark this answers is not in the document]
RobbeSneyders: For every runner, since the script is not baked into the image. In the docker & KfP runner, we'll probably include the script in the entrypoint instead of uploading it separately though.

RobbeSneyders:
+1 on the validation, would really help to give immediate feedback instead of only at runtime. Would keep it out of this PR though, so we can merge it and work in parallel on the remaining work.

src/fondant/core/component_spec.py
```diff
@@ -109,6 +109,48 @@ def to_file(self, path) -> None:
         with open(path, "w", encoding="utf-8") as file_:
             yaml.dump(self._specification, file_)
 
+    @classmethod
+    def from_dict(cls, component_spec_dict: t.Dict[str, t.Any]) -> "ComponentSpec":
+        """Load the component spec from a dictionary."""
+        return cls(component_spec_dict)
+
+    @classmethod
+    def from_args(
```
RobbeSneyders, line 118: I'm still in favor of making this the `__init__` method, and storing the different parts of the spec as different attributes instead of converting to a spec.

mrchtr:
> Would keep it out of this PR though, so we can merge it and work in parallel on the remaining work.

Yes indeed. I guess we could even collect some more validations. This were just the first edge cases which came into my mind.

src/fondant/core/component_spec.py
```diff
@@ -61,11 +61,54 @@ class ComponentSpec:
...
+            "image": image,
+        }
+
+        if description:
+            spec_dict["description"] = description
+
+        if consumes:
+            spec_dict["consumes"] = consumes
+        else:
+            spec_dict["consumes"] = {"additionalProperties": True}
```
RobbeSneyders, line 101: I think this logic is specific to the lightweight python components, and doesn't hold in the general case.

I would say that our specification is a [closed-world contract](https://en.wikipedia.org/wiki/Closed-world_assumption) (in contrast to eg. OpenAPI, which is an open-world contract). So if `consumes` is empty, `additionalProperties` should be `False` (the default).

For lightweight python components, `consumes` and `produces` should be `{"additionalProperties": True}` by default, since the user is expected to override them when applying the component.

So I would have the lightweight python components pass in `{"additionalProperties": True}` and don't set any default here.

src/fondant/pipeline/pipeline.py
```diff
@@ -362,16 +379,38 @@ def read(
...
-            input_partition_rows=input_partition_rows,
-            resources=resources,
-            cache=cache,
-            cluster_type=cluster_type,
-            client_kwargs=client_kwargs,
-        )
+        if issubclass(ref, BaseComponent):
+            if not ref.is_python_component():
+                err = """Only Python components are currently supported.
+                Make sure your class is decorated with the @python_component decorator."""
```
RobbeSneyders, line 385: This error message may be confusing, since we do accept non-Python components with this method, just not by direct reference.

RobbeSneyders (commented):
Thanks @GeorgesLorre.

Only some small comments remaining. Looks good overall, looking forward to merge this once the tests are fixed :rocket:

GeorgesLorre:
> > Would keep it out of this PR though, so we can merge it and work in parallel on the remaining work.
> 
> Yes indeed. I guess we could even collect some more validations. This were just the first edge cases which came into my mind.

Thx for the great ideas @mrchtr Will take this up in subsequent PR's!

src/fondant/pipeline/lightweight_component.py
```diff
@@ -0,0 +1,42 @@
...
+@dataclass
+class Image:
+    base_image: str = "fondant:latest"
+    extra_requires: t.Optional[t.List[str]] = None
+    script: t.Optional[str] = None
+
+
+class PythonComponent:
+    @classmethod
+    def image(cls) -> Image:
```
RobbeSneyders, line 15: Any reason for this not to be a (class)property?

src/fondant/pipeline/pipeline.py
```diff
@@ -362,16 +387,43 @@ def read(
...
-            client_kwargs=client_kwargs,
-        )
+        if inspect.isclass(ref) and issubclass(ref, PythonComponent):
+            name = ref.__name__
+            image = ref.image()
+            description = ref.__doc__ or "python component"
+
+            component_spec = ComponentSpec(
+                name,
+                image.base_image,  # TODO: revisit
```
RobbeSneyders, line 397: What needs to be revisited about this?

src/fondant/pipeline/pipeline.py
```diff
@@ -362,16 +387,43 @@ def read(
...
+            operation = ComponentOp.from_component_yaml(
+                ref,
+                produces=produces,
+                arguments=arguments,
+                input_partition_rows=input_partition_rows,
+                resources=resources,
+                cache=cache,
+                cluster_type=cluster_type,
+                client_kwargs=client_kwargs,
+            )
```
RobbeSneyders, line 426: This code is almost identically repeated three times. Would be good if we can abstract this away.

src/fondant/pipeline/lightweight_component.py
```diff
@@ -0,0 +1,42 @@
+import typing as t
+from dataclasses import dataclass
+from functools import wraps
+
+
+@dataclass
+class Image:
+    base_image: str = "fondant:latest"
```
RobbeSneyders, line 8: We should link this to the installed version of Fondant once we add the CI/CD for this.

src/fondant/pipeline/pipeline.py
```diff
@@ -213,10 +235,13 @@ def _configure_caching_from_image_tag(
...
 
-    @property
-    def dockerfile_path(self) -> t.Optional[Path]:
-        path = self.component_dir / "Dockerfile"
-        return path if path.exists() else None
+    def dockerfile_path(self, path: t.Union[str, Path]) -> t.Optional[Path]:
+        if self._is_custom_component(path):
+            component_dir = Path(path)
+        else:
+            component_dir = self._get_registry_path(str(path))
```
RobbeSneyders, line 242: I guess this change can be reverted? Although I would sill prefer it we can find a way to remove the `component_dir` argument from the `__init__` method.

RobbeSneyders (approved):
Thanks @GeorgesLorre!

Some minor comments, but let's tackle those in a follow-up PR.

PR #774, opened 2024-01-11
Bugfix subprocess error handling

PhilippeMoussalli:
Setting `capture_output=True` captures both the default output`stdout` and the error output `stderr` in the output variable which means that we're not streaming anymore the standard output of docker-compose up. 

This PR fixes this by only capturing the error `stderr` in the output, the standard output is streamed 

Before: 

```bash 
Traceback (most recent call last):
  File "/home/philippe/Scripts/express/a/launch.py", line 14, in <module>
    raise Exception(msg)
Exception: Command failed with error: ' Container a-web-1  Created
a-web-1  | Traceback (most recent call last):
a-web-1  |   File "app.py", line 4, in <module>
a-web-1  |     print(1 / 0)
a-web-1  | ZeroDivisionError: division by zero
Aborting on container exit...
 Container a-web-1  Stopping
 Container a-web-1  Stopped
```

After: 

```bash
Attaching to a-web-1
a-web-1  | This is a simple Python script.  #Standard output of print statement contained in the script 
a-web-1 exited with code 1
Traceback (most recent call last):
  File "/home/philippe/Scripts/express/a/launch.py", line 14, in <module>
    raise Exception(msg)
Exception: Command failed with error: ' Container a-web-1  Created
a-web-1  | Traceback (most recent call last):
a-web-1  |   File "app.py", line 4, in <module>
a-web-1  |     print(1 / 0)
a-web-1  | ZeroDivisionError: division by zero
Aborting on container exit...
 Container a-web-1  Stopping
 Container a-web-1  Stopped
```


RobbeSneyders:
Is it an issue that `stderr` is no longer streamed? I think there's ways around it, but they require using `Popen` directly I believe.

PhilippeMoussalli:
> Is it an issue that `stderr` is no longer streamed? I think there's ways around it, but they require using `Popen` directly I believe.

Not streaming  `stderr` it's fine because we will raise it anyways when we catch the error. I think the issue is mainly with streaming `stdout` since we don't have that logs that are appearing anymore. Setting `capture_output=True` seem to block streaming for both output types which is why now we only capture `stderr`

Screenshot from 2024-01-11 16-48-35

PhilippeMoussalli:
Closed, the favor of Handle local pipeline error propagation  #775

PhilippeMoussalli:
Closed in favor of #775

PR #777, opened 2024-01-15
Add RAG blogpost announcement

mrchtr:


docs/blog/posts/2024-15-01|RAG_finetuning_with_Fondant.md
janvanlooyml6, line 1: ... example use case to showcase... sounds a bit weird with twice the 'case'. maybe better just example.

janvanlooyml6 (approved):
Nice work!

PR #778, opened 2024-01-15
Fix load from pdf component

PhilippeMoussalli:
PR that fixes the load from PDF component. Previously the spec was passed as an argument to the component but with the new changes we only need to pass in the `produces` section of the spec


mrchtr (commented):
Actually the component test is failing. The order in the list comparison isn't the same. 
We can do the following instead:

```python
assert set(output_dataframe["file_name"].tolist()) == {"dummy_2.pdf", "dummy_1.pdf"}
            assert set(output_dataframe["text"].tolist()) == {"Dummy PDF file\n",
                                                              "Dummy PDF file\n"}
```

PR #779, opened 2024-01-15
Move cloud auth to sdk

PhilippeMoussalli:
The arguments that enabled authenticating to different clouds through volume mounting credentials were restricted to the CLI and were not explicitly available in the SDK.

This PR that moves the argument for authenticating to one of the clouds in the DockerRunner for the CLI to the SDK 

e.g 

authenticating to gcp 

```python
runner.run(auth_gcp=True)
```

### DIFF
--- a/src/fondant/cli.py
+++ b/src/fondant/cli.py
@@ -37,17 +37,6 @@ if t.TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 
-def get_cloud_credentials(args) -> t.Optional[str]:
-    if args.auth_gcp:
-        return CloudCredentialsMount.GCP.value
-    if args.auth_aws:
-        return CloudCredentialsMount.AWS.value
-    if args.auth_azure:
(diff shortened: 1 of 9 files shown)


src/fondant/cli.py
```diff
@@ -218,6 +202,9 @@ def start_explore(args):
         tag=args.tag,
         port=args.port,
         extra_volumes=extra_volumes,
+        auth_gcp=args.auth_gcp,
```
mrchtr, line 205: What do you think about combining these arguments into a single one, for instance `auth_provider=AZURE,AWS,GCP`? Might be easier to update the different auth options without changing the interface.

mrchtr (commented):
Thanks @PhilippeMoussalli!

src/fondant/cli.py
PhilippeMoussalli: good suggestion! Implemented

mrchtr (approved):

Issue #782, opened 2024-01-15
Add CI/CD to build default Fondant base image


RobbeSneyders:
Lightweight Python components need a default base image to run on, which we should provide. Probably for all Python versions we support. I would build them everytime we build our reusable images.

We should also make sure the base image with the appropriate Fondant version installed is used automatically ([comment](https://github.com/ml6team/fondant/pull/786#discussion_r1455025312)).

PR #786, opened 2024-01-16
Add support to run lightweight python components in docker runner

### LINKED ISSUE #753: Support Lightweight Python components in Docker runner
We should be able to execute Lightweight Python components with our Docker runner.

We can take inspiration from KfP which [seems to](https://github.com/kubeflow/pipelines/blob/58ce09e07d031964905020c749e77bf0f37e83d4/sdk/python/kfp/dsl/component_factory.py#L450) include the python script as part of the entrypoint of the docker container.

RobbeSneyders:
Fixes #753

### DIFF
--- a/examples/sample_pipeline/pipeline.py
+++ b/examples/sample_pipeline/pipeline.py
@@ -1,11 +1,14 @@
 # This file contains a sample pipeline. Loading data from a parquet file,
 # using the load_from_parquet component, chain a custom dummy component, and use
 # the reusable chunking component
+import os
 from pathlib import Path
 
+import pandas as pd
 import pyarrow as pa
 
-from fondant.pipeline import Pipeline
+from fondant.component import PandasTransformComponent
+from fondant.pipeline import Pipeline, lightweight_component
 
 BASE_PATH = Path("./.artifacts").resolve()
 BASE_PATH.mkdir(parents=True, exist_ok=True)
@@ -32,8 +35,28 @@ dataset = dataset.apply(
     "./components/dummy_component",
 )
 
-dataset.apply(
+dataset = dataset.apply(
     "chunk_text",
     arguments={"chunk_size": 10, "chunk_overlap": 2},
     consumes={"text": "text_data"},
+)
+
+
+@lightweight_component(
+    base_image="python:3.8",
+    extra_requires=[
+        f"fondant[component]@git+https://github.com/ml6team/fondant@"
+        f"{os.environ.get('FONDANT_VERSION', 'main')}",
+    ],
+)
+class CalculateChunkLength(PandasTransformComponent):
+    def transform(self, dataframe: pd.DataFrame) -> pd.DataFrame:
+        dataframe["chunk_length"] = dataframe["text"].apply(len)
+        return dataframe
+
+
+_ = dataset.apply(
+    ref=CalculateChunkLength,
+    consumes={"text": pa.string()},
+    produces={"chunk_length": pa.int32()},
 )
--- a/examples/sample_pipeline/run.sh
+++ b/examples/sample_pipeline/run.sh
@@ -27,6 +27,8 @@ trap cleanup EXIT
 # Bind local data directory to pipeline
 data_dir=$(readlink -f "data")
 
+export FONDANT_VERSION=$GIT_HASH
+
 # Run pipeline
 poetry run fondant run local pipeline.py \
   --extra-volumes $data_dir:/data --build-arg FONDANT_VERSION=$GIT_HASH
--- a/src/fondant/pipeline/compiler.py
+++ b/src/fondant/pipeline/compiler.py
@@ -3,6 +3,7 @@ import logging
 import os
 import shlex
 import tempfile
+import textwrap
 import typing as t
 from abc import ABC, abstractmethod
 from dataclasses import asdict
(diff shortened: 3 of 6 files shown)


tests/pipeline/test_python_component.py
```diff
@@ -44,5 +85,4 @@ def transform(self, dataframe: pd.DataFrame) -> pd.DataFrame:
         ref=AddN,
         produces={"x": pa.int32(), "y": pa.int32()},
         consumes={"x": pa.int32(), "y": pa.int32()},
-        arguments={"n": 1},
```
RobbeSneyders, line 47: We still need to integrate #763 to support arguments.

src/fondant/pipeline/compiler.py
```diff
@@ -45,6 +47,37 @@ def log_unused_configurations(self, **kwargs):
                     f" for runner `{self.__class__.__name__}`.",
                 )
 
+    @staticmethod
+    def _build_entrypoint(image: Image) -> t.List[str]:
```
RobbeSneyders, line 51: Don't really know how to test this without either replicating the logic or comparing against an unreadable string, so for now this is only tested via the integration test.

Open for ideas.

examples/sample_pipeline/pipeline.py
```diff
@@ -32,8 +35,28 @@
...
     arguments={"chunk_size": 10, "chunk_overlap": 2},
     consumes={"text": "text_data"},
 )
+
+
+@lightweight_component(
+    base_image="python:3.8",
+    extra_requires=[
+        f"fondant[component]@git+https://github.com/ml6team/fondant@"
+        f"{os.environ.get('FONDANT_VERSION', 'main')}",
```
mrchtr, line 49: What happens if we incorporate a bug in the actual branch and using the main branch/version to test the lightweight_component? Assume the test will be successful even if it shouldn't be, no? 

Usually, a user doesn't have to addd fondant as an extra requirement?

examples/sample_pipeline/run.sh
```diff
@@ -27,6 +27,8 @@ trap cleanup EXIT
 # Bind local data directory to pipeline
 data_dir=$(readlink -f "data")
 
+export FONDANT_VERSION=$GIT_HASH
```
mrchtr, line 30: I think you are not using the `GIT_HASH`.

src/fondant/pipeline/lightweight_component.py
```diff
@@ -1,14 +1,24 @@
...
 
 @dataclass
 class Image:
     base_image: str = "fondant:latest"
     extra_requires: t.Optional[t.List[str]] = None
     script: t.Optional[str] = None
 
+    def __post_init__(self):
+        if self.base_image is None:
+            # TODO: link to Fondant version
```
mrchtr, line 19: Is this TODO already somewhere noted down in an issue?

src/fondant/pipeline/lightweight_component.py
```diff
@@ -40,3 +50,47 @@ def image(cls) -> Image:
...
     return wrapper
+
+
+def build_python_script(component_cls: t.Type[Component]) -> str:
+    """Build a self-contained python script for the provided component class, which will act as
+    the `src/main.py` script to execute the component.
+    """
+    imports_source = textwrap.dedent(
+        """\
+        from typing import *
```
mrchtr, line 61: Nitpick: If this might get bigger in the future, maybe it would make sense to extract it to a separate file for readability.

src/fondant/pipeline/lightweight_component.py
RobbeSneyders: It should be part of #782

examples/sample_pipeline/pipeline.py
RobbeSneyders: > Usually, a user doesn't have to addd fondant as an extra requirement?

Indeed, this is only needed until we have a Fondant base image.
We might want to keep defining it as an extra requirement in our tests though to test the latest version.

> What happens if we incorporate a bug in the actual branch and using the main branch/version to test the lightweight_component? Assume the test will be successful even if it shouldn't be, no?

In CI/CD, the current branch is being tested. The `FONDANT_VERSION` environment variable is set in the `run.sh` script. I just provided a back-up to run it locally without this setup.

examples/sample_pipeline/run.sh
RobbeSneyders: The `FONDANT_VERSION` environment variable is used in `pipeline.py` (you commented on it :slightly_smiling_face:)

mrchtr (approved):

src/fondant/pipeline/compiler.py
GeorgesLorre: IMO comparing to a hard to read string is fine

tests/pipeline/test_python_component.py
```diff
@@ -1,21 +1,62 @@
...
+                )
+                return dd.from_pandas(df, npartitions=1)
+    """,
+    )
+
+
+def test_lightweight_component(tmp_path_factory):
+    pipeline = Pipeline(
+        name="dummy-pipeline",
+        base_path="/home/robbe/workspace/fondant/tests/pipeline/data",
```
GeorgesLorre, line 52: base path ?

GeorgesLorre (approved):
Nice !

src/fondant/pipeline/compiler.py
RobbeSneyders: Not sure how valuable it is though, because I would just copy paste the generated string at this moment. It would be the old compiler tests all over again :)

tests/pipeline/test_python_component.py
RobbeSneyders: Good catch. Had some issues with the relative path. Will check if I still run into them.

PR #790, opened 2024-01-17
Enable testing index Weaviate

PhilippeMoussalli:
PR that enables testing the weaviate indexing component 

Leverage the embedded weaviate component to setup the client and bypass the docker-compose setup 

https://weaviate.io/developers/weaviate/installation/embedded


components/index_weaviate/tests/test_component.py
```diff
@@ -0,0 +1,63 @@
...
+    )
+
+    result = query.do()
+
+    return result["data"]["Get"][class_name]
+
+
+def test_index_weaviate_component(monkeypatch):
+    client = weaviate.Client(embedded_options=EmbeddedOptions())
+    monkeypatch.setattr("weaviate.Client", lambda *args, **kwargs: client)
```
RobbeSneyders, line 33: Why is this needed? I didn't need this in the [retrieval component test](https://github.com/ml6team/fondant/blob/main/components/retrieve_from_weaviate/tests/test_component.py).

PhilippeMoussalli: Missed the fact that we have access to a local URL, adjusted!

RobbeSneyders (approved):
Thanks!

PR #793, opened 2024-01-18
Validation for lightweight components

### LINKED ISSUE #783: Add validation to lightweight python component decorator
See @mrchtr's comments here:
https://github.com/ml6team/fondant/pull/770#discussion_r1450164963
https://github.com/ml6team/fondant/pull/770#issuecomment-1889243678

mrchtr:
Adding validation to the PythonComponent. 
- change BaseComponents methods to abstractmethods
- validate that all abstractmethods were implemented in the PythonComponent
- validate that the function interface is the same as in the BaseComponents

Fix #783


src/fondant/pipeline/lightweight_component.py
```diff
@@ -40,6 +40,27 @@ def wrapper(cls):
...
+                ).return_annotation
+                if type_cls_implementation != type_base_cls:
+                    msg = (
+                        f"Invalid function definition of function {function_name}. The return type "
+                        f"has to be {type_base_cls}"
+                    )
+                    raise ValueError(
+                        msg,
+                    )
+
```
RobbeSneyders, line 63: Can you extract this into a separate function with a docstring? :)

src/fondant/pipeline/lightweight_component.py
```diff
@@ -40,6 +40,10 @@ def wrapper(cls):
             script=script,
         )
 
+        validate_required_functions_are_implemented(cls)
+        base_component_cls = cls.__bases__[0]
```
RobbeSneyders, line 44: This might not always get the correct base class.

src/fondant/pipeline/lightweight_component.py
```diff
@@ -49,6 +53,49 @@ def image(cls) -> Image:
 
         return AppliedPythonComponent
 
+    def validate_return_types(base_component_cls, cls_implementation):
```
RobbeSneyders, line 56: Any reason to only check the return type instead of the full signature?

GeorgesLorre:
One thing I noticed while writing tests:   
```
@lightweight_component()
    class Foo(DaskLoadComponent):
        def load(self) -> str:
            return ["bar"]works
```
but
```
@lightweight_component
    class Foo(DaskLoadComponent):
        def load(self) -> str:
            return ["bar"]
```
does not

Notice the () on the decorator. It might be nice to support [both](https://stackoverflow.com/questions/3931627/how-to-build-a-decorator-with-optional-parameters) or at least throw a clear error.

src/fondant/pipeline/lightweight_component.py
mrchtr: Good catch! I've added a separat function, iterate over all base classes and making sure that only one Fondant base class gets inherit.

src/fondant/pipeline/lightweight_component.py
mrchtr: Indeed no reason for it. Changed it.

mrchtr:
> One thing I noticed while writing tests:
> 
> ```
> @lightweight_component()
>     class Foo(DaskLoadComponent):
>         def load(self) -> str:
>             return ["bar"]works
> ```
> 
> but
> 
> ```
> @lightweight_component
>     class Foo(DaskLoadComponent):
>         def load(self) -> str:
>             return ["bar"]
> ```
> 
> does not
> 
> Notice the () on the decorator. It might be nice to support [both](https://stackoverflow.com/questions/3931627/how-to-build-a-decorator-with-optional-parameters) or at least throw a clear error.

Added the support for both variants. When we call the decorator without parentheses the default values of the `Image` will be used.

PhilippeMoussalli (approved):

PR #796, opened 2024-01-18
Enable reranking

PhilippeMoussalli:
Fixes https://github.com/ml6team/fondant/pull/796

Component that enables reranking. 

Reranking seems to currently only support text prompts, for this reason the retrieval component (similar to the recent changes to the indexing) now supports accepting both embeddings and text for the retrieval prompt

### DIFF
--- a/components/retrieve_from_weaviate/fondant_component.yaml
+++ b/components/retrieve_from_weaviate/fondant_component.yaml
@@ -1,14 +1,130 @@
 name: retrieve_from_weaviate
-description: Component that retrieves chunks from a weaviate vectorDB
+description: |
+  Component that retrieves chunks from a Weaviate vector database.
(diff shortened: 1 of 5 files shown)


components/retrieve_from_weaviate/fondant_component.yaml
```diff
@@ -1,14 +1,130 @@
...
+          "top_k": 3,
+          "additional_headers": {
+              "X-OpenAI-Api-Key": "YOUR-OPENAI-API-KEY"
+          }
+      },
+      consumes={
+          "text": "text"
+      }
+  )
+
```
RobbeSneyders, line 56: I think you forgot closing ticks here.

RobbeSneyders (approved):

PR #801, opened 2024-01-19
Add fondant base image

### LINKED ISSUE #782: Add CI/CD to build default Fondant base image

mrchtr:
Fix #782

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,24 @@
+FROM --platform=linux/amd64 python:3.8-slim as base
+
+# System dependencies
+RUN apt-get update && \
+    apt-get upgrade -y && \
+    apt-get install git -y
+
+# Install Fondant
+# This is split from other requirements to leverage caching
+ARG FONDANT_VERSION=main
+RUN pip3 install fondant[component,aws,azure,gcp]@git+https://github.com/ml6team/fondant@${FONDANT_VERSION}
+
+# Set the working directory to the component folder
+WORKDIR /component
+COPY src/ src/
+
+FROM base as test
+COPY tests/ tests/
+RUN pip3 install --no-cache-dir -r tests/requirements.txt
+RUN python -m pytest tests
+
+FROM base
+WORKDIR /component/src
+ENTRYPOINT ["fondant", "execute", "main"]
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,5 @@
-FROM --platform=linux/amd64 python:3.8-slim as base
+ARG PYTHON_VERSION
+FROM --platform=linux/amd64 python:${PYTHON_VERSION}-slim
 
 # System dependencies
 RUN apt-get update && \
@@ -10,15 +11,3 @@ RUN apt-get update && \
 ARG FONDANT_VERSION=main
 RUN pip3 install fondant[component,aws,azure,gcp]@git+https://github.com/ml6team/fondant@${FONDANT_VERSION}
 
-# Set the working directory to the component folder
-WORKDIR /component
-COPY src/ src/
-
-FROM base as test
-COPY tests/ tests/
-RUN pip3 install --no-cache-dir -r tests/requirements.txt
-RUN python -m pytest tests
-
-FROM base
-WORKDIR /component/src
-ENTRYPOINT ["fondant", "execute", "main"]
--- a/src/fondant/pipeline/lightweight_component.py
+++ b/src/fondant/pipeline/lightweight_component.py
@@ -1,23 +1,45 @@
 import inspect
 import itertools
+import sys
 import textwrap
 import typing as t
 from dataclasses import dataclass
 from functools import wraps
+from importlib.metadata import version
 
 from fondant.component import Component
+
+MIN_PYTHON_VERSION = (3, 8)
+MAX_PYTHON_VERSION = (3, 10)
 
 
 @dataclass
 class Image:
-    base_image: str = "fondant:latest"
+    base_image: str
     extra_requires: t.Optional[t.List[str]] = None
     script: t.Optional[str] = None
 
     def __post_init__(self):
         if self.base_image is None:
-            # TODO: link to Fondant version
-            self.base_image = "fondant:latest"
+            self.base_image = self.resolve_fndnt_base_image()
+
+    @staticmethod
+    def resolve_fndnt_base_image(use_ecr_registry=False):
+        """Resolve the correct fndnt base image using python version and fondant version."""
+        # Set python version to latest supported version
+        python_version = sys.version_info
+        if MIN_PYTHON_VERSION <= python_version < MAX_PYTHON_VERSION:
+            python_version = "3.10"
+        else:
+            python_version = f"{python_version.major}.{python_version.minor}"
+
+        fondant_version = version("fondant")
+        basename = (
+            "fndnt/fondant-base"
+            if use_ecr_registry
+            else "public.ecr.aws/fndnt/fondant-base"
+        )
+        return f"{basename}:{fondant_version}-python{python_version}"
 
 
 class PythonComponent:
--- a/.github/workflows/prep-release.yaml
+++ b/.github/workflows/prep-release.yaml
@@ -62,6 +62,9 @@ jobs:
       - name: Build data explorer
         run: ./scripts/build_explorer.sh -t $GITHUB_REF_NAME
 
+      - name: Build data explorer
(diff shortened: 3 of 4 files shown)


docker/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+
+# System dependencies
+RUN apt-get update && \
+    apt-get upgrade -y && \
+    apt-get install git -y
+
+# Install Fondant
+# This is split from other requirements to leverage caching
+ARG FONDANT_VERSION=main
+RUN pip3 install fondant[component,aws,azure,gcp]@git+https://github.com/ml6team/fondant@${FONDANT_VERSION}
```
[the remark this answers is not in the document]
PhilippeMoussalli: not sure if it's worth it, the extra dependencies are quite lightweight (gcsfs, adlfs, s3fs)

src/fondant/pipeline/lightweight_component.py
```diff
@@ -1,23 +1,45 @@
...
-            self.base_image = "fondant:latest"
+            self.base_image = self.resolve_fndnt_base_image()
+
+    @staticmethod
+    def resolve_fndnt_base_image(use_ecr_registry=False):
+        """Resolve the correct fndnt base image using python version and fondant version."""
+        # Set python version to latest supported version
+        python_version = sys.version_info
+        if MIN_PYTHON_VERSION <= python_version < MAX_PYTHON_VERSION:
+            python_version = "3.10"
```
PhilippeMoussalli, line 32: I guess this should then be 3.11

PhilippeMoussalli (commented):
Thanks Matthias! 

Do we also plan on providing base fondant pytorch images for CUDA components? I guess there we'll have to make more limited choices (which pytorch version, python versions are limited) 

https://github.com/cnstark/pytorch-docker

src/fondant/pipeline/lightweight_component.py
mrchtr: Good catch!

PhilippeMoussalli:
> > Do we also plan on providing base fondant pytorch images for CUDA components? I guess there we'll have to make more limited choices (which pytorch version, python versions are limited)
> 
> I thought about this. I guess we don't need them right now. It can be quite complex and I think when people start using the GPU we should recommend building a custom component the old way. This components won't be lightweight neither.

alright makes sense! Some tests still to be failing though

PhilippeMoussalli (approved):
Thanks :)

mrchtr:
Tested the build process in one of the latest commits.
Successfully executed run here: https://github.com/ml6team/fondant/actions/runs/7626293892/job/20772373591?pr=801

images/Dockerfile
GeorgesLorre, line 1: Not a big fan of the `images` folder name

src/fondant/pipeline/lightweight_component.py
```diff
@@ -1,23 +1,45 @@
...
+        else:
+            python_version = f"{MAX_PYTHON_VERSION[0]}.{MAX_PYTHON_VERSION[1]}"
+
+        fondant_version = version("fondant")
+        basename = (
+            "fndnt/fondant-base"
+            if not use_ecr_registry
+            else "public.ecr.aws/fndnt/fondant-base"
+        )
+        return f"{basename}:{fondant_version}-python{python_version}"
```
GeorgesLorre, line 42: I guess we'll need a check/warning that if the user uses a non fondant image they will need to add fondant as a requirement.

tests/pipeline/test_python_component.py
```diff
@@ -237,10 +247,12 @@ def load(self) -> dd.DataFrame:
...
+    operation_spec_without_image = json.loads(operation_spec)
+    operation_spec_without_image["specification"].pop("image", None)
+
+    assert operation_spec_without_image == {
         "specification": {
             "name": "CreateData",
-            "image": "fondant:latest",
             "description": "python component",
             "consumes": {"additionalProperties": True},
             "produces": {"additionalProperties": True},
```
GeorgesLorre, line 258: No test for custom image name ?

GeorgesLorre (commented):
Nice! Looking very good!

images/Dockerfile
mrchtr: First idea was to name the folder `docker`. But having a docker folder in the root conflicts with the python packaging. Up for other naming ideas! :)

src/fondant/pipeline/lightweight_component.py
mrchtr: Good point! I'll add it.

tests/pipeline/test_python_component.py
mrchtr: I've updated one of the tests to check the image name. The test itself replicated the logic of the function that generates the image name itself. When we are running the tests with tox we have to replace the image name dynamically. Do you have any idea how we can improve this? I think mocking `sys.version_info` could work. But not sure if this is the best option.

tests/pipeline/test_python_component.py
```diff
@@ -110,11 +112,19 @@ def transform(self, dataframe: pd.DataFrame) -> pd.DataFrame:
     )
     assert len(pipeline._graph.keys()) == 1 + 1
     assert pipeline._graph["AddN"]["dependencies"] == ["CreateData"]
+    pipeline._graph["AddN"]["operation"].operation_spec.to_json()
+
+    basename = "fndnt/fondant-base"
+    fondant_version = version("fondant")
+    python_version = sys.version_info
+    python_version = f"{python_version.major}.{python_version.minor}"
+    fondant_image_name = f"{basename}:{fondant_version}-python{python_version}"
```
GeorgesLorre, line 121: This you want to mock ? Tbh I think the way it is now is fine since it will nicely test all the python versions since we run tox with multiple versions.

tests/pipeline/test_python_component.py
GeorgesLorre: See my comment above.

I meant adding a test for when the user uses their own image and then checking if the name is propagated properly.

GeorgesLorre:
Btw I tested the base image in a local pipeline and it worked fine!

```
@lightweight_component(
    base_image="fndnt/fondant-base:dc8d9b208802ee85796238ff9a903c8bf66906e8-python3.8"
)
class FeatOne(PandasTransformComponent):
        dataframe["one"] = dataframe["x"].map(lambda x: x + 1)
        return dataframe
```

tests/pipeline/test_python_component.py
mrchtr: I've extracted the image name generation as fixture and modified a test for checking the custom image. Added a test for checking the extra_requirements warning as well.

GeorgesLorre (approved):
Nice! thx

How come I only see 1 tag on docker hub ?

src/fondant/pipeline/lightweight_component.py
```diff
@@ -1,23 +1,64 @@
...
+            msg = (
+                "You are not using a Fondant default base image, and Fondant is not part of"
+                "your extra requirements. Please make sure that you have installed fondant "
+                "inside your container. Alternatively, you can should add Fondant to "
+                "the extra requirements. \n"
+                "E.g. \n"
+                '@lightweight_component(..., extra_requires=["fondant"])'
+            )
+
+            logger.info(msg)
```
RobbeSneyders, line 43: Would make this at least a warning.

src/fondant/pipeline/lightweight_component.py
```diff
@@ -1,23 +1,64 @@
...
+
+    @staticmethod
+    def resolve_fndnt_base_image(use_ecr_registry=False):
```
RobbeSneyders, line 46: This doesn't seem to be called with `use_ecr_registry=True` anywhere.

RobbeSneyders (commented):
Thanks @mrchtr!

Left some small comments, but looks good!

For the name of the images, I would propose just `fondant`, and I would abbreviate the python versions to eg. py3.11.

RobbeSneyders (approved):

src/fondant/pipeline/lightweight_component.py
mrchtr: It isn't called. Thought we need it for the case we need it for the sagemaker runner. But I guess its fine to remove it.

PR #802, opened 2024-01-19
Update caching arguments to include `ComponentOp` and the `Image` class

PhilippeMoussalli:
PR that updates the cache key estimation to take into account the `ComponentSpec` instead of the `ComponentOp` (operation op has both the component spec and the inner `produces` and `consumes` as well as the `Image`

Tested on the example pipeline and caching seems to work fine for lightweight component, only issue is that it will always download the requirements (fondant + extra user dependencies) before skipping the execution of the component. This can become frustrating for iterative development with an increase of dependencies and components. 

We might be able to mitigate this by providing base images with pre-installed Fondant version and move the installation of the extras after the cache hit estimation. Best to tackle this in a separate PR. Issue ticket [here](https://github.com/ml6team/fondant/issues/805) 

The PR also include some small changes to function names to best reflect what they do


examples/sample_pipeline/data/sample.parquet
PhilippeMoussalli, line 1: Updated the parquet file to include more rows since I was running into issue with the small dataset (probably due to empty partitions)

src/fondant/core/component_spec.py
```diff
@@ -94,8 +94,8 @@ def __init__(
         image: str,
         *,
         description: t.Optional[str] = None,
-        consumes: t.Optional[t.Dict[str, t.Union[str, pa.DataType]]] = None,
-        produces: t.Optional[t.Dict[str, t.Union[str, pa.DataType]]] = None,
+        consumes: t.Optional[t.Dict[str, t.Union[str, pa.DataType, bool]]] = None,
```
PhilippeMoussalli, line 97: to accommodate for 'additional_field=True`

GeorgesLorre (approved):
Thx ! We'll see about the requirements in another PR

PR #804, opened 2024-01-19
Support lightweight Python components on Sagemaker

### LINKED ISSUE #755: Support Lightweight Python components in Sagemaker runner
We should be able to execute Lightweight Python components with our Sagemaker runner.

While the other runners might be able to include the full Python source code as part of the entrypoint command, this is not possible with Sagemaker due to the [256 character limit it imposes](https://github.com/ml6team/fondant/issues/593). Instead, we probably want to upload the source code as an additional file for the script processor to use.

RobbeSneyders:
Fixes #755 

image

### DIFF
--- a/src/fondant/pipeline/compiler.py
+++ b/src/fondant/pipeline/compiler.py
@@ -571,7 +571,7 @@ class SagemakerCompiler(Compiler):  # pragma: no cover
                 msg,
             )
 
-    def _get_build_command(
+    def _build_command(
         self,
         metadata: Metadata,
         arguments: t.Dict[str, t.Any],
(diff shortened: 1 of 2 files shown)


src/fondant/pipeline/compiler.py
```diff
@@ -639,19 +639,24 @@ def _check_ecr_pull_through_rule(self) -> None:
...
-                repositoryNames=[f"{self.ecr_namespace}/{full_ref}"],
+                repositoryNames=[repository_name],
             )
-            uri = repo_response["repositories"][0]["repositoryUri"] + ":" + tag
+            return repo_response["repositories"][0]["repositoryUri"] + ":" + tag
+
+        if ref == "fndnt":
+            logging.info("Reusable component detected, patching URI")
+            uri = pull_through(f"{self.ecr_namespace}/{full_ref}")
+        elif ref == "public.ecr.aws":
```
mrchtr, line 657: Maybe we should start to collect constant string values at one place.

src/fondant/pipeline/compiler.py
```diff
@@ -639,19 +639,24 @@ def _check_ecr_pull_through_rule(self) -> None:
...
+            logging.info("Public AWS ECR component detected, patching URI")
+            uri = pull_through(f"{self.ecr_namespace}/{'/'.join(repo)}")
         else:
             logging.info("Custom component detected")
```
mrchtr, line 661: Not related to your changes, but shouldn't we test here that the image is in a AWS registry any way? If the custom image isn't stored in ECR it will not work, right?

PhilippeMoussalli: Typically those things checks occur at runtime (e.g. in KFP). The local service account available at compile time might differ from the one at runtime and they may have different access rights to the custom image.

PhilippeMoussalli (approved):
Thanks Robbe!

src/fondant/pipeline/compiler.py
GeorgesLorre: It is also the user responsibility to make sure the component is available here, we have to assume it is there. 
We have docs on this: https://github.com/ml6team/fondant/blob/main/docs/runners/sagemaker.md#using-custom-fondant-components-on-sagemaker

GeorgesLorre (approved):
Nice ! minimal changes and maximal impact

src/fondant/pipeline/compiler.py
RobbeSneyders: Agree with defining reused string values, however this one is only used once.

Issue #813, opened 2024-01-24
Enable caching for write components 


PhilippeMoussalli:
Write components do not write any new manifests since they don't alter the data. However, the cache estimation is based on checking for existing manifests based on the cache key. We need to somehow enable this either by writing a dummy manifest (propagating the previous one) or handling the logic separately. 

Reproducible example: 

```python
import dask.dataframe as dd
import pandas as pd
import pyarrow as pa

from fondant.component import (
    DaskLoadComponent,
    DaskWriteComponent,
    PandasTransformComponent,
)
from fondant.pipeline import Pipeline, lightweight_component

pipeline = Pipeline(
    name="all-bio-is-just-chemistry",
    base_path="/Users/georgeslorre/ML6/internal/express/sweet-pipelines/pure/artifacts",
)


@lightweight_component(
    base_image="fndnt/fondant-base:dc8d9b208802ee85796238ff9a903c8bf66906e8-python3.8"
)
class CreateData(DaskLoadComponent):
    def load(self) -> dd.DataFrame:
        df = pd.DataFrame(
            {
                "x": [1, 2, 3],
                "y": [4, 5, 6],
            },
            index=pd.Index(["a", "b", "c"], name="id"),
        )
        return dd.from_pandas(df, npartitions=1)


dataset = pipeline.read(
    ref=CreateData,
    produces={"x": pa.int32(), "y": pa.int32()},
)


@lightweight_component(
    base_image="fndnt/fondant-base:dc8d9b208802ee85796238ff9a903c8bf66906e8-python3.8"
)
class FeatOne(PandasTransformComponent):
    def transform(self, dataframe: pd.DataFrame) -> pd.DataFrame:
        dataframe["one"] = dataframe["x"].map(lambda x: x + 1)
        return dataframe


f1 = dataset.apply(
    ref=FeatOne,
    produces={"x": pa.int32(), "y": pa.int32(), "one": pa.int32()},
    consumes={"x": pa.int32(), "y": pa.int32()},
)


@lightweight_component(
    base_image="fndnt/fondant-base:dc8d9b208802ee85796238ff9a903c8bf66906e8-python3.8"
)
class WriteOut(DaskWriteComponent):
    def write(self, dataframe: dd.DataFrame) -> None:
        dataframe.to_parquet("./artifacts/out")


_ = f1.write(
    ref=WriteOut,
    consumes={"x": pa.int32(), "y": pa.int32(), "one": pa.int32()},
)
```

PR #814, opened 2024-01-24
Enable write components to cache

PhilippeMoussalli:
Fixes https://github.com/ml6team/fondant/issues/813

### DIFF
--- a/src/fondant/component/executor.py
+++ b/src/fondant/component/executor.py
@@ -336,6 +336,7 @@ class Executor(t.Generic[Component]):
             component,
             manifest=input_manifest,
         )
+
         output_manifest = input_manifest.evolve(
             operation_spec=self.operation_spec,
             run_id=self.metadata.run_id,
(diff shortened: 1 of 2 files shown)


GeorgesLorre (approved):
Easy enough

RobbeSneyders:
As discussed during standup, not sure if it makes sense to create a manifest for a write component, since it should not output a dataset.

I would also assume that the manifest it produces, points to a non-existing index. Since the index reference in the manifest is updated, but the index is not actually written.

PhilippeMoussalli:
> As discussed during standup, not sure if it makes sense to create a manifest for a write component, since it should not output a dataset.
> 
> I would also assume that the manifest it produces, points to a non-existing index. Since the index reference in the manifest is updated, but the index is not actually written.

Makes sense, changed the approach. For read/transform components we write the cache file with reference to the cached manifest path, for write components we simply write the same file but with an empty string (no reference) since there is no written manifest. The cache estimation still occurs based on the cache key value

RobbeSneyders (approved):
Thanks!

PR #815, opened 2024-01-25
Simplify component naming

GeorgesLorre:
This fixes the inconsistent folder names between runs (cached vs non-cached) and cleans up the different component/folder naming logic we have.


src/fondant/core/component_spec.py
```diff
@@ -101,7 +101,7 @@ def __init__(
         tags: t.Optional[t.List[str]] = None,
     ):
         spec_dict: t.Dict[str, t.Any] = {
-            "name": name,
+            "name": self.sanitized_component_name(name),
```
GeorgesLorre, line 104: This should be the only place where the incoming name is cleaned and stored

src/fondant/pipeline/pipeline.py
```diff
@@ -305,6 +300,10 @@ def _get_registry_path(name: str) -> Path:
             raise ValueError(msg)
         return component_dir
 
+    @property
+    def component_id(self) -> str:
```
RobbeSneyders, line 304: Is there a reason to call this `id` here instead of `name`?

GeorgesLorre: I'll update it

src/fondant/pipeline/compiler.py
```diff
@@ -353,48 +354,55 @@ def from_fondant_component_spec(
             },
         }
 
-        cleaned_component_name = fondant_component.sanitized_component_name
-
+        kfp_safe_name = (
```
mrchtr, line 357: Nice name! Makes it clear.

src/fondant/pipeline/compiler.py
```diff
@@ -353,48 +354,55 @@ def from_fondant_component_spec(
...
-                            "taskInfo": {"name": cleaned_component_name},
+                            "taskInfo": {"name": kfp_safe_name},
                         },
                     },
                 },
                 "inputDefinitions": input_definitions,
             },
             "schemaVersion": "2.1.0",
-            "sdkVersion": "kfp-2.0.1",
+            "sdkVersion": "kfp-2.6.0",
```
mrchtr, line 405: Any side effects by increasing the version?

mrchtr (approved):
Thanks @GeorgesLorre!

src/fondant/pipeline/compiler.py
GeorgesLorre: No, now it is in line with the KFP version we use now.

PR #817, opened 2024-01-26
Update lightweight docs

PhilippeMoussalli:
Changed the name of the components for consistency. 

Custom components -> can be **Lightweight components** or **Containerized components** 

The name **python lightweight/containerized component** makes a bit less sense imo since everything is python based 

Other changes:
* resized the architecture image since it would take some time to load in the the docs


mrchtr (approved):
Thanks @PhilippeMoussalli.

PR #819, opened 2024-01-26
Simplify component init interface

PhilippeMoussalli:
Removes the need to having to define `**kwargs` when the user defines the component `__init__` function. This was needed before to pass the the `produces` and `consumes` to some read/write components without the user having to explicitly define them. This was resolved by removing those two arguments from the component signature. 

Tested on a pipeline with load/write components.


GeorgesLorre (approved):
Nice and easy

PR #820, opened 2024-01-30
Use fondant dev image if fondant dev version is installed

mrchtr:
When fondant is installed locally (fondant_version=0.1.dev0) the base image tag is set to `dev` automatically. 
The previous PR (#818) was merged to fast without this change.


.github/workflows/build.yaml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+  pull_request:
```
mrchtr, line 7: Has to be reverted before we merge it.

scripts/build_base_image.sh
```diff
@@ -4,46 +4,32 @@ set -e
 function usage {
   echo "Usage: $0 [options]"
   echo "Options:"
-  echo "  -t, --tag <value>        Tag to add to image, repeatable"
+  echo "  -t, --tag <value>        Set the tag (default: latest)"
+  echo "  -v, --version <value>    Fondant version"
```
RobbeSneyders, line 8: What's the reason to split these? Looking at how they're used in the Github actions, I would expect it to work with a single value in all those cases?

tests/pipeline/test_python_component.py
```diff
@@ -290,11 +282,40 @@ def load(self) -> dd.DataFrame:
...
+    micro: int
+
+    def __lt__(self, other):
+        return (self.major, self.minor, self.micro) <= other
+
+    def __ge__(self, other):
+        return other < (self.major, self.minor, self.micro)
+
+
+def test_fndnt_base_image_resolution():
```
RobbeSneyders, line 308: Might be nice to mock `version` as well in the same way?

RobbeSneyders (commented):
Thanks @mrchtr!

scripts/build_base_image.sh
mrchtr: I've split it into two cause otherwise the order of the tags in the list is relevant. 
In the component build script the first tag is used to determine the version of fondant. 
E.g. defining `[...] -t GITHUB_HASH -t dev` would work, but not `[...] -t dev -t GITHUB_HASH`.

scripts/build_base_image.sh
```diff
@@ -4,46 +4,30 @@ set -e
 function usage {
   echo "Usage: $0 [options]"
   echo "Options:"
-  echo "  -t, --tag <value>        Tag to add to image, repeatable"
+  echo "  -t, --tag <value>        Set the tag (default: latest)"
```
RobbeSneyders, line 7: I don't see `latest` being set as default anywhere :monocle_face:

RobbeSneyders (approved):
Thanks @mrchtr!

PR #821, opened 2024-01-30
Fix Readme script generation

GeorgesLorre:
This fixes the failing builds but is it what we want ?

### DIFF
--- a/scripts/component_readme/generate_readme.py
+++ b/scripts/component_readme/generate_readme.py
@@ -2,7 +2,6 @@ import argparse
 from pathlib import Path
 
 import jinja2
-
 from fondant.core.component_spec import ComponentSpec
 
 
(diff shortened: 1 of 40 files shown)


RobbeSneyders:
Looking at the hub, I don't mind the underscored look. However I don't like the similarity (but not exact match) with the id, since I think this can be confusing.

I would either try to use the original name, or just use the id instead.

scripts/component_readme/generate_readme.py
```diff
@@ -21,14 +20,17 @@ def generate_readme(component_spec: ComponentSpec, *, component_dir: Path) -> st
     return template.render(
         id=component_dir.name,
         name=component_spec.name,
-        component_folder_name=component_spec.component_folder_name,
+        component_folder_name=component_spec.name,
```
RobbeSneyders, line 23: Can we remove the `component_folder_name` argument and just use the `id` everywhere in its place instead?

RobbeSneyders (commented):
I believe :pray:

RobbeSneyders (approved):

PR #829, opened 2024-01-30
Add produces to lightweight component

PhilippeMoussalli:
PR that adds the `produces` argument to the lightweight component


src/fondant/pipeline/lightweight_component.py
```diff
@@ -135,12 +139,26 @@ def get_spec_consumes(
 
         return spec_consumes
 
+    @classmethod
+    def get_spec_produces(cls):
+        """Get the produces spec for the component."""
+        produces = cls.produces()
+
+        if produces is None:
+            return produces
```
RobbeSneyders, line 148:
```suggestion
            return {"additionalProperties": True}
```

I think we should keep this as the default. That should fix the integration test as well.

PhilippeMoussalli: Makes sense, added!

RobbeSneyders (approved):

PR #832, opened 2024-01-31
Fix code inspection notebook

PhilippeMoussalli:
When running the lightweight components in a notebook, the component class code inspection fails with error `source code is not available`. 
Full error log here [log.txt](https://github.com/ml6team/fondant/files/14114022/new.3.txt)

The error is related to the fact that `inspect` relies on the class located in a source file which is not the case in a notebook where everything runs interactively. This is described in details [here](https://stackoverflow.com/questions/35854373/python-get-source-code-of-class-using-inspect#:~:text=getsource%20works%20by%20actually%20opening%20the%20source%20file%20that%20defines%20the%20object%20you%27re%20looking%20at.)

Implemented the workaround suggested [here](https://github.com/ipython/ipython/issues/11249)

### DIFF
--- a/src/fondant/pipeline/lightweight_component.py
+++ b/src/fondant/pipeline/lightweight_component.py
@@ -9,12 +9,13 @@ from functools import wraps
 from importlib import metadata
 
 import pyarrow as pa
+from IPython import get_ipython
+from IPython.core.magics.code import extract_symbols
 
 from fondant.component import BaseComponent, Component
 from fondant.core.schema import Field, Type
 
 logger = logging.getLogger(__name__)
-
 
 MIN_PYTHON_VERSION = (3, 8)
 MAX_PYTHON_VERSION = (3, 11)
@@ -264,6 +265,33 @@ def lightweight_component(
     return wrapper
 
 
+def new_getfile(_object, _old_getfile=inspect.getfile):
+    if not inspect.isclass(_object):
+        return _old_getfile(_object)
+
+    # Lookup by parent module (as in current inspect)
+    if hasattr(_object, "__module__"):
+        object_ = sys.modules.get(_object.__module__)
+        if hasattr(object_, "__file__"):
+            return object_.__file__
+
+    # If parent module is __main__, lookup by methods (NEW)
+    for name, member in inspect.getmembers(_object):
+        if (
+            inspect.isfunction(member)
+            and _object.__qualname__ + "." + member.__name__ == member.__qualname__
+        ):
+            return inspect.getfile(member)
+
+    msg = f"Source for {_object!r} not found"
+    raise TypeError(msg)
+
+
+def is_running_in_jupyter():
+    shell = get_ipython().__class__.__name__
+    return shell == "ZMQInteractiveShell"
+
+
 def build_python_script(component_cls: t.Type[Component]) -> str:
     """Build a self-contained python script for the provided component class, which will act as
     the `src/main.py` script to execute the component.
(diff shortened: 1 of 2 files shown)


src/fondant/pipeline/lightweight_component.py
```diff
@@ -264,6 +263,39 @@ def consumes(cls) -> t.Optional[t.Dict[str, t.Dict[t.Any, t.Any]]]:
...
+    raise TypeError(msg)
+
+
+def is_running_in_notebook():
+    """Check if the code is running in a Jupyter notebook."""
+    try:
+        from IPython import get_ipython
+
+        shell = get_ipython().__class__.__name__
+        return shell == "ZMQInteractiveShell"
```
RobbeSneyders, line 294: What about running in an IPython terminal instead of a notebook? (if `shell == "TerminalInteractiveShell"`)

src/fondant/pipeline/lightweight_component.py
```diff
@@ -281,7 +313,21 @@ def build_python_script(component_cls: t.Type[Component]) -> str:
     """,
     )
 
-    component_source = inspect.getsource(component_cls)
+    if is_running_in_notebook():
+        from IPython.core.magics.code import extract_symbols
+
+        inspect.getfile = new_getfile
```
RobbeSneyders, line 319: Is this line needed if we're calling `new_getfile` directly below?

src/fondant/pipeline/lightweight_component.py
```diff
@@ -281,7 +313,21 @@ def build_python_script(component_cls: t.Type[Component]) -> str:
...
+        component_source = "".join(
+            inspect.linecache.getlines(  # type: ignore[attr-defined]
+                new_getfile(component_cls),
+            ),
+        )
+        component_source = extract_symbols(component_source, component_cls.__name__)[0][
+            0
+        ]
```
RobbeSneyders, line 327: Y u do this Black :cry: 

```suggestion
        component_source = extract_symbols(
            component_source, 
            component_cls.__name__,
        )[0][0]
```

Can you try changing it to this and see if Black keeps it this way? :stuck_out_tongue:

RobbeSneyders (commented):
Thanks @PhilippeMoussalli 

Would be nice to include this in our test suite (eg. using https://stackoverflow.com/questions/40172281/unit-tests-for-functions-in-a-jupyter-notebook). Can you either include it in this PR or open a ticket to add it later instead?

src/fondant/pipeline/lightweight_component.py
PhilippeMoussalli: haha it does this
```python
  component_source = extract_symbols(
      component_source,
      component_cls.__name__,
  )[
      0
  ][0]
```

PhilippeMoussalli: i'll break it into two lines

PhilippeMoussalli:
> Thanks @PhilippeMoussalli
> 
> Would be nice to include this in our test suite (eg. using https://stackoverflow.com/questions/40172281/unit-tests-for-functions-in-a-jupyter-notebook). Can you either include it in this PR or open a ticket to add it later instead?

Best to tackle it in a separate PR since it seems a bit tricky, opened an [issue](https://github.com/ml6team/fondant/issues/834)

src/fondant/pipeline/lightweight_component.py
RobbeSneyders: Does this fix work for the other shells as well?

PhilippeMoussalli: I tested on my IDE python consol shell (`PyDevTerminalInteractiveShell`), jupter notebook (`ZMQInteractiveShell`) and colab (`Shell`). I presume those are the main 3 options that will be used for compilation (even IDE console seems a bit unlikely)

mrchtr (approved):

PR #835, opened 2024-02-01
Adjust packages import lightweight components

PhilippeMoussalli:
PR that adjusts the import defines in lightweight components to be defined on top of the packaged file. This is to have imports available at a global scope. With the previous implementation, imports defined in `__init__` or other functions in the components were only available within those functions.


RobbeSneyders:
I don't think this is something we should do, since we are taking code that doesn't run in native Python an we make it run when packaged with Fondant. I don't believe this magic will be straightforward to users.

How I think we can tackle this, is by finding global imports used by the user code, and adding those to the script. This will take code that runs in native Python and makes sure it keeps running when packaged with Fondant.

PR #837, opened 2024-02-02
fix typos

andres-vv:
Most of my remarks were in the custom component page. But I see that this page has been removed in the meantime.

The image I removed had a black background which was not very visible on the documentation. But since custom component page is removed it is not used anymore. Hence the removal.

Your welcome for my sick contribution.


GeorgesLorre (approved):
TLDR

PR #840, opened 2024-02-05
Install only test deps before running tox

RobbeSneyders:
Tox will install the necessary dependencies per environment, so we don't need to do this beforehand. This should speed up the builds a bit.


RobbeSneyders:
Doesn't seem to speed up a lot. I think Poetry still resolves the full group of dependencies, but only installs the test ones. Still an improvement though.

GeorgesLorre (approved):

PR #842, opened 2024-02-05
Fix exception when invoke consumes with invalid field schema

mrchtr:
Encountered a bug in exception message generation. We were trying to access the `name` field of the `ComponentOp`, which is no longer available. We are now using `component_op.component_name` instead.


RobbeSneyders (approved):
Thanks @mrchtr!

PR #847, opened 2024-02-07
Create local artifact directory if it does not exist

mrchtr:
This bothers me already for some time. Change the mounting of the local base path in the docker compile. Now a local base path is created.

### DIFF
--- a/src/fondant/pipeline/compiler.py
+++ b/src/fondant/pipeline/compiler.py
@@ -11,6 +11,7 @@ from dataclasses import asdict
 from pathlib import Path
 
 import yaml
+from fsspec.registry import known_implementations
 
 from fondant.core.component_spec import ComponentSpec
 from fondant.core.exceptions import InvalidPipelineDefinition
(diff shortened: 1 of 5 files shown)


src/fondant/pipeline/compiler.py
```diff
@@ -137,23 +138,45 @@ def _patch_path(base_path: str) -> t.Tuple[str, t.Optional[DockerVolume]]:
         if local it patches the base_path and prepares a bind mount
         Returns a tuple containing the path and volume.
         """
+
+        def is_remote_path(path: Path) -> bool:
+            """Check if the path is remote."""
+            _path = str(path)
+            prefixes = set(known_implementations.keys()) - {"local", "file"}
+            return any(_path.startswith(prefix) for prefix in prefixes)
```
RobbeSneyders, line 146: Wouldn't it be easier to swap this check and check if it's local instead?

tests/test_cli.py
```diff
@@ -270,7 +270,7 @@ def test_local_run(mock_docker_installation):
     """Test that the run command works with different arguments."""
     args = argparse.Namespace(
         local=True,
-        ref="some/path",
+        ref=__name__,
```
RobbeSneyders, line 273: Why are we using `__name__` here?

tests/pipeline/test_runner.py
```diff
@@ -59,8 +59,11 @@ def test_docker_runner(mock_docker_installation):
         )
 
 
-def test_docker_runner_from_pipeline(mock_docker_installation):
-    with mock.patch("subprocess.call") as mock_call:
+def test_docker_runner_from_pipeline(mock_docker_installation, tmp_path_factory):
+    with mock.patch("subprocess.call") as mock_call, tmp_path_factory.mktemp(
+        "temp",
+    ) as fn:
+        PIPELINE.base_path = str(fn)
```
RobbeSneyders, line 66: Probably not ideal to set an attribute on a global variable since it creates a side effect? I would either pull it into the test, or make it a pipeline factory function that takes a base path.

tests/pipeline/test_compiler.py
```diff
@@ -168,6 +169,7 @@ def test_docker_compiler(setup_pipeline, tmp_path_factory):
     example_dir, pipeline, _ = setup_pipeline
     compiler = DockerCompiler()
     with tmp_path_factory.mktemp("temp") as fn:
+        pipeline.base_path = str(fn)
```
RobbeSneyders, line 172: Not a big fan of this. Can we turn the fixture into a factory function?

See [this example from connexion](https://github.com/spec-first/connexion/blob/f8f461c5240f6d01b0d8c7104387993ce811b2c7/tests/conftest.py#L75).

RobbeSneyders (commented):
Thanks @mrchtr, nice of you to pick this up :stuck_out_tongue:

tests/pipeline/test_compiler.py
mrchtr: If it's fine for you I would exclude it here. I'll create an separat issue for this. I took a look into the test and I think it would result in a bigger refactoring. It would make indeed sense to have a pipeline factory cause it is needed in several test cases. Not only here in this file.

tests/test_cli.py
mrchtr: I don't understand completely why the tests have worked before. The actual module contains a test pipeline which is used as reference for the test cases. `__name__` makes more sense here.
However, I would cleanup the test cases as well. I think we can extract some fixtures here and make it more robust.

src/fondant/pipeline/compiler.py
mrchtr: Not sure. The remote prefixes are well defined by the `known_implementations`. The local paths can be more complex (start with prefix, without prefix, absolute path, relative path, ...). Thought a check for a remote path is easier.

tests/pipeline/test_compiler.py
GeorgesLorre: We already have code like this that we could reuse more.

eg: https://github.com/ml6team/fondant/blob/7ec04dcbe3e2aa6a63e91d28e201bbff35304abf/tests/pipeline/test_compiler.py#L132

This is one is even parameterized so you can easily run the tests on multiple pipelines.

RobbeSneyders (approved):

PR #849, opened 2024-02-07
Make Fondant installable via git

RobbeSneyders:
This PR moves our components into the source folder so Fondant becomes installable via git.

The reason to have them at the top level before was because we didn't have a hub. Now that we have the hub page on the fondant.ai site, I think the trade-off has shifted.

I will update the references in the use case repositories as well.


mrchtr (commented):
Is the merge to feature/remove-reusable-RAG-components on purpose?

RobbeSneyders:
Yes, since I didn't want to move the components that you are deleting. We can wait with merging until after that one is merged.

GeorgesLorre (approved):
nice QoL

mrchtr (approved):

PR #852, opened 2024-02-14
Move Dask Client configuration to Component class and use multi-GPU in `embed_images` component

RobbeSneyders:
It doesn't really make sense to let the user configure the Dask client used by a component, since this is specific to the component implementation. I therefore moved it to the Component class with a default implementation of the Dask `LocalCluster`. Note that due to a bug the `LocalCluster` was not yet used as default. This therefore changes the default for all components, which were using the threaded scheduler before.

One downside to the current implementation is that we require the component developer to call `super().__init__()` in their `__init__` method. This is best practice though and highlighted by IDEs.

### DIFF
--- a/docs/components/components.md
+++ b/docs/components/components.md
@@ -1,3 +1,5 @@
+from distributed import Client
+
 # Components
 
 Fondant makes it easy to build data preparation pipelines leveraging reusable components. Fondant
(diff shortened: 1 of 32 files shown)


GeorgesLorre:
No explicit tests ?

Do we need to update the docs ?

RobbeSneyders:
> No explicit tests ?

Do you have a proposal for a useful test scenario?

> Do we need to update the docs ?

I'll add something indeed.

mrchtr (commented):
I think we should set the `LocalCluster` for as default for all components and let the user overwrite the client if needed. 
The other components still not using the `LocalCluster`, but it is the recommended way. Is there any reason why we shouldn't set the `LocalCluster` as default for the other components?

RobbeSneyders:
@mrchtr with this change we are setting the `LocalCluster` as default for all components.

mrchtr:
> @mrchtr with this change we are setting the `LocalCluster` as default for all components.

Ah my bad, nvm

GeorgesLorre (approved):
👍

Issue #565, opened 2023-10-29
General inference component


RobbeSneyders:
Implementing GPU components can be a bit tricky since we need to make sure that both preprocessing and inference are batched, would be nice if we can implement a generic inference transform component class that faciltiates this. A bit similar to Beam's [run inference API](https://beam.apache.org/documentation/transforms/python/elementwise/runinference/)

PR #855, opened 2024-02-15
Make unique index sorted

RobbeSneyders:
The unique index we create is not sorted, while I believe this was originally the goal. Ran into some issues merging large dataframes and this was one of them.

### DIFF
--- a/src/fondant/components/load_from_csv/src/main.py
+++ b/src/fondant/components/load_from_csv/src/main.py
@@ -72,14 +72,15 @@ class CSVReader(DaskLoadComponent):
                 Function that sets a unique index
                 based on the partition and row number.
                 """
+                n_digits = len(str(len(dataframe)))
+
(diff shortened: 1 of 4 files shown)


GeorgesLorre (approved):
Is this something we can do/check in the parent class ?

mrchtr (commented):
Could we adjust the component test too?

RobbeSneyders:
> Is this something we can do/check in the parent class ?

We've discussed a couple of times already to create better base components, this could indeed be part of that. Seems like we only have an issue for a base inference component (https://github.com/ml6team/fondant/issues/565), but would be useful for load and write components as well.

This is just a quick fix though, would take that up as a separate effort.

> Could we adjust the component test too?

Let's add a test when we move this into a generic load component.

Opened an issue for it: https://github.com/ml6team/fondant/issues/860

PR #856, opened 2024-02-15
Update component dir in build script

RobbeSneyders:
The build pipeline is currently failing. This should fix it.

### DIFF
--- a/scripts/build_components.sh
+++ b/scripts/build_components.sh
@@ -20,7 +20,7 @@ function usage {
 
 # Parse the arguments
 while [[ "$#" -gt 0 ]]; do case $1 in
-  -r |--registry) registry="$2"; shift;;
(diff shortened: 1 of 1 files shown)


scripts/build_components.sh
```diff
@@ -20,7 +20,7 @@ function usage {
 
 # Parse the arguments
 while [[ "$#" -gt 0 ]]; do case $1 in
-  -r |--registry) registry="$2"; shift;;
+  -rg|--registry) registry="$2"; shift;;
```
mrchtr, line 23: Is `-r` still used somewhere?

RobbeSneyders: I don't think so. It was duplicate before, since it's used for `--repo` as well.

mrchtr (approved):

PR #859, opened 2024-02-16
Infer consume operation if not present in dataset interface

### LINKED ISSUE #836: Passing all columns to the write component when no consumes is specified
## Background

When the `consumes`  section in the component specification contains `additionalProperties=true` , it allows passing additional columns along the defined ones to the component.

This is especially needed for reusable write components.

```python
dataset.write(
    "write_to_file",
    "consumes": {
    	"text": pa.string()
    }
)
```

Name-to-name mapping cannot be used. For example:

```python
dataset.write(
    "write_to_file",
    consumes={
        "text_data": "text_data"
    }
)
```

This will raise an InvalidPipelineDefinition, 
`fondant.core.exceptions.InvalidPipelineDefinition: Received a string value for key text_data in the consumes argument passed to the operation, but text_data is not defined in the consumes section of the component spec.`, due to the fact that no columns are specified in the component specification.

Furthermore, if the component is used without defining `consumes`  in the pipeline, no columns will be passed to the component.

```python
dataset.write(
    "write_to_file"
)
```

On the other hand, we have implemented custom logic for lightweight Python components. This allows us to use these components without invoking `consumes`  in the pipeline.

```python
@lightweight_component
class WriteToAnywhere(DaskWriteComponent):
    def write(self, dataframe: dd.DataFrame) -> None:
        print(dataframe.columns)

dataset.write(
    WriteToAnywhere
)
```

Mapping the names from the global dataset to the components dataset is not working too.

## Goal

Ideally we should permit all three options, when `additionalProperties=true`, 

(1) Do not define any `consumes`  in the pipeline interface. This will pass all available columns to the component.

(2) Define column mapping (e.g. `"text_data": "text"`), which maps from the global dataset to the component dataset.

(3) Define which columns to read from the global dataset using pyarrow types, such as `"text_data": pa.string()`.

## Approach

We complete the component specifications of lightweight components using the global schema of the dataset already in the [pipeline.py](https://github.com/ml6team/fondant/blob/main/src/fondant/pipeline/pipeline.py#L224).

A similar approach could be adopted for containerized components, as all the necessary information can be obtained from the component yaml and the schema of the global dataset.

We would have consistent implementation for both lightweight and containerized components at the same location in the code, and the ability to validate and catch potential errors during pipeline initialization. 

This enables (1) for the containerized components. (2) we can achieve this for both component types by using the global dataset schema. Validate the mapping and complete the ComponentSpec. 

(3) is still working, and we could add extra validation.

Alternatively, if the consumes is None and additionalProperties is set to true, we could directly load all columns of the dataset in the `data_io`.

mrchtr:
Basic implementation, I still have to add tests. I wanted to get some feedback first.

- Added a method to the lightweight components to generate a `ComponentSpec` based on the attributes.
- Added a method in the pipeline to infer the consumption based on the `ComponentSpec`.
In cases where a user hasn't specified a `consume` in the pipeline operations, we now infer this. If a component spec contains a `consumes` section and `additionalProperties` are set to true, we load all columns. If `additionalProperties` is set to false, we limit the columns defined in the component spec.

Fix #836

### DIFF
--- a/src/fondant/component/component.py
+++ b/src/fondant/component/component.py
@@ -1,13 +1,10 @@
 """This module defines interfaces which components should implement to be executed by fondant."""
 
-import logging
 import typing as t
 from abc import abstractmethod
 
 import dask.dataframe as dd
 import pandas as pd
-
-logger = logging.getLogger(__name__)
 
 
 class BaseComponent:
--- a/src/fondant/component/component.py
+++ b/src/fondant/component/component.py
@@ -1,10 +1,16 @@
 """This module defines interfaces which components should implement to be executed by fondant."""
 
+import logging
 import typing as t
 from abc import abstractmethod
 
 import dask.dataframe as dd
 import pandas as pd
+import pyarrow as pa
+
+from fondant.core.schema import Field, Type
+
+logger = logging.getLogger(__name__)
 
 
 class BaseComponent:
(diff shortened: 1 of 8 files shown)


tests/pipeline/test_pipeline.py
```diff
@@ -208,23 +208,22 @@ def test_valid_pipeline(
     dataset = pipeline.read(
         Path(components_path / component_names[0]),
         arguments=component_args,
-        produces={"images_array": pa.binary()},
+        produces={"images_data": pa.binary()},
```
mrchtr, line 211: Going to revert this, and extend it with more test cases.

src/fondant/core/component_spec.py
```diff
@@ -214,6 +214,27 @@ def consumes(self) -> t.Mapping[str, Field]:
             },
         )
 
+    @property
+    def consumes_additional_properties(self) -> bool:
+        """Returns a boolean indicating whether the component consumes additional properties."""
+        return self._specification.get("consumes", {}).get(
```
GeorgesLorre, line 220: This might be nice to implement on the spec (same for the produces)

src/fondant/component/component.py
```diff
@@ -1,11 +1,14 @@
 """This module defines interfaces which components should implement to be executed by fondant."""
 
+import logging
 import typing as t
 from abc import abstractmethod
 
 import dask.dataframe as dd
 import pandas as pd
 
+logger = logging.getLogger(__name__)
```
RobbeSneyders, line 10: This logger is not used.

src/fondant/pipeline/pipeline.py
```diff
@@ -185,25 +184,59 @@ def __init__(
...
             **kwargs,
         )
 
+    @classmethod
+    def _infer_consumes(cls, component_spec, dataset_fields):
+        """Infer the consumes section of the component spec."""
+        if component_spec.consumes_is_defined is False:
+            msg = (
+                "The consumes section of the component spec is not defined. "
+                "Cannot infer consumes."
```
RobbeSneyders, line 218: This logging message shows again that a split of names could be useful :sweat_smile:

src/fondant/pipeline/pipeline.py
```diff
@@ -185,25 +184,59 @@ def __init__(
...
             component_dir = Path(path)
         else:
             component_dir = cls._get_registry_path(str(path))
+
         component_spec = ComponentSpec.from_file(
             component_dir / cls.COMPONENT_SPEC_NAME,
         )
 
+        # If consumes is not defined in the pipeline, we will try to infer it
+        if kwargs.get("consumes", None) is None:
```
RobbeSneyders, line 198:
```suggestion
        if kwargs.get("consumes") is None:
```

`None` is the default :slightly_smiling_face:

src/fondant/pipeline/pipeline.py
```diff
@@ -185,25 +184,59 @@ def __init__(
...
+        """Infer the consumes section of the component spec."""
+        if component_spec.consumes_is_defined is False:
+            msg = (
+                "The consumes section of the component spec is not defined. "
+                "Cannot infer consumes."
+            )
+            logger.info(msg)
+            return None
+
+        # Component has consumes and additionalProperties, we will load all dataset columns
```
RobbeSneyders, line 223: Wondering about the case here where there's both a schema and `additionalProperties`. But I guess this is fine for now. We might want to have more logic here once we add support for `additionalProperties` schemas instead of just the boolean.

src/fondant/pipeline/pipeline.py
```diff
@@ -223,31 +256,14 @@ def from_ref(
...
-                    description=description,
-                    consumes=spec_consumes,
-                    produces=spec_produces,
-                    args={
-                        name: arg.to_spec()
-                        for name, arg in infer_arguments(ref).items()
-                    },
-                )
+                # If consumes is not defined in the pipeline, we will try to infer it
+                if kwargs.get("consumes", None) is None:
```
RobbeSneyders, line 262:
```suggestion
                if kwargs.get("consumes") is None:
```

RobbeSneyders (approved):
Thanks @mrchtr!

Left some comments, but they're mostly nits. Looks a lot cleaner now!

src/fondant/core/component_spec.py
GeorgesLorre: I took a better look and never mind. I was lost in the many specs 😛 .

GeorgesLorre (approved):

PR #861, opened 2024-02-19
Change dask_client to general setup method

RobbeSneyders:
This PR changes the `dask_client` method introduced in #852 into a general `setup` method.

This method differs from `__init__` in that it allows users to return a state, which is passed into the `teardown` method by Fondant.  This is necessary since the Dask client is not pickleable, and setting it as an instance attribute leads to issues when executing the `transform` method across processes (as is the case in the `PandasTransformComponent`).

### DIFF
--- a/src/fondant/component/executor.py
+++ b/src/fondant/component/executor.py
@@ -287,6 +287,8 @@ class Executor(t.Generic[Component]):
         component.consumes = self.operation_spec.inner_consumes
         component.produces = self.operation_spec.inner_produces
 
+        state = component.setup()
+
         output_df = self._execute_component(
             component,
             manifest=input_manifest,
@@ -298,7 +300,7 @@ class Executor(t.Generic[Component]):
         )
         self._write_data(dataframe=output_df, manifest=output_manifest)
 
-        component.teardown()
+        component.teardown(state)
 
         return output_manifest
 
--- a/tests/component/test_component.py
+++ b/tests/component/test_component.py
@@ -303,7 +303,7 @@ def test_load_component(metadata):
 
 
 @pytest.mark.usefixtures("_patched_data_writing")
-def test_teardown_method(metadata):
+def test_setup_teardown_methods(metadata):
     # Mock CLI arguments load
     operation_spec = OperationSpec(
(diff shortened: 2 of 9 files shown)


tests/component/test_component.py
```diff
@@ -442,27 +447,23 @@ def test_pandas_transform_component(metadata):
...
 
-    init = patch_method_class(MyPandasComponent.__init__)
-    transform = patch_method_class(MyPandasComponent.transform)
     executor_factory = ExecutorFactory(MyPandasComponent)
     executor = executor_factory.get_executor()
-    with mock.patch.object(MyPandasComponent, "__init__", init), mock.patch.object(
-        MyPandasComponent,
-        "transform",
-        transform,
-    ):
```
RobbeSneyders, line 462: This was giving issues with recursion when pickling. I couldn't completely figure out the cause and how to fix it.

tests/component/test_component.py
```diff
@@ -442,27 +447,23 @@ def test_pandas_transform_component(metadata):
...
-        executor.execute(MyPandasComponent)
-        init.mock.assert_called_once()
-        assert transform.mock.call_count == N_PARTITIONS
+    executor.execute(MyPandasComponent)
+    assert init_called == 1
```
RobbeSneyders, line 466: We cannot take the same approach for the `transform` method, since it's executed across Dask workers. Couldn't figure out how to make this work, so dropped the assertion for the `transform` method for now.

src/fondant/component/executor.py
```diff
@@ -298,7 +300,7 @@ def _run_execution(
         )
         self._write_data(dataframe=output_df, manifest=output_manifest)
 
-        component.teardown()
+        component.teardown(state)
```
mrchtr, line 303: Shouldn't this be `component.teardown(*state)` in case someone returns more as a single state?

RobbeSneyders: This wouldn't work for a single object. I think it's most clear if we just inject exactly what has been returned by the setup method. If it returns a tuple, we inject a tuple.

mrchtr (approved):
Thanks @RobbeSneyders.

PR #862, opened 2024-02-19
Add `gpu` extra with `dask-cuda` and bump minimum Python version to 3.9

RobbeSneyders:
Separately resolving `dask-cuda` in our components leads to dependency conflicts. Since we put a lot of focus on ML components, I think it makes sense to add it as an extra.

It only supports Python 3.9+ though, so we need to drop Python 3.8, which will be end of life later this year.


RobbeSneyders:
Good catch. We have Python versions in a lot of places :sweat_smile:

mrchtr (approved):

PR #865, opened 2024-02-20
Renaming inner and outer consumes/produces

mrchtr:
Propose to rename the inner and outer consumes/produces to something more meaningful. 

Idea was to have self explaining names. An OperationSpec is consuming fields from the dataset, and produces fields to dataset (previous outer). The OperationSpec holding `operation_consumes`, and `operation_produces` (previous inner). As well as the mapping (`operation_to_dataset_mapping` and `dataset_to_operation_mapping`).

### DIFF
--- a/src/fondant/component/data_io.py
+++ b/src/fondant/component/data_io.py
@@ -136,7 +136,9 @@ class DaskDataLoader(DataIO):
             msg = "No data could be loaded"
             raise RuntimeError(msg)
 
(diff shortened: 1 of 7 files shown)


RobbeSneyders:
Thanks @mrchtr. This doesn't change anything user facing right? I'm wondering if we should make a distinction there as well.

mrchtr:
This PR is outdated. The changes here are already part of #867

PR #867, opened 2024-02-20
Add test for consume name to name mapping

### LINKED ISSUE #863: Allow mapping of dataset field names with `additionalProperties: true`

mrchtr:
Fix #863

### DIFF
--- a/src/fondant/core/component_spec.py
+++ b/src/fondant/core/component_spec.py
@@ -353,10 +353,10 @@ class OperationSpec:
         }
         self._validate_mappings()
 
-        self._inner_consumes: t.Optional[t.Mapping[str, Field]] = None
-        self._outer_consumes: t.Optional[t.Mapping[str, Field]] = None
-        self._inner_produces: t.Optional[t.Mapping[str, Field]] = None
-        self._outer_produces: t.Optional[t.Mapping[str, Field]] = None
+        self._operation_consumes: t.Optional[t.Mapping[str, Field]] = None
+        self._consumes_of_dataset: t.Optional[t.Mapping[str, Field]] = None
+        self._operation_produces: t.Optional[t.Mapping[str, Field]] = None
+        self._produces_to_dataset: t.Optional[t.Mapping[str, Field]] = None
 
     def to_dict(self) -> dict:
         def _dump_mapping(
@@ -414,11 +414,14 @@ class OperationSpec:
                     msg = f"Unexpected type {type(value)} received for key {key} in {name} mapping"
                     raise InvalidPipelineDefinition(msg)
 
-    def _inner_mapping(self, name: str) -> t.Mapping[str, Field]:
-        """Calculate the "inner mapping" of the operation. This is the mapping that the component
+    def _dataset_to_operations_mapping(self, name: str) -> t.Mapping[str, Field]:
+        """Calculate the operations mapping.
+        Maps dataset fields to the fields that the operation will receive.
+        This is the mapping that the component
         `transform` (or equivalent) method will receive. This is calculated by starting from the
         component spec section, and updating it with any string to type mappings from the
         argument mapping.
+
 
         Args:
             name: "consumes" or "produces"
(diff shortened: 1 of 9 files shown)


RobbeSneyders (commented):
Don't we need to compile the pipeline to actually test this? I believe it will fail then.

src/fondant/core/component_spec.py
```diff
@@ -464,25 +464,34 @@ def _operation_schema_to_dataset_schema(self, name: str) -> t.Mapping[str, Field
...
+            operations_column_name,
+            dataset_column_name_or_type,
+        ) in column_mapping_args.items():
+            # If the value is not a string, it means it's a type, so we skip it
+            if not isinstance(dataset_column_name_or_type, str):
                 continue
 
-            if key in spec_mapping:
-                mapping[value] = Field(name=value, type=mapping.pop(key).type)
+            if operations_column_name in operations_schema:
```
mrchtr, line 485: We are able to add a check for `addtionalProperties` here, and ignore this exception if `additionalProperties` are allowed. However, in that case we have to change the string mapping to a `string: pa.DataType` mapping to avoid issues later on. The `ComponentOp` would need to access the dataset schema.
I think it might be cleaner to implement this behaviour before we are initialising the `ComponentOp`, infer the consumes, validate and fix it (if needed). The `ComponentOp` will be initialised with a valid consumes definition afterwards.

tests/pipeline/test_pipeline.py
```diff
@@ -674,3 +675,50 @@ def test_infer_consumes_if_not_defined(
         "images_array": {"type": "binary"},
         "embeddings_data": {"items": {"type": "float32"}, "type": "array"},
     }
+
+
+@pytest.mark.parametrize(
```
RobbeSneyders, line 680: Aren't you just parameterizing with a single example here?

tests/pipeline/test_pipeline.py
```diff
@@ -674,3 +675,50 @@ def test_infer_consumes_if_not_defined(
...
+    dataset.apply(
+        Path(components_path / component_names[1]),
+        arguments=component_args,
+        consumes={"images_data": "images_data"},
+    )
+
+    assert dataset.pipeline._graph["third_component"][
+        "operation"
+    ].operation_spec.to_dict()["consumes"] == {
+        "images_data": "images_data",
```
RobbeSneyders, line 723: What are we testing here? Shouldn't this be:
```python
{
    "images_data": pa.binary()
}
```

tests/pipeline/test_pipeline.py
```diff
@@ -332,28 +332,29 @@ def test_invalid_pipeline_schema(
...
-            arguments=component_args,
-            consumes=spec[2]["consumes"],
-        )
-        dataset.apply(
-            Path(components_path / component_names[3]),
-            arguments=component_args,
-            consumes=spec[3]["consumes"],
-        )
-        with pytest.raises(InvalidPipelineDefinition):
+        with pytest.raises(InvalidPipelineDefinition):  # noqa: PT012
```
RobbeSneyders, line 335: Can we put this around the specific code block we expect to fail?

tests/pipeline/test_pipeline.py
```diff
@@ -332,28 +332,29 @@ def test_invalid_pipeline_schema(
...
+                arguments=component_args,
+                consumes=spec[2]["consumes"],
+            )
+            dataset.apply(
+                Path(components_path / component_names[3]),
+                arguments=component_args,
+                consumes=spec[3]["consumes"],
+            )
+
             pipeline._validate_pipeline_definition("test_pipeline")
```
RobbeSneyders, line 358: Do we still need this if the error is now raised during `apply`?

src/fondant/pipeline/pipeline.py
```diff
@@ -145,13 +146,18 @@ def __init__(
         cache: t.Optional[bool] = True,
         resources: t.Optional[Resources] = None,
         component_dir: t.Optional[Path] = None,
+        dataset_fields: t.Optional[t.Mapping[str, Field]] = None,
```
RobbeSneyders, line 149: I'm wondering if we shouldn't split all this `consumes` / `produces` logic into a separate class, eg. a `ComponentOpFactory`. I think it could help with maintainability and testability to keep the `ComponentOp` limited to a representation of the operation, close to a dataclass.

src/fondant/core/component_spec.py
RobbeSneyders: Agree, see my comment below.

RobbeSneyders (commented):
Thanks @mrchtr! I think we can indeed refactor a bit more to split the functionality more cleanly.

PR #868, opened 2024-02-20
Add pre- and post-build script to work around Poetry bug

RobbeSneyders:
The new packaging strategy introduced in #849 doesn't work correctly for wheels due to a bug in poetry (https://github.com/python-poetry/poetry/issues/8994), which breaks installing both from Github and PyPi. This PR introduces a workaround.

Since the bug is related to the `src` directory, this workaround adds a `pre-build/sh` script which unpacks the `src` directory before building, and a `post-build.sh` script which moves the package back into the `src` directory afterwards.

This leads to the following result:
- Installing from Github leads to an installation including all component files
- Installing locally without running the build script leads to an installation including all component files
- Installing from PyPi leads to an installation with only the `fondant_component.yaml` files for each component

In the ideal case, we get the same result for all installation methods, but this already at least leads to a working result for each method.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,8 +21,8 @@ maintainers = [
     "Niels Rogge <niels.rogge@ml6.eu>",
 ]
 repository = "https://github.com/ml6team/fondant"
-include = ["*.txt", "*.rst", "*.md", "src/fondant/components/**/*.yaml"]
-exclude = ["src/fondant/components"]  # Exclude everything except for component specs
+include = ["*.txt", "*.rst", "*.md", "fondant/components/**/*.yaml"]
+exclude = ["fondant/components"]  # Exclude everything except for component specs
 classifiers = [
     "Development Status :: 2 - Pre-Alpha",
     "Intended Audience :: Developers",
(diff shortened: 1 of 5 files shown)


GeorgesLorre (approved):
can we open a ticket on our board to make sure we remove this workaround later?

pyproject.toml
```diff
@@ -75,6 +75,9 @@ vertex = ["docker", "kfp", "google-cloud-aiplatform"]
 sagemaker = ["sagemaker", "boto3"]
 docker = ["docker"]
 
+all = ["dask", "dask-cuda", "s3fs", "adlfs", "gcsfs", "docker", "kfp", "google-cloud-aiplatform",
```
mrchtr, line 78: Guess this is useful for local development in case you want to install all dependencies. Should we adjust the readme?

mrchtr (approved):

pyproject.toml
RobbeSneyders: For local development, this can still be done with `poetry install --all-extras`. I just needed this to install all extras when installing from the wheel, which I think is only needed in our CI/CD.

RobbeSneyders:
> can we open a ticket on our board to make sure we remove this workaround later?

Based on the response of the Poetry issue, I'm not sure if we will be able to :) I would follow up on that one instead of having a blocked issue on our repo.

Issue #870, opened 2024-02-21
Local runner fails to bind port of dask dashboard


mrchtr:
Encountered a bug with:
- fondant `0.11dev1`
- Docker version 25.0.1, build 29cf629
- Docker Compose version v2.24.2

The Docker container cannot be started accordingly because during the start of a component, the port of the Dask dashboard is still allocated from the previous completed component.

E.g.
```
Error response from daemon: driver failed programming external connectivity on endpoint controlnet-pipeline-write_to_file-1 (cc30d244e30a23e8d565000b34c48bbc0e9444b8f706a3b4ecf22db073e7abfe): Bind for 0.0.0.0:8787 failed: port is already allocated
```

Temporary work around could be to disable the dask dashboard port binding, since it isn't used really at the moment.

RobbeSneyders:
Did you run into this issue yourself? Can you check that your environment is clean and you don't have any detached services still active?

Also, we might not have to move the port completely, but just define it differently: https://github.com/docker/compose/issues/4950

RobbeSneyders:
FYI, I can't reproduce this myself.

PR #872, opened 2024-02-22
Temporarily disabled exposing the Dask diagnostic dashboard 

mrchtr:
Temporarily disabled exposing the Dask diagnostic dashboard  port to the docker setup.
Executing docker containers after each other using docker compose, leads sometimes to the case that the port is already in use.

We should come up with a solution if we want to use the dask dashboard again. 

Fix: #870

### DIFF
--- a/src/fondant/pipeline/compiler.py
+++ b/src/fondant/pipeline/compiler.py
@@ -27,6 +27,12 @@ from fondant.pipeline import (
 logger = logging.getLogger(__name__)
 
 DASK_DIAGNOSTIC_DASHBOARD_PORT = 8787
+
+# Temporarily disabled exposing the Dask diagnostic dashboard port to the docker setup.
+# Executing docker containers after each other using docker compose,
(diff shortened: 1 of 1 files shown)


RobbeSneyders:
Could you try the fix mentioned here: https://github.com/docker/compose/issues/4950#issuecomment-342882168

mrchtr:
> Could you try the fix mentioned here: [docker/compose#4950 (comment)](https://github.com/docker/compose/issues/4950#issuecomment-342882168)

It solves the issue but by doing this a random port of host system is used. You can access the dashboard but you have to use docker commands to see which port is used. Not ideal.

RobbeSneyders:
Does it only use a different port if the same one is already used? Because in that case, it still seems like a better solution.

mrchtr:
@RobbeSneyders would you up for testing it on your local machine? For me it seems to be always the random port, even for the first component.

RobbeSneyders (approved):

PR #876, opened 2024-02-26
Add image retrieval from FAISS index

mrchtr:
The component expects a dataset in the form of:
- A Dask dataset containing id mapping (image id to image url)
- The faiss index itself

The component will load the dataset from the remote storage into the component, initialise a CLIP model, 
and retrieve similar images based on given prompts.


src/fondant/components/retrieve_images_from_faiss_index/src/main.py
```diff
@@ -0,0 +1,71 @@
...
+
+import dask.dataframe as dd
+import faiss
+import fsspec
+import pandas as pd
+from fondant.component import DaskTransformComponent
+from transformers import AutoTokenizer, CLIPTextModelWithProjection
+
+ID_MAPPING = "id_mapping"
+FAISS_INDEX = "faiss3"
```
RobbeSneyders, line 11: These shouldn't be hardcoded since they are variables in `autofaiss`. Should we just accept two paths?

src/fondant/components/retrieve_images_from_faiss_index/src/main.py
```diff
@@ -0,0 +1,71 @@
...
+
+            with open("faiss_index", "wb") as out:
+                out.write(file_contents)
+
+        self.search_index = faiss.read_index("faiss_index")
+
+        dataset_url = f"{dataset_url}/{ID_MAPPING}"
+        dataset_index = dd.read_parquet(dataset_url)
+
+        self.image_urls = dataset_index["url"].compute().to_list()
```
RobbeSneyders, line 40: There's probably no benefit in using Dask here.

RobbeSneyders (commented):
Thanks @mrchtr!

Would be ideal if we let the user search with embeddings or images as well. Similar to the `retrieve_laion_by_embedding` component. That should be possible now in a single component with the `additionalFields` support, although maybe not really nice yet without schema support.

We can do it in a separate PR anyway.

src/fondant/components/retrieve_images_from_faiss_index/src/main.py
mrchtr: I think this one has to be a dask component. We are downloading the index from the remote storage. Doing it in a pandas component, we would do it several times. Do you see an option to implement this behaviour in a pandas component?

RobbeSneyders:
Also, we probably want to use a GPU, I did it here for the embedding component:
https://github.com/ml6team/fondant/blob/main/src/fondant/components/embed_images/src/main.py#L43

src/fondant/components/retrieve_images_from_faiss_index/fondant_component.yaml
```diff
@@ -0,0 +1,41 @@
...
+produces:
+  image_url:
+    type: string
+  prompt_id:
+    type: string
+
+previous_index: prompt_id
+
+args:
+  dataset_url:
```
RobbeSneyders, line 25:
```suggestion
  url_mapping_path:
```

src/fondant/components/retrieve_images_from_faiss_index/src/main.py
```diff
@@ -0,0 +1,106 @@
...
+    def __init__(  # noqa PLR0913
+        self,
+        dataset_path: str,
+        faiss_index_path: str,
+        image_index_column_name: str,
+        clip_model: str = "laion/CLIP-ViT-B-32-laion2B-s34B-b79K",
+        num_images: int = 2,
+    ):
+        self.model = CLIPTextModelWithProjection.from_pretrained(clip_model)
+        self.tokenizer = AutoTokenizer.from_pretrained(clip_model)
```
RobbeSneyders, line 29: I think we want to initialize these once per worker. See the [`embed_images`](https://github.com/ml6team/fondant/blob/1ff9b1a7639d68d2f951905b01cac0ba104dc556/src/fondant/components/embed_images/src/main.py#L97-L104) component for example.

src/fondant/components/retrieve_images_from_faiss_index/fondant_component.yaml
```diff
@@ -0,0 +1,41 @@
...
+    description: Url of the dataset
+    type: str
+  faiss_index_path:
+    description: Url of the dataset
+    type: str
+  image_index_column_name:
+    description: Name of the column in the dataset that contains the image index
+    type: str
```
RobbeSneyders, line 33: I don't think we need this, we can just use the index of the `id_mapping`. We might want to take the `url_column_name` as a parameter, since this is not fixed in autofaiss.

src/fondant/components/retrieve_images_from_faiss_index/src/main.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def __init__(  # noqa PLR0913
+        self,
+        dataset_path: str,
+        faiss_index_path: str,
+        image_index_column_name: str,
+        clip_model: str = "laion/CLIP-ViT-B-32-laion2B-s34B-b79K",
+        num_images: int = 2,
+    ):
+        self.model = CLIPTextModelWithProjection.from_pretrained(clip_model)
```
RobbeSneyders, line 28: We need to put the model on the device we want to use.

src/fondant/components/retrieve_images_from_faiss_index/src/main.py
```diff
@@ -0,0 +1,106 @@
...
+                    "Dataframe does not contain a prompt or embedding column. "
+                    "Please provide one of both."
+                )
+                raise ValueError(msg)
+
+            indices = self.retrieve_from_index(query, self.number_of_images)
+            for i in indices:
+                url = self.dataset[self.dataset[self.image_index_column_name] == i][
+                    "url"
+                ]
```
RobbeSneyders, line 94: I think it would be better to read the id mapping in memory, instead of looking through the whole Dask dataframe each time.

src/fondant/components/retrieve_images_from_faiss_index/tests/resources/create_dummy_dataset.py
```diff
@@ -0,0 +1,22 @@
...
+    ddf = dd.from_pandas(
+        pd.DataFrame(data),
+        npartitions=1,
+    )  # You can adjust the number of partitions as needed
+
+    # Store as Parquet
+    ddf.to_parquet("./dataset.parquet")
+
+
+create_image_index_mapping()
```
RobbeSneyders, line 22: Does this run automatically? Or should we add this to a `conftest.py` file?

src/fondant/components/retrieve_images_from_faiss_index/tests/test_component.py
```diff
@@ -0,0 +1,36 @@
...
+        {
+            "id": ["a", "b", "c", "d"],
+            "image_url": ["http://a", "http://b", "http://c", "http://d"],
+            "prompt_id": ["1", "1", "2", "2"],
+        },
+    )
+
+    component = RetrieveImagesFromFaissIndex(
+        dataset_url="./tests/resources",
+    )
```
RobbeSneyders, line 27: This doesn't seem up to date with the latest version of the component.

src/fondant/components/retrieve_images_from_faiss_index/src/main.py
```diff
@@ -0,0 +1,115 @@
...
+                prompt = row["prompt"]
+                query = self.embed_prompt(prompt)
+            else:
+                msg = "Dataframe does not contain a prompt column."
+                raise ValueError(msg)
+
+            indices = self.retrieve_from_index(query, self.number_of_images)
+            for i in indices:
+                url = self.image_urls[i]
+                row_to_add = (index, prompt, i, url) if prompt else (index, i, url)
```
mrchtr, line 103:
```suggestion
                row_to_add = (index, prompt, i, url)
```

src/fondant/components/retrieve_images_from_faiss_index/src/main.py
```diff
@@ -0,0 +1,122 @@
...
+                results.append(row_to_add)
+
+        results_df = pd.DataFrame(
+            results,
+            columns=["id", "prompt", "image_url"],
+        )
+        results_df = results_df.set_index("id")
+        return results_df
+
+    def transform(self, dataframe: dd.DataFrame) -> dd.DataFrame:
```
RobbeSneyders, line 109: Is there a reason to use a `DaskTransformCompnoent` instead of a `PandasTransformComponent`?

mrchtr: Thought I need more control about the dask dataframe. Changed it to a PandasComponent.

src/fondant/components/retrieve_images_from_faiss_index_by_prompt/src/main.py
RobbeSneyders, line 1: Can we make this name shorter?
Also we're retrieving urls, not images.  :sweat_smile: 

Maybe `retrieve_from_clip_by_prompt`?

RobbeSneyders (approved):
Thanks @mrchtr!

PR #878, opened 2024-02-26
Allow more supported docker versions

RobbeSneyders:
The docker version installed on the VM I'm using is `20.10.17` and everything is working if I remove this check. So we can lower it to support a wider range.


GeorgesLorre (approved):
Did you also check the docker compose version ?

RobbeSneyders:
> Did you also check the docker compose version ?

It doesn't have a compose version pre-installed. I installed a recent version myself.

PR #879, opened 2024-02-26
Add struct to types

PhilippeMoussalli:
PR that adds `Struct` to list of valid types in Dask dataframe. Required for Mediahuis to ingest data with a certain format (documents with nested named attributes)


mrchtr (commented):
Would you up for adjusting the integration test as well to test this type?

RobbeSneyders (approved):
Thanks @PhilippeMoussalli!

RobbeSneyders:
> Would you up for adjusting the integration test as well to test this type?

Not sure if this is needed. Since it's also our simple example, I would keep it limited to basic and recommended things.

mrchtr (approved):

PR #883, opened 2024-02-28
Bugfix remote file detection

PhilippeMoussalli:
PR that adjusts the method of detection of the `base_path`

My local `base_path` named `data_dir` was interpreted as a remote path since it partially matched one of the support fsspec schemes (`data`). Now we strictly detect the protocol from the base path and check if it matches any of the supported fsspec schemes.


mrchtr (approved):
Thanks @PhilippeMoussalli!

PR #884, opened 2024-02-28
Fix default lightweight component

PhilippeMoussalli:
PR that enables defining default values for lightweight components

### DIFF
--- a/src/fondant/core/component_spec.py
+++ b/src/fondant/core/component_spec.py
@@ -55,15 +55,18 @@ class Argument:
         return lookup[self.type]
 
     def to_spec(self):
-        return {
-            k: v
-            for k, v in {
-                "description": self.description,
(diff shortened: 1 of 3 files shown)


src/fondant/core/component_spec.py
```diff
@@ -55,15 +55,18 @@ def kubeflow_type(self) -> str:
...
-            if v is not None
+        spec = {
+            "type": self.type.__name__,
+            "description": str(self.description),
         }
+        if self.optional:
+            spec["default"] = (
+                self.default
+                if type(self.default) in [dict, list, int]
+                else str(self.default)
```
RobbeSneyders, line 66: Can you explain this logic? Do we need to convert eg. floats to string?

RobbeSneyders (approved):

Issue #882, opened 2024-02-28
Enable defining timestamps with tz


philippe-ml6:
Currently passing timezones in the timestamp schema seems to result in error
```
pa.timestamp("us", tz='UTC')
```

```bash
    component_spec = ref.get_component_spec()
  File "/home/philippe/Scripts/ml6-chat-search-backend/venv/lib/python3.10/site-packages/fondant/pipeline/lightweight_component.py", line 110, in get_component_spec
    return ComponentSpec(
  File "/home/philippe/Scripts/ml6-chat-search-backend/venv/lib/python3.10/site-packages/fondant/core/component_spec.py", line 126, in __init__
    self._validate_spec()
  File "/home/philippe/Scripts/ml6-chat-search-backend/venv/lib/python3.10/site-packages/fondant/core/component_spec.py", line 155, in _validate_spec
    raise InvalidComponentSpec.create_from(e)
fondant.core.exceptions.InvalidComponentSpec: {'type': None} is not valid under any of the given schemas

Failed validating 'anyOf' in schema['properties']['produces']['additionalProperties']:
    {'anyOf': [{'$ref': '#/definitions/field'},
               {'$ref': '#/definitions/additionalProperties'}]}

On instance['produces']['updated']:
    {'type': None}

```

PR #888, opened 2024-02-29
Enable timezone type for timestamp pyarrow type

PhilippeMoussalli:
Fixes https://github.com/ml6team/fondant/issues/882

openapi spec: https://swagger.io/docs/specification/data-models/data-types/#:~:text=date%2Dtime%20%E2%80%93%20the%20date%2Dtime%20notation%20as%20defined%20by%20RFC%203339%2C%20section%205.6%2C%20for%20example%2C%202017%2D07%2D21T17%3A32%3A28Z

### DIFF
--- a/src/fondant/core/schema.py
+++ b/src/fondant/core/schema.py
@@ -159,6 +159,7 @@ class Type:
             else:
                 type_ = cls._validate_data_type(data_type)
             validated_fields.append(pa.field(name, type_))
+
         return cls(pa.struct(validated_fields))
 
     @classmethod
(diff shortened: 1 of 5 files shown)


src/fondant/core/schema.py
```diff
@@ -197,7 +197,13 @@ def from_dict(cls, json_schema: dict):
             fields = [(name, cls.from_dict(prop)) for name, prop in properties.items()]
             return cls.struct(fields)
 
+        if type_name == "timestamp":
+            return cls(type_name)
```
RobbeSneyders, line 201: I don't think we need to support this as a type, only as a format, right?

RobbeSneyders (commented):
Thanks @PhilippeMoussalli!

src/fondant/core/schema.py
philippe-ml6: Good catch, leftover from when I was trying something different. Removed

RobbeSneyders (approved):

PR #892, opened 2024-03-06
Implement non-blocking to_parquet

RobbeSneyders:
This is necessary to prevent OOM when processing large datasets. Added all the info in comments in the code.

### DIFF
--- a/src/fondant/component/data_io.py
+++ b/src/fondant/component/data_io.py
@@ -4,7 +4,10 @@ import typing as t
 from collections import defaultdict
 
 import dask.dataframe as dd
+import dask.distributed
+import pyarrow as pa
 from dask.diagnostics import ProgressBar
+from dask.distributed import as_completed
 
 from fondant.core.component_spec import OperationSpec
 from fondant.core.manifest import Manifest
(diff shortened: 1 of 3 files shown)


src/fondant/component/data_io.py
```diff
@@ -204,7 +207,34 @@ def _write_dataframe(self, dataframe: dd.DataFrame) -> dd.core.Scalar:
...
+        # https://dask.discourse.group/t/improving-pipeline-resilience-when-using-to-parquet-and-preemptible-workers/2141
+        to_parquet_tasks = [
+            d.to_parquet(
+                os.path.join(location, f"part.{i}.parquet"),
+                schema=pa.schema(list(schema.items())),
+            )
+            for (i, d) in enumerate(dataframe.to_delayed())
+        ]
+
+        client: dask.distributed.Client = dask.distributed.get_client()
```
mrchtr, line 231: `get_client()` will retrieve the global client I guess. But shouldn't we use the client we have initialised in the setup method explicitly to make clear where the client comes from?

mrchtr (approved):
Thanks @RobbeSneyders. It seems like we have finally a robust solution for our oom issues.

src/fondant/component/data_io.py
RobbeSneyders: In our setup there should only be one client, and `get_client` retrieves the client connected to the current worker. So this should provide the correct one.

src/fondant/component/data_io.py
```diff
@@ -204,7 +207,34 @@ def _write_dataframe(self, dataframe: dd.DataFrame) -> dd.core.Scalar:
...
+        futures = client.compute(to_parquet_tasks)
+
+        # As each future completes, release it so the memory can be reclaimed
+        for future in as_completed(futures):
+            future.result()
+            future.release()
 
     @staticmethod
     def _create_write_task(
```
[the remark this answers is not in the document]
RobbeSneyders: Good catch!

PR #894, opened 2024-03-06
Add retrieve from faiss by embedding component

mrchtr:
Adding component to retrieve image_urls from the faiss index using embeddings. Similar implementation to `retrieve_from_faiss_by_prompt`.


RobbeSneyders (approved):
Thanks @mrchtr!

PR #901, opened 2024-03-11
Refactor pipeline interface

mrchtr:
Feature branch for refactoring the pipeline interface. 

Going to create separate PRs to maintain an overview of all changes and avoid breaking the main branch.


RobbeSneyders:
FYI, the docs pipeline is failing because we reference the `Pipeline` class in the docs.

docs/pipeline.md
```diff
@@ -1,20 +1,21 @@
...
-(DAG) of different components that need to be executed. With Fondant, you can use both reusable
-components and custom components, and chain them into a pipeline.
+A Fondant Dataset is a checkpoint in a Directed Acyclic Graph 
+(DAG) of one or more different components that need to be executed. With Fondant, you can use both reusable
+components and custom components, and chain them together.
 
-## Composing a pipeline
 
-Start by creating a `pipeline.py` file and adding the following code.
+[//]: # (TODO update this section once we have the workspace)
```
RobbeSneyders, line 8: What is the idea of the workspace now?

src/fondant/cli.py
```diff
@@ -840,47 +872,45 @@ def pipeline_from_string(string_ref: str) -> Pipeline:  # noqa: PLR0912
...
+        raise DatasetImportError(msg)
 
-    if len(pipeline_instances) > 1:
+    # Skip this one and choose the first dataset instance?
+    if len(dataset_instances) > 1:
         msg = (
-            f"Found multiple instantiated pipelines in {module_str}. Only one pipeline "
-            f"can be present"
+            f"Found multiple instantiated datasets in {module_str}. Use the first dataset to start "
+            f"the execution."
```
RobbeSneyders, line 909:
```suggestion
            f"Found multiple instantiated datasets in {module_str}. Using the first dataset to start "
            f"the execution."
```

Shouldn't we use the last dataset? Or should it not matter? And if it doesn't matter, should we even log this?

src/fondant/component/data_io.py
```diff
@@ -196,16 +196,23 @@ def validate_dataframe_columns(dataframe: dd.DataFrame, columns: t.List[str]):
...
         """Create dataframe writing task."""
-        location = (
-            f"{self.manifest.base_path}/{self.manifest.pipeline_name}/"
-            f"{self.manifest.run_id}/{self.operation_spec.component_name}"
+        location = self.manifest.get_dataset_columns_locations(
+            columns=dataframe.columns,
         )
 
+        if len(set(location)) > 1:
+            msg = "Writing to multiple locations is currently not supported."
```
RobbeSneyders, line 204: How would we reach this state?

src/fondant/core/component_spec.py
```diff
@@ -411,7 +411,7 @@ def _validate_mappings(self) -> None:
             for key, value in mapping.items():
                 if not isinstance(value, (str, pa.DataType)):
                     msg = f"Unexpected type {type(value)} received for key {key} in {name} mapping"
-                    raise InvalidPipelineDefinition(msg)
+                    raise InvalidWorkspaceDefinition(msg)
```
RobbeSneyders, line 414: Is `InvalidWorkspaceDefinition` a logical name here? Shouldn't we use `InvalidOperation` or something similar?

src/fondant/core/exceptions.py
```diff
@@ -15,7 +15,7 @@ class InvalidComponentSpec(ValidationError, FondantException):
     """Thrown when a component spec cannot be validated against the schema."""
 
 
-class InvalidPipelineDefinition(ValidationError, FondantException):
+class InvalidWorkspaceDefinition(ValidationError, FondantException):
     """Thrown when a pipeline definition is invalid."""
```
RobbeSneyders, line 19: pipeline :)

src/fondant/core/manifest.py
```diff
@@ -24,18 +24,19 @@ class Metadata:
     Class representing the Metadata of the manifest.
 
     Args:
-        base_path: the base path used to store the artifacts
-        pipeline_name: the name of the pipeline
+        dataset_name: the name of the pipeline
```
RobbeSneyders, line 27: pipeline :)

src/fondant/core/manifest.py
```diff
@@ -260,25 +279,62 @@ def evolve(  # : PLR0912 (too many branches)
...
+        # TODO: handle index location - do we have to update/rewrite the index to the new location?
+        if working_dir:
+            field = Field.create(
+                name="index",
+                working_dir=working_dir,
+                run_id=run_id,
+                component_id=component_id,
+                dataset_name=self.dataset_name,
+            )
+            evolved_manifest.add_or_update_field(field, overwrite=False)
```
RobbeSneyders, line 311:
```suggestion
            index = Field.create(
                name="index",
                working_dir=working_dir,
                run_id=run_id,
                component_id=component_id,
                dataset_name=self.dataset_name,
            )
            evolved_manifest.add_or_update_field(index, overwrite=False)
```

For clarity

src/fondant/core/schema.py
```diff
@@ -274,6 +274,33 @@ def __repr__(self):
     def __eq__(self, other):
         return vars(self) == vars(other)
 
+    @classmethod
+    def create(  # noqa: PLR0913
```
RobbeSneyders, line 278: I understand that you want to group this logic somewhere. However now the `Field` class needs to know about all these arguments. I'm not sue this is the best approach.

src/fondant/dataset/dataset.py
```diff
@@ -472,6 +484,9 @@ def register_operation(
         input_dataset: t.Optional["Dataset"],
         output_dataset: t.Optional["Dataset"],
     ) -> None:
+        if self._graph is None:
+            self._graph = OrderedDict()
```
RobbeSneyders, line 488: Is this necessary? `self._graph` is created as an `OrderedDict` in the `__init__` method.

src/fondant/dataset/dataset.py
```diff
@@ -508,15 +529,14 @@ def read(
...
             An intermediate dataset.
         """
-        if self._graph:
-            msg = "For now, at most one read component can be applied per pipeline."
-            raise InvalidPipelineDefinition(
-                msg,
-            )
+        if dataset_name is None:
+            fake = Faker()
+            dataset_name = f"{fake.word()}-{fake.word()}"
```
RobbeSneyders, line 539: Does this need to be unique? And if it does, could we just use `dataset-{random_number}`? It's a shame that we need a new dependency for this.

src/fondant/dataset/dataset.py
```diff
@@ -560,32 +581,20 @@ def depth_first_traversal(node: str):
...
         """
         self.sort_graph()
-        self._validate_pipeline_definition(run_id)
+        self._validate_dataset_definition()
 
-    def _validate_pipeline_definition(self, run_id: str):
+    def _validate_dataset_definition(self):
         """
-        Validates the pipeline definition by ensuring that the consumed and produced subsets and
+        Validates the workspace definition by ensuring that the consumed and produced subsets and
```
RobbeSneyders, line 597: Here you use "workspace definition", and in the function name "dataset definition", which I think both are not ideal. I think "operation definition" fits probably better?

src/fondant/dataset/dataset.py
```diff
@@ -658,42 +672,42 @@ def __repr__(self) -> str:
...
-            pipeline: The pipeline this dataset is a part of.
-        """
-        self.manifest = manifest
-        self.pipeline = pipeline
-
     @property
     def fields(self) -> t.Mapping[str, Field]:
         """The fields of the manifest as an immutable mapping."""
-        return dict(self.manifest.fields)
+        if self.manifest:
```
RobbeSneyders, line 678: Is this necessary? This is a required argument to the `__init__` method.

RobbeSneyders (commented):
Thanks @mrchtr, great work!

I haven't reviewed the tests yet (the PR is quite big :sweat_smile:), but already left my comments on the code.

src/fondant/dataset/dataset.py
mrchtr: Convinced - we just need to make sure we have some random dataset name.

docs/pipeline.md
mrchtr: There will be no workspace. We will use a working_directory to store all the execution artifacts. We will update the documentation in a separate PR.

src/fondant/cli.py
mrchtr: I've checked it again. We don't need it since we don't create a deepcopy of the graph. Every dataset will contain the same computational graph after applying a new operations. We are not able to create a deepcopy since the graph contains some mappingproxy objects. Shouldn't be a problem for now, if we just allow a single dataset object in a module. 

For the case we allow something like this: 

```python
dataset_1 = Dataset.read(...)
dataset_2 = dataset_1.apply(...)
dataset_3 = dataset_2.apply(...)
```
we will run into issues. For now we could add a validation that every dataset of a module contains the same graph, wdyt?

GeorgesLorre (approved):

Issue #887, opened 2024-02-29
Add workspace context


mrchtr:
The workspace will be used to hold all relevant execution information, e.g. name, base_path, run_id, etc.
It has still to be decided where and how to initialise the workspace. 

Add cli command to specify workspace.

mrchtr:
We had an offline discussion and decided to move away from a workspace. The cli takes an `working_directory` argument and uses this as the former `base_path`

PR #902, opened 2024-03-11
Implementation new dataset interface

mrchtr:
First steps for the implementation of the new dataset interface: 
- Removed Pipeline class
- Added Workspace singleton to hold pipeline name, base_path, etc. .. (shouldn't be the focus of this PR)
- Moved `Pipeline.read(..)` to Dataset class


tests/pipeline/test_pipeline.py
```diff
@@ -644,3 +646,18 @@ def test_consumes_name_to_name_mapping(
     ].operation_spec.to_dict()["consumes"] == {
         "images_data": {"type": "binary"},
     }
+
+
+def test_new_pipeline_interface():
```
mrchtr, line 651: Simple test case to test the implemented changes.

src/fondant/cli.py
```diff
@@ -761,7 +761,7 @@ def _called_with_wrong_args(f):
         del tb
 
 
-def pipeline_from_string(string_ref: str) -> Pipeline:  # noqa: PLR0912
+def pipeline_from_string(string_ref: str) -> Dataset:  # noqa: PLR0912
```
RobbeSneyders, line 764:
```suggestion
def dataset_from_string(string_ref: str) -> Dataset:  # noqa: PLR0912
```

src/fondant/cli.py
```diff
@@ -856,14 +856,14 @@ def pipeline_from_string(string_ref: str) -> Pipeline:  # noqa: PLR0912
     )
 
 
-def pipeline_from_module(module_str: str) -> Pipeline:
+def pipeline_from_module(module_str: str) -> Dataset:
```
RobbeSneyders, line 859:
```suggestion
def dataset_from_module(module_str: str) -> Dataset:
```

RobbeSneyders (commented):
Thanks @mrchtr!

Not sure if going with a singleton is the best approach here. It can make testing harder and can become a bit too magic. We might still want to inject it into the `Dataset` class instead.

GeorgesLorre:
Starting to look good @mrchtr 

I agree that that the workspace concept might still need some further refinement. How we instantiate it and how we link it to an execution but this already serving us. We could refine it in a follow up PR or it might still be mergable in the runners/compilers.

The pipeline/Dataset changes are looking good, still some todo's but the mvp is here:
- move the `.read()` method
- add validation

mrchtr:
> Starting to look good @mrchtr
> 
> I agree that that the workspace concept might still need some further refinement. How we instantiate it and how we link it to an execution but this already serving us. We could refine it in a follow up PR or it might still be mergable in the runners/compilers.
> 
> The pipeline/Dataset changes are looking good, still some todo's but the mvp is here:
> 
> * move the `.read()` method
> * add validation
> 
> I'm ok with merging this in the other branch and focusing on green pipelines. Then we can add the missing functionality.

Sounds like a plan! Going to fix the tests and then we can divide the work from there on.

src/fondant/dataset/dataset.py
```diff
@@ -440,30 +440,62 @@ def get_nested_dict_hash(input_dict):
...
+
+    @staticmethod
+    def _validate_workspace_name(name: str) -> str:
+        pattern = r"^[a-z0-9][a-z0-9_-]*$"
+        if not re.match(pattern, name):
+            msg = f"The workspace name violates the pattern {pattern}"
+            raise InvalidWorkspaceDefinition(msg)
+        return name
+
+    def get_run_id(self) -> str:
```
mrchtr, line 465: Not sure if the caching is currently working. 
The run id is used to determine changes. The workspace should hold run ids in the future as well. I could think about a combination of runner type and working directory.

src/fondant/dataset/dataset.py
```diff
@@ -440,30 +440,62 @@ def get_nested_dict_hash(input_dict):
...
-class Pipeline:
-    """Class representing a Fondant Pipeline."""
+class Workspace:
+    """Workspace holding environment information for a Fondants execution environment."""
 
     def __init__(
         self,
         name: str,
-        *,
         base_path: str,
```
mrchtr, line 449: I would rename this into `working_directory` as part of #887 to make the difference between `Dataset` and `Workspace` more clear. 
A `Dataset` hold all relevant information (e.g. name, description, manifest, potential metadata in the future). 
A `Workspace` is used to determine the execution environment (e.g. runner information, working directories).

src/fondant/dataset/dataset.py
```diff
@@ -440,30 +440,62 @@ def get_nested_dict_hash(input_dict):
         return get_nested_dict_hash(component_op_uid_dict)
 
 
-class Pipeline:
-    """Class representing a Fondant Pipeline."""
+class Workspace:
```
mrchtr, line 443: I would do the clean up of this class as part of #887.

src/fondant/dataset/dataset.py
```diff
@@ -512,11 +549,7 @@ def read(
         Returns:
             An intermediate dataset.
         """
-        if self._graph:
-            msg = "For now, at most one read component can be applied per pipeline."
-            raise InvalidPipelineDefinition(
-                msg,
-            )
+        # TODO: add method call to retrieve workspace context, and make passing workspace optional
```
mrchtr, line 552: Every operation method (read, apply, write) takes the workspace as an argument. 
@GeorgesLorre I guess we can use the workspace initialised in the `cli` to pass it somehow to these methods. Maybe as part of arguments? I would propose to clean this up as part of #887.

src/fondant/cli.py
```diff
@@ -607,18 +607,28 @@ def run_local(args):
     from fondant.dataset.runner import DockerRunner
 
     extra_volumes = []
+    # use workspace from cli command
+    # if args.workspace exists
+
+    workspace = getattr(args, "workspace", None)
```
mrchtr, line 613: I've added already the workspace to the cli. For now, the implementation is using a default workspace. 
The run command loads the dataset object of the file (former pipeline) and executes the operations. 

As part of #887  I would handle a proper default workspace initialisation, and we can think about refer to a workspace file (for instance .workspace/local.env containing information of environment and runner) or a definition inside the source code. 
At least something like this `fondant run local ... --workspace ./local.env` feels natural.

mrchtr (commented):
I've added some notes to discuss the open topics. Feel free to have a look @GeorgesLorre. :)

src/fondant/dataset/dataset.py
GeorgesLorre: This is indeed not ideal. Ideally we inject these values at compile time. That way we can keep the datasets separate from the workspace.

src/fondant/dataset/dataset.py
GeorgesLorre: I would not put more effort into it for this PR. we can test and fix separately

GeorgesLorre (approved):

PR #904, opened 2024-03-11
Add string as fallback index type when writing data

RobbeSneyders:
`pa.from_numpy_dtype` fails when passing an object dtype, which Dask uses when the dtype is not explicitly known. I propose to fall back on `string` if this is the case. This will probably be correct for the index column in most cases when the `dtype` is unknown, and non-breaking in most other cases.


mrchtr (approved):
Thanks @RobbeSneyders!

PR #908, opened 2024-03-18
Skip transformation of partition if partition is empty

mrchtr:
n the case that a `PandasTransformComponent` receives an empty dask partition, the execution fails. There could be several reasons why a partition is empty, and it may depend on a custom implementation. We should catch this case and not halt the component's execution.
This PR adds a check to see if the `wrapped_transform` receives an empty partition. If it does, we skip the execution and return the received dataframe.

### DIFF
--- a/src/fondant/component/executor.py
+++ b/src/fondant/component/executor.py
@@ -492,18 +492,19 @@ class PandasTransformExecutor(TransformExecutor[PandasTransformComponent]):
         """
 
         def wrapped_transform(dataframe: pd.DataFrame) -> pd.DataFrame:
-            # Call transform method
-            if dataframe.empty:
-                logger.info("Received empty partition, skipping transformation.")
-                return dataframe
-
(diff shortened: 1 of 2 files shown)


src/fondant/component/executor.py
```diff
@@ -491,14 +492,19 @@ def wrap_transform(
...
             columns = [
                 name for name, field in operation_spec.operation_produces.items()
             ]
 
+            if dataframe.empty:
+                logger.info("Received empty partition, skipping transformation.")
+                return pd.DataFrame(columns=columns)
+
+            # Call transform method
+            dataframe = transform(dataframe)
```
RobbeSneyders, line 505: Can we simplify this like this?

```suggestion
            if not dataframe.empty:
                dataframe = transform(dataframe)
            else:
                logger.info("Received empty partition, skipping transformation.")
```

PR #914, opened 2024-03-27
Bug fix for retrieve from faiss by prompt

mrchtr:
There was a bug in the retrieve from faiss by prompt component related to cpu. 
The number of workers was calculated as float instead of an int.


RobbeSneyders (approved):
I actually thought this casted to an int.

PR #917, opened 2024-04-02
Feature/dataset based compilers

GeorgesLorre:


src/fondant/dataset/dataset.py
```diff
@@ -857,4 +858,4 @@ def write(
             resources=resources,
             cache=cache,
         )
-        self._apply(operation)
+        return self._apply(operation)
```
mrchtr, line 861: The only difference to the `.apply` method is that the `write` method does ignore the produces. 

I would leave it like this but what do you think about adding a deprecation warning to the `write` method? 

We could use this method in the future export the dataset from the working directory to a remote storage for instance.

mrchtr (commented):
Thanks @GeorgesLorre!

docs/runners/local.md
```diff
@@ -15,35 +15,37 @@ Make sure that you have installed Docker compose on your system. You can find mo
 about this in the [installation](../guides/installation.md) guide.
 
 
-### Running a pipeline with the local runner
+### Running a dataset with the local runner
```
mrchtr, line 18: I am not the biggest fan of "Running a dataset". Whats about "Materialize a dataset"

docs/runners/sagemaker.md
```diff
@@ -48,26 +48,28 @@ AWS with a role that has all the required permissions to launch a SageMaker pipe
...
+        working_dir=<s3_bucket>,
         role_arn=<role_arn>,
-        pipeline_name=<pipeline_name>
+        pipeline_name=<sagemaker_pipeline_name>
     )
     ```
 
 
-Once your pipeline is running you can monitor it using the SageMaker [Studio](https://aws.amazon.com/sagemaker/studio/).
+Once your workflow is running you can monitor it using the SageMaker [Studio](https://aws.amazon.com/sagemaker/studio/).
```
mrchtr, line 72: I like `workflow`. Maybe we can either use workflow or materialized dataset instead of pipeline.

docs/runners/sagemaker.md
```diff
@@ -1,11 +1,11 @@
 ### SageMaker Runner
 
-Leverage [AWS SageMaker](https://aws.amazon.com/sagemaker/) to run your Fondant pipelines.
+Leverage [AWS SageMaker](https://aws.amazon.com/sagemaker/) to run your Fondant datasets.
```
mrchtr, line 3:
```suggestion
Leverage [AWS SageMaker](https://aws.amazon.com/sagemaker/) to run your Fondant workflow.
```

docs/runners/sagemaker.md
```diff
@@ -1,11 +1,11 @@
...
 
-This makes it easy to scale up your pipelines in a serverless manner without worrying about infrastructure
+This makes it easy to scale up your datasets in a serverless manner without worrying about infrastructure
```
mrchtr, line 5:
```suggestion
This makes it easy to scale up your workflow in a serverless manner without worrying about infrastructure
```

docs/runners/sagemaker.md
```diff
@@ -1,11 +1,11 @@
...
 deployment.
 
-The Fondant SageMaker runner will compile your pipeline to a SageMaker pipeline spec and submit it to SageMaker.
+The Fondant SageMaker runner will compile your Fondant dataset to a SageMaker pipeline spec and submit it to SageMaker.
```
mrchtr, line 8:
```suggestion
The Fondant SageMaker runner will compile your Fondant workflow to a SageMaker pipeline spec and submit it to SageMaker.
```

docs/runners/vertex.md
```diff
@@ -3,13 +3,13 @@
...
 help you
-orchestrate your Fondant pipelines in a serverless manner. This makes it easy to scale up your
+orchestrate your Fondant workflows in a serverless manner. This makes it easy to scale up your
 pipelines without worrying about infrastructure
 deployment.
 
 Vertex AI pipelines leverages Kubeflow pipelines under the hood. The Vertex compiler will take your
-pipeline and compile it to a Kubeflow pipeline spec.
-This spec can be used to run your pipeline on Vertex.
+Fondant dataset and compile it to a Kubeflow pipeline spec.
```
mrchtr, line 11:
```suggestion
Fondant workflow and compile it to a Kubeflow pipeline spec.
```

docs/runners/vertex.md
```diff
@@ -3,13 +3,13 @@
...
+This spec can be used to run your dataset on Vertex.
```
mrchtr, line 12:
```suggestion
This spec can be used to run your workflow on Vertex.
```

docs/runners/kfp.md
```diff
@@ -16,39 +16,40 @@ Make sure to install Fondant with the Kubeflow runner extra.
 pip install fondant[kfp]
 ```
 
-### Running a pipeline with Kubeflow
+### Running a dataset with Kubeflow
```
mrchtr, line 19:
```suggestion
### Materialize a dataset with Kubeflow
```

mrchtr (commented):
Thanks @GeorgesLorre! I like the workflow wording instead of pipeline in the documentation. Maybe we can use this more often. 

We could explain these concepts somewhere else. 
E.g.: A dataset can be materialized by running a workflow. The workflow has to be compiled first.

src/fondant/dataset/dataset.py
GeorgesLorre: I'm not sure I completely understand. Are we deprecating the write ?

mrchtr: Thats the question if we want to do it. The apply and write looking quite similar now.

GeorgesLorre:
@mrchtr I updated the docs based on your suggestions

mrchtr (approved):

PR #920, opened 2024-04-03
Fixing prep release pipeline

mrchtr:


mrchtr:
Changes are not needed. I've found a way to increase the session duration of our AWS token.

PR #924, opened 2024-04-05
Empty produces leading into list index out of range

mrchtr:
When no produces is defined the location of the dataset couldn't be determined. Leading into list index out of range error. 
Using the dataset index to determine the dataset location.


GeorgesLorre (approved):
thanks!

PR #936, opened 2024-04-22
Bugfix:  Fondant base image build

mrchtr:
In `0.12.0` the lightweight component that uses a default image are not working. 

- During the build of base images the image tag list wasn't initialised correctly within the loop. This means we have overwritten the previous build images. Every base image now contains python version `10.11.9`
- We have limited dask to`<2024.3.0`, but only dask `>=2024.4.1` is compatible with python `>=10.11.9`. Limit python version of the base image build.


GeorgesLorre (approved):
nice find

PR #938, opened 2024-12-03
Rename blog post md files for windows file-system compatibility

alexdjulin:
Some MD files use the | character, which is not a valid finename character on Windows, creating some issues when cloning the repo


mrchtr:
> Some MD files use the | character, which is not a valid finename character on Windows, creating some issues when cloning the repo

Thanks @alexdjulin for the fix. Could you try to fix the failing pipeline? The rest looks good to me.

